PR #2, opened 2024-11-23
Update markdown and restructure files

romoh:
As part of the repo prep


.gitignore
```diff
@@ -1,390 +1 @@
...
-[Dd]ebug/
-[Dd]ebugPublic/
-[Rr]elease/
-[Rr]eleases/
-x64/
-x86/
-[Aa][Rr][Mm]/
-[Aa][Rr][Mm]64/
-bld/
 build/
```
cwize1, line 1: I think we also use `out/`

PR #6, opened 2024-12-02
Refresh partitions after veritysetup so that hash partition's UUID is accurate

fintelia:
Same as https://github.com/microsoft/azurelinux/pull/11247 but against new repository.

---
This change corrects the value set for the systemd.verity_root_hash= kernel cmdline parameter when hashDeviceMountIdType in the configuration is set to uuid.
 
In particular, the veritysetup command earlier in the function generates the UUID for the verity hash partition. But since the diskPartitions array is queried before that call, the partition would be marked as having an empty filesystem UUID.

An alternative would be to edit only the specific partition's UUID in the diskPartitions array, either by passing a manually generated UUID to veritysetup (via the --uuid=UUID flag) or by querying the chosen UUID after veritysetup had run.

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
@@ -814,6 +814,12 @@ func customizeVerityImageHelper(buildDir string, baseConfigPath string, config *
 		return fmt.Errorf("failed to stat file (%s):\n%w", grubCfgFullPath, err)
 	}
 
+	// Refresh disk partitions so that the hash partition's UUID is correct
+	diskPartitions, err = diskutils.GetDiskPartitions(loopback.DevicePath())
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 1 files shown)


toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -814,6 +814,12 @@ func customizeVerityImageHelper(buildDir string, baseConfigPath string, config *
 		return fmt.Errorf("failed to stat file (%s):\n%w", grubCfgFullPath, err)
 	}
 
+	// Refresh disk partitions so that the hash partition's UUID is correct
```
cwize1, line 817: It would be good this closer to the `veritysetup` call so that the relationship between the two is clearer.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
fintelia: I located it right before the `updateGrubConfigForVerity` that uses the updated diskPartitions value, but happy to move it

cwize1 (approved):

PR #12, opened 2024-12-03
Remove dependency on the 'rpm' package.

gmileka:
MIC ISO flow uses 'rpm' to query information about installed packages. However, some input images do not have 'rpm' installed - only 'rpmlib'.

MIC uses tdnf to install packages - so, it is safe to assume that tdnf is a required package for MIC operation.

This fix switches from using 'rpm' to using 'tdnf' so no new dependencies are required for images without 'rpm' installed.

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
@@ -231,8 +231,10 @@ func callTdnf(tdnfArgs []string, tdnfMessagePrefix string, imageChroot *safechro
 }
 
 func isPackageInstalled(imageChroot *safechroot.Chroot, packageName string) bool {
+
 	err := imageChroot.UnsafeRun(func() error {
-		return shell.ExecuteLive(true /*squashErrors*/, "rpm", "-qi", packageName)
+		_, _, err := shell.Execute("tdnf", "info", packageName, "--repo", "@system")
+		return err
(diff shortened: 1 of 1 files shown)


toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
```diff
@@ -272,23 +274,32 @@ func parseReleaseString(releaseInfo string) (packageRelease uint32, distroName s
...
 
-	releaseInfo := ""
-	err = imageChroot.UnsafeRun(func() error {
-		releaseInfo, _, err = shell.Execute("rpm", "-q", "--queryformat", "%{RELEASE}", packageName)
-		return err
-	})
-	if err != nil {
-		return info, fmt.Errorf("failed to query current package information for (%s):\n%w", packageName, err)
+	// Regular expressions to match Version and Release
+	versionRegex := regexp.MustCompile(`(?m)^Version\s+:\s+(\S+)`)
```
cwize1, line 287: nit: Put the regexes at the top of the file.

cwize1 (approved):

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
gmileka: cursious, why Chris though? Those are not constants that will be used in other places...

PR #14, opened 2024-12-04
Clarify logs of build host metadata.

cwize1:
We currently log information about the build host to assist with debugging. But some customers have confused the logs for metadata about the OS being customized. So, make it explicit that the logs are for the build host. Also, add the distro + version of the base image to the logs.

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
@@ -679,6 +679,13 @@ func customizeImageHelper(buildDir string, baseConfigPath string, config *imagec
 	}
 	defer imageConnection.Close()
 
+	imageConnection.Chroot().UnsafeRun(func() error {
+		distro, version := getDistroAndVersion()
+		logger.Log.Infof("Base OS distro: %s", distro)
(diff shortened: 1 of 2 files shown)


toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -679,6 +679,13 @@ func customizeImageHelper(buildDir string, baseConfigPath string, config *imagec
 	}
 	defer imageConnection.Close()
 
+	imageConnection.Chroot().UnsafeRun(func() error {
+		distro, version := getDistroAndVersion()
+		logger.Log.Infof("Base OS distro: %s", distro)
```
romoh, line 684: nit - should it be `Host OS` instead?

romoh (changes requested):
Minor suggestion

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
cwize1: These logs are for the base OS. Though I assume this comment is for the other file.

gmileka (approved):

romoh (approved):

PR #15, opened 2024-12-04
Allow 'ro' partitions.

cwize1:
When customizing a partition, don't apply the 'ro' mount option to allow the partitions to be customized.

In addition, require the user to specify the 'ro' option on verity partitions.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/imagecustomizer/docs/configuration.md
+++ b/toolkit/tools/imagecustomizer/docs/configuration.md
@@ -575,6 +575,9 @@ Specifies the configuration for dm-verity integrity verification.
 Note: Currently only root partition (`/`) is supported. Support for other partitions
 (e.g. `/usr`) may be added in the future.
 
+Note: The [filesystem](#filesystem-type) item pointing to this verity device, must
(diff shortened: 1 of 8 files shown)


toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -575,6 +575,9 @@ Specifies the configuration for dm-verity integrity verification.
 Note: Currently only root partition (`/`) is supported. Support for other partitions
 (e.g. `/usr`) may be added in the future.
 
+Note: The [filesystem](#filesystem-type) item pointing to this verity device, must
+include the `ro` option in the [mountPoint.options](#options-string).
```
amritakohli, line 579: so are we allowing it or requiring it? also why is it required? (curious - don't know a lot about this topic)

cwize1: It's now required for verity. (Previously we would add the flag silently for verity.) And it is now an optional for non-verity.

romoh (approved):

gmileka (approved):

PR #16, opened 2024-12-05
[ic/iso] Use relative paths when locating files.

gmileka:
When building the LiveOS ISO, we scan the input image (vhdx/qcow/raw/iso) for certain files (bootloaders, grub, etc).
We've been looking for them by name regardless of path. While this works, it could break should a file with the same name is dropped somewhere else on the input image.
To protect against such case, this change looks for file using their relative paths.

### **Checklist**
- [n/a] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
@@ -27,19 +27,19 @@ import (
 )
 
 const (
-	rawBootloaderDir = "/boot/efi/EFI/BOOT"
-	isoBootloaderDir = "/efi/boot"
+	osEspBootloaderDir = "/boot/efi/EFI/BOOT"
+	isoBootloaderDir   = "/efi/boot"
 
-	bootx64Binary        = "bootx64.efi"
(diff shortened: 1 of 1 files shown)


toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -27,12 +27,24 @@ import (
 )
 
 const (
-	bootx64Binary         = "bootx64.efi"
-	grubx64Binary         = "grubx64.efi"
-	grubx64NoPrefixBinary = "grubx64-noprefix.efi"
+	rawBootloaderDir = "/boot/efi/EFI/BOOT"
```
cwize1, line 30: What does "raw" mean? Perhaps rename to `osEspBootloaderDir`?

cwize1 (dismissed):

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: It applies to vhdx/qcow/raw images... picked raw as the most generic.
sure, osEsp prefix is better...

cwize1 (approved):

PR #18, opened 2024-12-06
OSModifier: Add services and modules support in EMU API

elainezhao96:
---
cherry picked commits from the approved pr: https://github.com/microsoft/azurelinux/pull/11153

What does the PR accomplish, why was it needed?

the goal is to add services and modules support in EMU API to support trident. 

- Create a modulelist type to avoid code duplication
- Add functions to update services and modules in osmodifier

### **Checklist**
- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/imagecustomizerapi/module.go
+++ b/toolkit/tools/imagecustomizerapi/module.go
@@ -14,13 +14,11 @@ type Module struct {
 	Options  map[string]string `yaml:"options"`
 }
 
-type ModuleList struct {
-	Modules []Module `yaml:"modules"`
-}
+type ModuleList []Module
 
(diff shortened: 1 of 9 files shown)


cwize1 (dismissed):

cwize1 (changes requested):
Can you add a proper description to the PR explaining what the change does and why. It can be helpful to have proper change descriptions when looking through the git history in the future.

cwize1 (approved):

elainezhao96:
commit [3222d9b] cannot be signed through force push, as it was duplicated with other code when I picked it from old repo, and was a little messy, so created new PR and will close this one.

PR #22, opened 2024-12-10
[ic/iso] Support for SELinux

gmileka:
[ic/iso] Support for SELinux.

To enable SELinux for LiveOS ISOs, we need to:
1. Update the SELinux policies in the selinux-policy package.
2. Update the overlay logic to mark the overlay folders correctly
   in dracut's overlayfs module.

Both these changes require updated versions of corresponding packages.

This change updates MIC such that:
- check if the installed selinux-policy and dracut packages are the
  new packages (or newer).
- if either one is older, selinux is disabled and a warning is printed
  out.

Note that disabling selinux for LiveOS isos has been the default behavior
in the past.

The changes include the following refactoring:
- moving package related functions to a new packageinformation.go file.
- introducing a new liveosisoutils.go to hold some of the new helper
  functions. This is because the main liveosisobild.go has grown too
  big and we need to start breaking it down.

---

### **Checklist**
- [n/a] Tests added/updated


toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
```diff
@@ -242,79 +234,6 @@ func isPackageInstalled(imageChroot *safechroot.Chroot, packageName string) bool
 	return true
 }
 
-func parseReleaseString(releaseInfo string) (packageRelease uint32, distroName string, distroVersion uint32, err error) {
```
gmileka, line 245: moved to packageinformation.go

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
```diff
@@ -242,79 +234,6 @@ func isPackageInstalled(imageChroot *safechroot.Chroot, packageName string) bool
...
-	distroVersionUint64, err := strconv.ParseUint(distroVersionString, 10 /*base*/, 32 /*size*/)
-	if err != nil {
-		return 0, "", 0, fmt.Errorf("failed to parse distro version (%s) into an unsigned integer:\n%w", distroVersionString, err)
-	}
-	distroVersion = uint32(distroVersionUint64)
-
-	return packageRelease, distroName, distroVersion, nil
-}
-
-func getPackageInformation(imageChroot *safechroot.Chroot, packageName string) (info packageInformation, err error) {
```
gmileka, line 276: moved to packageinformation.go

toolkit/tools/pkg/imagecustomizerlib/dracututils.go
```diff
@@ -7,26 +7,11 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"strconv"
 
 	"github.com/microsoft/azurelinux/toolkit/tools/internal/file"
 	"github.com/microsoft/azurelinux/toolkit/tools/internal/safechroot"
 )
 
-const (
```
gmileka, line 16: moved to liveosutils.go

toolkit/tools/pkg/imagecustomizerlib/dracututils.go
```diff
@@ -63,66 +48,3 @@ func addDracutDriver(dracutDriverName string, imageChroot *safechroot.Chroot) er
 	}
 	return addDracutConfig(dracutConfigFile, lines)
 }
-
-func getDracutVersion(rootfsSourceDir string) (dracutPackageInfo *DracutPackageInformation, err error) {
```
gmileka, line 67: Removed since we can use the now-generalized `getPackageInformation()` (can handle versions with different component numbers (y, y.y, y.y.y, etc)).

toolkit/tools/pkg/imagecustomizerlib/dracututils.go
```diff
@@ -63,66 +48,3 @@ func addDracutDriver(dracutDriverName string, imageChroot *safechroot.Chroot) er
...
-		PackageVersion: uint32(versionUint64),
-		PackageRelease: packageInfo.packageRelease,
-		DistroName:     packageInfo.distroName,
-		DistroVersion:  packageInfo.distroVersion,
-	}
-
-	return dracutPackageInfo, nil
-}
-
-func verifyDracutPXESupport(packageInfo *DracutPackageInformation) error {
```
gmileka, line 99: moved to liveosisoutils.go

toolkit/tools/pkg/imagecustomizerlib/packageinformation.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	if pi.PackageRelease < minimumVersionInfo.PackageRelease {
+		return fmt.Errorf("did not find required package release version (%s) (or newer) - found (%s)", minimumVersion, currentVersion)
+	}
+
+	return nil
+}
+
+func parseReleaseString(releaseInfo string) (packageRelease uint32, distroName string, distroVersion uint32, err error) {
```
gmileka, line 79: moved from customizepackages.go

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -802,7 +813,12 @@ func (b *LiveOSIsoBuilder) prepareLiveOSDir(inputSavedConfigsFilePath string, wr
 		return err
 	}
 
-	b.artifacts.dracutPackageInfo, err = getDracutVersion(writeableRootfsDir)
+	b.artifacts.dracutPackageInfo, err = getPackageInformationFromRootfsDir(writeableRootfsDir, "dracut")
+	if err != nil {
+		return err
+	}
+
+	b.artifacts.selinuxPolicyPackageInfo, err = getPackageInformationFromRootfsDir(writeableRootfsDir, "selinux-policy")
```
cwize1, line 821: MIC supports the user using their own custom SELinux policies. (See, UpdateSELinuxModeInConfigFile.) If the user is using the `selinux-policy` package, then we should do the version check. Otherwise, assume the user is using a compatible policy.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -316,12 +317,14 @@ func (b *LiveOSIsoBuilder) prepareRootfsForDracut(writeableRootfsDir string) err
 // outputs:
 // - returns a SavedConfigs objects with the new merged values.
 func updateSavedConfigs(savedConfigsFilePath string, newKernelArgs imagecustomizerapi.KernelExtraArguments,
```
cwize1, line 319: nit:

```golang
func updateSavedConfigs(savedConfigsFilePath string, newKernelArgs imagecustomizerapi.KernelExtraArguments,
	newPxeIsoImageBaseUrl string, newPxeIsoImageFileUrl string, newDracutPackageInfo *PackageVersionInformation,
	newSELinuxPackageInfo *PackageVersionInformation,
) (updatedSavedConfigs *SavedConfigs, err error) {
```

This makes it easier to distinguish between the function signature and the first statement.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -431,10 +437,15 @@ func (b *LiveOSIsoBuilder) updateGrubCfg(isoGrubCfgFileName string, pxeGrubCfgFi
 		return fmt.Errorf("failed to update the root kernel argument in the iso grub.cfg:\n%w", err)
 	}
 
-	inputContentString, err = updateSELinuxCommandLineHelperAll(inputContentString, imagecustomizerapi.SELinuxModeDisabled,
-		true /*allowMultiple*/, false /*requireKernelOpts*/)
+	err = verifyLiveOsSelinuxSupport(savedConfigs.OS.DracutPackageInfo, savedConfigs.OS.SELinuxPolicyPackageInfo)
 	if err != nil {
-		return fmt.Errorf("failed to set SELinux mode:\n%w", err)
+		logger.Log.Warnf("SELinux cannot be enabled due to older dracut and selinux-policy package versions.\n%w", err)
```
cwize1, line 442: If the versions are too old and the base image has SELinux enabled, then I think it is ok to only report a warning. But if the user explicitly asks for SELinux, then I think we should error out.

toolkit/tools/pkg/imagecustomizerlib/packageinformation.go
```diff
@@ -0,0 +1,204 @@
...
+			version += "."
+		}
+		version += strconv.FormatUint(versionComponent, 10)
+	}
+	return version
+}
+
+func (pi *PackageVersionInformation) getFullVersionString() string {
+	// yy.yy.yy-zz.azl3
+	return pi.getVersionString() + "-" + strconv.FormatUint(uint64(pi.PackageRelease), 10) + "." +
```
cwize1, line 36: Use `fmt.Sprintf`

toolkit/tools/pkg/imagecustomizerlib/packageinformation.go
```diff
@@ -0,0 +1,204 @@
...
+
+type PackageVersionInformation struct {
+	PackageVersionComponents []uint64 `yaml:"PackageVersionComponents"`
+	PackageRelease           uint32   `yaml:"PackageRelease"`
+	DistroName               string   `yaml:"DistroName"`
+	DistroVersion            uint32   `yaml:"DistroVersion"`
+}
+
+func (pi *PackageVersionInformation) getVersionString() string {
+	var version string
```
cwize1, line 24: Ideally you should use [strings.Builder](https://pkg.go.dev/strings#Builder) here.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -802,7 +813,12 @@ func (b *LiveOSIsoBuilder) prepareLiveOSDir(inputSavedConfigsFilePath string, wr
 		return err
 	}
 
-	b.artifacts.dracutPackageInfo, err = getDracutVersion(writeableRootfsDir)
+	b.artifacts.dracutPackageInfo, err = getPackageInformationFromRootfsDir(writeableRootfsDir, "dracut")
```
cwize1, line 816: `getPackageInformationFromRootfsDir` sets up the chroot environment. Doing that twice in a row seems wasteful.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: pending

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: pending

toolkit/tools/pkg/imagecustomizerlib/packageinformation.go
gmileka: pending

toolkit/tools/pkg/imagecustomizerlib/packageinformation.go
gmileka: pending

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: pending

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: fixed.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: fixed.

toolkit/tools/pkg/imagecustomizerlib/packageinformation.go
gmileka: fixed.

toolkit/tools/pkg/imagecustomizerlib/packageinformation.go
gmileka: fixed.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: fixed.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: Updated the logic. I did not take into account a user turning on selinux from the kernel command line arguments directly assuming that they should use `os.selinux.mode`. I think we do not have to cover this case - especially since it's a transient state until the updated packages are in all base images.

cwize1 (approved):

PR #24, opened 2024-12-10
Image Customizer: Restrict capabilities on scripts.

cwize1:
Restrict the set of Linux (root) capabilities given to custom scripts. This is intended to prevent scripts from accidentally affecting the build host machine. This is not intended as a security boundary, just a sanity check.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/imagecustomizer/docs/configuration.md
+++ b/toolkit/tools/imagecustomizer/docs/configuration.md
@@ -1171,6 +1171,20 @@ For example, `/boot` will be mounted before `/boot/efi`.
 
 Points to a script file (typically a Bash script) to be run during customization.
 
+Scripts are run with a limited set of capabilities. Specifically:
+
+- `CAP_CHOWN`
+- `CAP_DAC_OVERRIDE`
+- `CAP_DAC_READ_SEARCH`
+- `CAP_FOWNER`
+- `CAP_SETFCAP`
+
+Restricting the set of capabilities helps prevent scripts from accidentally affecting
+the host kernel.
+
+WARNING: Custom scripts are not considered to be on security boundary.
+Only use config files that you trust (or run image customizer in a security sandbox).
+
 <div id="script-path"></div>
 
 ### path [string]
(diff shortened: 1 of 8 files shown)


toolkit/tools/pkg/imagecustomizerlib/runscripts_test.go
```diff
@@ -66,3 +66,37 @@ resolv.conf exists
...
+	testTmpDir := filepath.Join(tmpDir, "TestCustomizeImageRunScriptsModprobe")
+	buildDir := filepath.Join(testTmpDir, "build")
+	configFile := filepath.Join(testDir, "runscripts-modprobe.yaml")
+	outImageFilePath := filepath.Join(testTmpDir, "image.raw")
+
+	// Customize image.
+	err = CustomizeImageWithConfigFile(buildDir, configFile, baseImage, nil, outImageFilePath, "raw", "",
+		"" /*outputPXEArtifactsDir*/, false /*useBaseImageRpmRepos*/, false /*enableShrinkFilesystems*/)
+	assert.ErrorContains(t, err, "failed to customize raw image")
+	assert.ErrorContains(t, err, "script (postCustomization[0]) failed")
```
romoh, line 101: Is the propagated error more descriptive? in other words, will the user understand what they are doing incorrectly?

cwize1: We also report the last line of stderr in the error. So, it depends on the command that fails and how good its error messages are.

I didn't include the stderr messages here in case they are inconsistent between platforms/versions.

toolkit/tools/imagecustomizer/docs/configuration.md
```diff
@@ -1171,6 +1171,20 @@ For example, `/boot` will be mounted before `/boot/efi`.
 
 Points to a script file (typically a Bash script) to be run during customization.
 
+Scripts are run with a limited set of capabilities. Specifically:
```
romoh, line 1174: nit - maybe add a hyperlink to https://man7.org/linux/man-pages/man7/capabilities.7.html ?

cwize1: Fixed.

romoh (approved):

PR #28, opened 2024-12-12
VM Tests: Run customized OS in VM.

cwize1:
Expand the basic test so that it takes the image produced by image customizer and then runs it in a VM using QEMU/KVM via libvirt.

---

### **Checklist**
- [x] Tests added/updated

### DIFF
--- a/test/vmtests/vmtests/test_no_change.py
+++ b/test/vmtests/vmtests/test_no_change.py
@@ -58,4 +58,4 @@ def test_no_change(
     vm.start()
 
     # Wait for VM to boot by waiting for it to request an IP address from the DHCP server.
-    vm.get_vm_ip_address(timeout=15)
+    vm.get_vm_ip_address(timeout=30)
--- a/test/vmtests/vmtests/test_no_change.py
+++ b/test/vmtests/vmtests/test_no_change.py
@@ -1,12 +1,19 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
+import os
 from pathlib import Path
+from typing import List
 
+import libvirt  # type: ignore
 from docker import DockerClient
 
 from .conftest import TEST_CONFIGS_DIR
+from .utils import local_client
+from .utils.closeable import Closeable
 from .utils.imagecustomizer import run_image_customizer
+from .utils.libvirt_utils import VmSpec, create_libvirt_domain_xml
+from .utils.libvirt_vm import LibvirtVm
 
 
 def test_no_change(
@@ -14,9 +21,13 @@ def test_no_change(
     image_customizer_container_url: str,
     core_efi_azl2: Path,
     test_temp_dir: Path,
+    test_instance_name: str,
+    libvirt_conn: libvirt.virConnect,
+    close_list: List[Closeable],
 ) -> None:
     config_path = TEST_CONFIGS_DIR.joinpath("nochange-config.yaml")
     output_image_path = test_temp_dir.joinpath("image.qcow2")
+    diff_image_path = test_temp_dir.joinpath("image-diff.qcow2")
 
     run_image_customizer(
         docker_client,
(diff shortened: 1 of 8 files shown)


test/vmtests/vmtests/test_no_change.py
```diff
@@ -26,3 +37,25 @@ def test_no_change(
...
+    domain_xml = create_libvirt_domain_xml(VmSpec(vm_name, 4096, 4, diff_image_path))
+
+    vm = LibvirtVm(vm_name, domain_xml, libvirt_conn)
+    close_list.append(vm)
+
+    # Start VM.
+    vm.start()
+
+    # Wait for VM to boot by waiting for it to request an IP address from the DHCP server.
+    vm.get_vm_ip_address(timeout=15)
```
gmileka, line 61: 15 seconds might be too short? (my local tests usually take longer)

gmileka: but it's for iso, so, there are a few more things that take place...

test/vmtests/vmtests/utils/libvirt_utils.py
```diff
@@ -0,0 +1,156 @@
...
+
+def _gen_disk_device_name(prefix: str, next_disk_indexes: Dict[str, int]) -> str:
+    disk_index = next_disk_indexes.get(prefix, 0)
+    next_disk_indexes[prefix] = disk_index + 1
+
+    match prefix:
+        case "vd" | "sd":
+            # The disk device name is required to follow the standard Linux device naming
+            # scheme. That is: [ sda, sdb, ..., sdz, sdaa, sdab, ... ]. However, it is
+            # unlikely that someone will ever need more than 26 disks. So, keep is simple
```
gmileka, line 148: nit: `keep it`

test/vmtests/vmtests/utils/libvirt_vm.py
```diff
@@ -0,0 +1,96 @@
...
+        interface_name = next(iter(interfaces))
+        addrs = interfaces[interface_name]["addrs"]
+        if len(addrs) < 1:
+            return None
+
+        addr = addrs[0]["addr"]
+        assert isinstance(addr, str)
+        return addr
+
+    def close(self) -> None:
```
gmileka, line 62: nit: would it be more clear to have two separate funcitons: `stop` and `delete`? The `close` can be the destructor calling both `stop` and `delete`.

test/vmtests/vmtests/test_no_change.py
cwize1: In my local testing, I see 6-11 seconds. So, I'll bump this up to 30 seconds.

test/vmtests/vmtests/utils/local_client.py
```diff
@@ -0,0 +1,194 @@
...
+                # Log the line.
+                if log_enabled:
+                    line_strip_newline = line[:-1] if line.endswith("\n") else line
+                    logging.log(self._log_level, "%s: %s", self._log_name, line_strip_newline)
+
+            self._pipe.close()
+            self._output = output.getvalue()
+
+
+class LocalProcess:
```
gmileka, line 92: Are we writing a new framework for running processes or is it being ported? Is this related to LISA in anyway? (I don't have context - but probably we have similar code somewhere)

cwize1: This is new code. But the API was inspired by the [LISA node](https://github.com/microsoft/lisa/blob/main/lisa/node.py#L244) type. Though, I did simplify some of the params. 

Having stderr and stdout be automatically logged while also collecting them as strings is incredibly useful. It makes debugging test failures a lot easier. Hence, I copied the idea.

In a future change, I will add a very similar API for SSH.

test/vmtests/vmtests/utils/libvirt_vm.py
cwize1: There are actually 2 stop methods that libvirt exposes: graceful and default. So, if were to add a `stop` method, should I expose both or only one of them?

In addition, what should be the exception behavior of the `stop` and `delete` functions? Should they be best effort (like `close` is) or should they fail fast on the first error that is hit?

To be honest, I am not sure this test suite will ever need the ability to `stop` the VM outside of cleanup. So, I am inclined to keep it simple for now. If we ever need the extra functions, we can deal with the complexity then.

test/vmtests/vmtests/utils/libvirt_vm.py
```diff
@@ -0,0 +1,96 @@
...
+        # self.console_logger.attach(domain, console_log_file_path)
+
+        # Start the VM.
+        self.domain.resume()
+
+    # Wait for the VM to boot and then get the IP address.
+    def get_vm_ip_address(self, timeout: float = 30) -> str:
+        timeout_time = time.time() + timeout
+
+        while True:
```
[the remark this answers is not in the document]
cwize1: Fixed.

test/vmtests/vmtests/utils/libvirt_utils.py
cwize1: Fixed.

test/vmtests/vmtests/test_no_change.py
cwize1: Fixed.

gmileka (approved):

PR #30, opened 2024-12-13
VM Tests: Connect to VM using SSH.

cwize1:
Add the ability to connect to VM using SSH so that the contents of the VM can be verified.

This change includes a wrapper around the paramiko library for running commands over SSH. This wrapper's primary job is to live log stderr and stdout while also collecting them as strings.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/test/vmtests/vmtests/test_no_change.py
+++ b/test/vmtests/vmtests/test_no_change.py
@@ -2,8 +2,9 @@
 # Licensed under the MIT License.
 
 import os
+from getpass import getuser
 from pathlib import Path
-from typing import List
+from typing import List, Tuple
 
 import libvirt  # type: ignore
 from docker import DockerClient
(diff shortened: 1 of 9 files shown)


test/vmtests/vmtests/test_no_change.py
```diff
@@ -58,4 +68,20 @@ def test_no_change(
...
+    with SshClient(vm_ip_address, key_path=ssh_private_key_path, known_hosts_path=ssh_known_hosts_path) as vm_ssh:
+        vm_ssh.run("cat /proc/cmdline").check_exit_code()
+
+        os_release_path = test_temp_dir.joinpath("os-release")
+        vm_ssh.get_file(Path("/etc/os-release"), os_release_path)
+
+        with open(os_release_path, "r") as os_release_fd:
+            os_release_text = os_release_fd.read()
+
+            assert "ID=mariner" in os_release_text
```
gmileka, line 86: Curious, why assert on being mariner 2.0?

gmileka: (this code seems to be agnostic of the target os)

gmileka (approved):

test/vmtests/vmtests/test_no_change.py
cwize1: This test is currently running against the `core_efi_azl2` base image. So, asserting the image is Mariner 2.0 here is fine.

At some point, we will need to expand this test suite so that it tests against multiple Azure Linux versions. And we will need to ensure that tests can change their verification logic based on the input image.

PR #33, opened 2024-12-13
OSModifier: Extend EMU API to update verity and root device

elainezhao1:
---
This is a pr that moved from https://github.com/microsoft/azurelinux/pull/10584/files from old repo with comments addressed.

Changes included:
- Add verity and root device to EMU API (Did not make Verity a list like in MIC because Trident logic filters root Verity)
- Add a dedicated verity type for EMU as there are less fields needed in trident (see [here](https://dev.azure.com/mariner-org/ECF/_git/trident?path=/src/engine/boot/grub.rs&version=GBmain&line=182&lineEnd=183&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents ))
- revised EMU GRUB update logic to create bootCustomizer and call WriteToFile just once

### **Checklist**

### DIFF
--- a/toolkit/tools/osmodifierapi/verity.go
+++ b/toolkit/tools/osmodifierapi/verity.go
@@ -1,0 +1,53 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package osmodifierapi
+
+import (
+	"fmt"
+	"regexp"
+
+	"github.com/microsoft/azurelinux/toolkit/tools/imagecustomizerapi"
+)
+
+var (
+	verityNameRegex = regexp.MustCompile("^[a-z]+$")
+)
+
+type Verity struct {
+	// ID is used to correlate `Verity` objects with `FileSystem` objects.
+	Id string `yaml:"id"`
+	// The name of the mapper block device.
+	// Must be 'root' for the rootfs (/) filesystem.
+	Name string `yaml:"name"`
+	// The ID of the 'Partition' to use as the data partition.
+	DataDeviceId string `yaml:"dataDeviceId"`
+	// The ID of the 'Partition' to use as the hash partition.
+	HashDeviceId string `yaml:"hashDeviceId"`
+	// How to handle corruption.
+	CorruptionOption imagecustomizerapi.CorruptionOption `yaml:"corruptionOption"`
+}
+
+func (v *Verity) IsValid() error {
+	if v.Id == "" {
+		return fmt.Errorf("'id' may not be empty")
+	}
+
+	if !verityNameRegex.MatchString(v.Name) {
+		return fmt.Errorf("invalid 'name' value (%s)", v.Name)
+	}
+
+	if v.DataDeviceId == "" {
+		return fmt.Errorf("'dataDeviceId' may not be empty")
+	}
+
+	if v.HashDeviceId == "" {
+		return fmt.Errorf("'hashDeviceId' may not be empty")
+	}
+
+	if err := v.CorruptionOption.IsValid(); err != nil {
+		return fmt.Errorf("invalid corruptionOption:\n%w", err)
+	}
+
+	return nil
+}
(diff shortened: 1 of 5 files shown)


toolkit/tools/osmodifierapi/verity.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+type Verity struct {
+	// ID is used to correlate `Verity` objects with `FileSystem` objects.
+	Id string `yaml:"id"`
+	// The name of the mapper block device.
+	// Must be 'root' for the rootfs (/) filesystem.
+	Name string `yaml:"name"`
+	// The ID of the 'Partition' to use as the data partition.
+	DataDeviceId string `yaml:"dataDeviceId"`
```
cwize1, line 24: Rename to `DataDevice` and `HashDevice`.

(The `Id` suffix refers to the `Id` property on `Partition` type.)

toolkit/tools/pkg/osmodifierlib/modifierutils.go
```diff
@@ -37,43 +37,80 @@ func doModifications(baseConfigPath string, osConfig *osmodifierapi.OS) error {
 		return err
 	}
 
-	if osConfig.Overlays != nil {
-		bootCustomizer, err := imagecustomizerlib.NewBootCustomizer(dummyChroot)
-		if err != nil {
-			return err
-		}
+	err = imagecustomizerlib.AddKernelCommandLine(osConfig.KernelCommandLine.ExtraCommandLine, dummyChroot)
```
cwize1, line 40: You could call `bootCustomizer.AddKernelCommandLine` here.

cwize1 (changes requested):

fintelia:
Might be worth checking whether the differences between MIC and Trident's verity handling are because Trident doesn't need all the same functionality, or if they're partly things that Trident just hasn't implemented yet. For instance, I know that /usr verity is something that we've talked about wanting for Trident, though that would be instead of root verity so there'd still only be one verity device.

toolkit/tools/osmodifierapi/verity.go
elainezhao1: hmm, I did this to align with the verity struct in MIC API, do I need to rename them?

cwize1: This is a slightly different concept than what is currently in the MIC API.

In the MIC API, `DataDeviceId` is a reference to the `Id` field on the `Partition` type. Whereas in this API, `DataDeviceId` contains a string like `"PARTUUID=..."`.

In MIC, we will eventually add an equivalent API. But when we do, we will name it `DataDevice`. Hence, the name suggestion.

elainezhao1:
> read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.

@microsoft-github-policy-service agree company="Microsoft"

elainezhao1:
> Might be worth checking whether the differences between MIC and Trident's verity handling are because Trident doesn't need all the same functionality, or if they're partly things that Trident just hasn't implemented yet. For instance, I know that /usr verity is something that we've talked about wanting for Trident, though that would be instead of root verity so there'd still only be one verity device.

Thanks for bringing it up. For now, this change is just focused on supporting root verity since it fits the current needs. Once /usr verity is supported, I'll revisit this and make adjustments. Should be easy and quick to update the code when the time comes.

toolkit/tools/osmodifierapi/verity.go
```diff
@@ -0,0 +1,53 @@
...
+	if v.Id == "" {
+		return fmt.Errorf("'id' may not be empty")
+	}
+
+	if !verityNameRegex.MatchString(v.Name) {
+		return fmt.Errorf("invalid 'name' value (%s)", v.Name)
+	}
+
+	if v.DataDevice == "" {
+		return fmt.Errorf("'dataDeviceId' may not be empty")
```
cwize1, line 41: `dataDevice` and `hashDeviceId` in error messages.

cwize1 (approved):

toolkit/tools/osmodifierapi/verity.go
elainezhao1: will fix in future pr

PR #34, opened 2024-12-13
OSModifier: Set enforcing=0 for EMU selinux permissive

elainezhao96:
This pr includes changes originally made in https://github.com/microsoft/azurelinux/pull/10584/files from old repo.

There was an issue https://dev.azure.com/mariner-org/ECF/_workitems/edit/9648/ found in trident test pipeline and back then the solution is explicitly set enforcing=0 for selinux permissive for EMU. 

So if Selinux is `permissive` and passed to EMU API, the grub.cfg will have `security=selinux selinux=1 enforcing=0`, EMU would not touch /etc/selinux/config

---

### **Checklist**


toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -55,6 +55,9 @@ const (
 	// CmdlineSELinuxEnforcingArg is the arg required for forcing SELinux to be in enforcing mode.
 	CmdlineSELinuxEnforcingArg = "enforcing=1"
 
+	// CmdlineSELinuxPermissiveArg is the arg for SELinux to be in force-permissive mode.
+	CmdlineSELinuxPermissiveArg = "enforcing=0"
```
elainezhao96, line 59: re [comment](https://github.com/microsoft/azurelinux/pull/10584/files#r1844721925) , now we moved to the new repo, do we still ask for permission to modify code in this file?

cwize1: Nope

toolkit/tools/pkg/imagecustomizerlib/grubcfgutils.go
```diff
@@ -590,25 +590,38 @@ func GrubArgsToString(args []string) string {
...
-		newSELinuxArgs = []string{installutils.CmdlineSELinuxDisabledArg}
-
+		return []string{installutils.CmdlineSELinuxDisabledArg}, nil
 	case imagecustomizerapi.SELinuxModeForceEnforcing:
-		newSELinuxArgs = []string{installutils.CmdlineSELinuxSecurityArg, installutils.CmdlineSELinuxEnabledArg,
-			installutils.CmdlineSELinuxEnforcingArg}
-
+		return []string{installutils.CmdlineSELinuxSecurityArg, installutils.CmdlineSELinuxEnabledArg,
+			installutils.CmdlineSELinuxEnforcingArg}, nil
 	case imagecustomizerapi.SELinuxModePermissive, imagecustomizerapi.SELinuxModeEnforcing:
```
cwize1, line 601: Would it make sense to add a new SELinux mode called `SELinuxModeForcePermissive`, which is the same as `SELinuxModePermissive` but adds `enforcing=0`?

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/grubcfgutils.go
elainezhao96: Sounds good. Can I make this a follow-up change so it can be included in EMU to get the Trident pipeline passing?

cwize1: Sure

cwize1 (approved):

PR #40, opened 2024-12-17
Add initial COSI support 

elainezhao96:
PR summary:

This is a phase 1 COSI support, verity images support is not included.  Enable COSI output, which involves steps like converting vhdx image to raw, extract partition images, generate metadata, and package everything into the final COSI tarball.

some local testing
with command `sudo ./imagecustomizer --log-level debug  --build-dir ./build  --image-file core-3.0.20241213.vhdx   --output-image-file ./output-image-test.cosi --output-image-format cosi --config-file ../pkg/imagecustomizerlib/testdata/newpartitionsuuids-config.yaml`

image
 
---

### **Checklist**
- [ ] Tests added/updated


toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -0,0 +1,199 @@
...
+	MountPoint          string
+	FsType              string
+	FsUuid              string
+	OsReleasePath       *string
+	GrubCfgPath         *string
+	ContainsRpmDatabase bool
+}
+
+func (ex ExpectedImage) ShouldMount() bool {
+	return ex.OsReleasePath != nil || ex.GrubCfgPath != nil || ex.ContainsRpmDatabase
```
elainezhao96, line 36: trident has [code](https://dev.azure.com/mariner-org/ECF/_git/trident?path=/tools/cmd/mkcosi/variants/common.go&version=GBmain&line=361&lineEnd=379&lineStartColumn=1&lineEndColumn=3&lineStyle=plain&_a=contents ) to extract this info. Do I need to add logic to mount the partition and search for os release file path or grub config path?

frhuelsz: That code is a temporary COSI generator (mkcosi), it's not part of trident itself. I'm not an expert on MIC internals, but from the outside it would seem easier to extract this info when MIC has the full image mounted instead of doing it after the fact, like mkcosi does. 
The mkcosi tool is just doing the best it can given the circumstances, so I don't recommend copying too much from it. :) <sup>(I wrote it with the full intention of throwing it away, so it is by no means good or well-designed code.)</sup>

elainezhao96: right, I excluded trident's mount logic. I should remove this function but focus on extracting the values. what is ContainsRpmDatabase and do you think it is something we can obtain?

frhuelsz: In mkcosi it signals that the partition contains `/var/lib/rpm` (where the RPM database is located), this should not be needed in MIC since the full mount tree is actually constructed for the chroot, so you don't really need to track which partition has what.
mkcosi's inputs are the partition images produced by MIC, I chose to be lazy and pre-define things such as `ContainsRpmDatabase` instead of doing the work of finding it because mkcosi only supports two image variants.

frhuelsz: In general, the whole concept of "expected image" from mkcosi may be superfluous in MIC.

cwize1 (changes requested):

docs/imagecustomizer/cli.md
```diff
@@ -31,7 +31,7 @@ The file path to write the final customized image to.
 
 The image format of the the final customized image.
 
-Options: vhd, vhd-fixed, vhdx, qcow2, raw, and iso.
+Options: vhd, vhd-fixed, vhdx, qcow2, raw, iso, and cosi.
 
 At least one of `--output-image-format` and `--output-split-partitions-format` is
```
romoh, line 36: Do we plan to deprecate `output-split-partitions-format` in this PR or a separate one?

cwize1: I think we should have 1-2 releases where we support both APIs. That will make it easier for customers to switch at their own pace.

docs/imagecustomizer/cli.md
```diff
@@ -31,7 +31,7 @@ The file path to write the final customized image to.
 
 The image format of the the final customized image.
 
-Options: vhd, vhd-fixed, vhdx, qcow2, raw, and iso.
+Options: vhd, vhd-fixed, vhdx, qcow2, raw, iso, and cosi.
```
romoh, line 34: we should probably expand a bit more on what cosi is in a separate markdown file

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
elainezhao96: removed. Thanks!

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -0,0 +1,199 @@
...
+	data.Metadata.Image.Sha384 = sha384
+
+	tmpFile, err := os.Open(data.Source)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open uncompressed file %s: %w", data.Source, err)
+	}
+
+	defer tmpFile.Close()
+
+	stat, err = tmpFile.Stat()
```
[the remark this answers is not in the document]
elainezhao96: Thanks for pointing it out! Add an `UncompressedSize` field to `outputPartitionMetadata` and assign it during `extractPartitions`

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -0,0 +1,199 @@
...
+type ExtractedImageData struct {
+	OsRelease string
+	GrubCfg   string
+}
+
+func buildCosiFile(sourceDir string, outputFile string, expectedImages []ExpectedImage) error {
+	metadata := MetadataJson{
+		Version: "1.0",
+		OsArch:  "x86_64",
+		Id:      uuid.New().String(),
```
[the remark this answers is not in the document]
elainezhao96: updated customizeOSContents to return imageUuidStr and imageUuid

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -461,13 +470,151 @@ func convertImageFile(inputPath string, outputPath string, format string) error
...
+		}
+
+		// Convert UUID string to [UuidSize]byte
+		uuidArray, err := stringToUuidArray(uuidStr)
+		if err != nil {
+			return fmt.Errorf("failed to convert UUID to array for partition %s: %w", partitionDevPath, err)
+		}
+
+		// Extract partitions to output files
+		err = extractPartitions(imageLoopback.DevicePath(), outputDir, outputBasename, outputSplitPartitionsFormat, uuidArray)
```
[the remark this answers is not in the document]
elainezhao96: Thanks for clarifying! Fixed.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -461,13 +470,151 @@ func convertImageFile(inputPath string, outputPath string, format string) error
...
+	}
+
+	return partitions, nil
+}
+
+func constructExpectedImages(partitions []outputPartitionMetadata) []ExpectedImage {
+	var expectedImages []ExpectedImage
+
+	for _, partition := range partitions {
+		var partType PartitionType
```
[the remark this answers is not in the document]
elainezhao96: Updated the code to use the fields from `outputPartitionMetadata` and removed PartitionType.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -461,13 +470,151 @@ func convertImageFile(inputPath string, outputPath string, format string) error
...
+	if err != nil {
+		return fmt.Errorf("failed to cleanly close loopback: %w", err)
+	}
+
+	return nil
+}
+
+func convertToCosi(ic *ImageCustomizerParameters) error {
+	logger.Log.Infof("Extracting partition files")
+	err := extractPartitionsForCosi(ic.rawImageFile, ic.outputImageDir, ic.outputImageBase, "raw-zst")
```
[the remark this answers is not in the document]
elainezhao96: moved them to build/cosimages/

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -351,72 +352,72 @@ func customizeOSContents(ic *ImageCustomizerParameters) error {
...
 	partitionsCustomized, newRawImageFile, partIdToPartUuid, err := customizePartitions(ic.buildDirAbs,
 		ic.configPath, ic.config, ic.rawImageFile)
 	if err != nil {
-		return err
+		return [UuidSize]byte{}, "", err
 	}
 	ic.rawImageFile = newRawImageFile
 
 	// Create a uuid for the image
 	imageUuid, imageUuidStr, err := createUuid()
```
cwize1, line 367: I think the code might be a little simpler if the call to `createUuid` was moved into `createImageCustomizerParameters`, adding `imageUuid` and `imageUuidStr` to `ImageCustomizerParameters`.

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -0,0 +1,188 @@
...
+
+func convertToCosi(ic *ImageCustomizerParameters, imageUuid [UuidSize]byte, imageUuidStr string) error {
+	logger.Log.Infof("Extracting partition files")
+	outputDir := filepath.Join(ic.buildDir, "cosiimages")
+	err := os.MkdirAll(outputDir, os.ModePerm)
+	if err != nil {
+		return fmt.Errorf("failed to create folder %s:\n%w", outputDir, err)
+	}
+
+	imageLoopback, err := safeloopback.NewLoopback(ic.rawImageFile)
```
cwize1, line 34: Call `imageLoopback.CleanClose` at end of function.

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -0,0 +1,188 @@
...
+func buildCosiFile(sourceDir string, outputFile string, expectedImages []outputPartitionMetadata, imageUuidStr string) error {
+	metadata := MetadataJson{
+		Version: "1.0",
+		OsArch:  runtime.GOARCH,
+		Id:      imageUuidStr,
+		Images:  make([]Image, len(expectedImages)),
+	}
+
+	if len(expectedImages) == 0 {
+		return errors.New("no images to build")
```
cwize1, line 64: Use `fmt.Errorf` instead of `errors.New`.

`errors.New` is used for global error objects, which are used when you want to support error type checking.

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -0,0 +1,188 @@
...
+	if _, err := io.Copy(sha384, file); err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("%x", sha384.Sum(nil)), nil
+}
+
+func populateMetadata(data ImageBuildData) error {
+	stat, err := os.Stat(data.Source)
+	if err != nil {
+		return fmt.Errorf("filed to stat %s: %w", data.Source, err)
```
cwize1, line 174: `:\n%w` not `: %w`.

Same comment elsewhere.

cwize1 (changes requested):

docs/imagecustomizer/cli.md
elainezhao96: Will add a doc for cosi in next phase 2 pr

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -0,0 +1,192 @@
+package imagecustomizerlib
+
+import (
+	"archive/tar"
+	"crypto/sha512"
+	_ "embed"
```
cwize1, line 6: Is `embed` import needed?

cwize1 (approved):

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
elainezhao96: oops it should be removed. I'll do it in follow up pr

PR #41, opened 2024-12-17
split main sections

romoh:
Docs update :
* Split top level sections of the configuration documentation to its own file.
* Update githib pages navigation menu
---

### **Checklist**
- [ ] Tests added/updated
- [x] Documentation updated (if needed)
- [x] Code conforms to style guidelines

PR #43, opened 2024-12-18
Split up configuration doc.

cwize1:
Currently the configuration doc is long and difficult to navigate. This change splits it up so that each object type gets its own page.

This change is mostly cut/paste, with only some minor formatting fixes in a few places.

---

### **Checklist**
- [x] Tests added/updated


cwize1:
Changes pulled in #41

PR #53, opened 2024-12-20
use rootdir instead of chroot for image release file

amritakohli:
In the function to add release file, we pass in chroot and find its root dir, which consequently requires us to initialize a chroot in its test. Since the only purpose of passing in chroot is to get its root dir, we can simplify the function (and test) by passing in the root dir directly.

---

### **Checklist**
- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/releasefile_test.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/releasefile_test.go
@@ -11,7 +11,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/microsoft/azurelinux/toolkit/tools/internal/safechroot"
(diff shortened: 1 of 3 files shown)


cwize1 (approved):

toolkit/tools/pkg/imagecustomizerlib/releasefile_test.go
```diff
@@ -21,23 +20,19 @@ func TestAddCustomizerRelease(t *testing.T) {
 	}
 
 	proposedDir := filepath.Join(tmpDir, "TestAddCustomizerRelease")
```
cwize1, line 22: Might be good to rename `proposedDir` to something else.

PR #59, opened 2025-01-01
Add json schema generation tool

romoh:
Add a cli that generates image customizer API json schema. 
The generated schema file can be leveraged by other tools for API dependency and config validation.

Here's an example using it in VSCode for syntax validation:
image

### **Checklist**
- [x] Tests added/updated


toolkit/tools/imagecustomizerschemacli/main.go
```diff
@@ -0,0 +1,38 @@
...
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/invopop/jsonschema"
+	"github.com/microsoft/azurelinux/toolkit/tools/imagecustomizerapi"
+)
+
+func main() {
+	if len(os.Args) < 1 {
```
liulanze, line 17: Is the `os.Args[0]` the binary name itself? Should here be:
```
if len(os.Args) < 2 {
...
}
```

toolkit/tools/imagecustomizerschemacli/main.go
```diff
@@ -0,0 +1,38 @@
...
+		os.Exit(1)
+	}
+
+	reflector := jsonschema.Reflector{}
+	reflector.RequiredFromJSONSchemaTags = true
+	schema := reflector.Reflect(&imagecustomizerapi.Config{})
+
+	schemaJSON, err := json.MarshalIndent(schema, "", "  ")
+	if err != nil {
+		log.Fatalf("Failed to marshal schema: %v", err)
```
liulanze, line 28: Question, why using `log.Fatalf`? Usually `fmt.Errorf()` is the tool for error handling.

toolkit/tools/imagecustomizerschemacli/go.mod
```diff
@@ -0,0 +1,26 @@
+module jsonschemacli
```
cwize1, line 1: Please don't add a new `go.mod` file. It doesn't cause any harm to include this as part of the existing `go.mod` file.

toolkit/tools/imagecustomizerschemacli/Makefile
```diff
@@ -0,0 +1,21 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+BINARY_NAME := bin/imagecustomizerschemacli
+GO_FILES := $(wildcard *.go)
```
cwize1, line 5: Go already has good incremental builds. So, it is ok to just use a `.PHONY` rule and avoid this complexity.

toolkit/tools/imagecustomizerschemacli/Makefile
```diff
@@ -0,0 +1,21 @@
...
+	go build -o $(BINARY_NAME) .
+
+run: build
+	@echo "Running $(BINARY_NAME) with $(SCHEMA_FILE)"
+	./$(BINARY_NAME) $(SCHEMA_FILE)
+
+clean:
+	rm -rf bin
+
+.PHONY: all build run clean
```
cwize1, line 21: I think it is better to have separate `.PHONY` lines for each command, so that the `.PHONY` can be placed in the same location as the command itself. This makes it a lot easier to see that a command is a `.PHONY`.

toolkit/tools/imagecustomizerapi/partitionsize.go
```diff
@@ -66,6 +67,13 @@ func (s *PartitionSize) UnmarshalJSON(data []byte) error {
 	return parseAndSetPartitionSize(stringValue, s)
 }
 
+func (PartitionSize) JSONSchema() *jsonschema.Schema {
```
cwize1, line 70: You'll probably want a `JSONSchema` function on each type that has a `UnmarshalYAML` function.

toolkit/tools/imagecustomizerschemacli/main.go
cwize1, line 17: It might be goodness to use a proper args parsing library, in case we ever want to expand the args in the future.

toolkit/tools/imagecustomizerschemacli/Makefile
```diff
@@ -0,0 +1,21 @@
...
+
+BINARY_NAME := bin/imagecustomizerschemacli
+GO_FILES := $(wildcard *.go)
+SCHEMA_FILE := schema.json
+
+all: build run
+
+build: $(GO_FILES)
+	mkdir -p bin
+	go build -o $(BINARY_NAME) .
```
cwize1, line 12: Does this command need to be built? We could just use `go run`.

toolkit/tools/imagecustomizerschemacli/Makefile
```diff
@@ -0,0 +1,21 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+BINARY_NAME := bin/imagecustomizerschemacli
+GO_FILES := $(wildcard *.go)
+SCHEMA_FILE := schema.json
```
cwize1, line 6: Should this `schema.json` be checked-in?

toolkit/tools/imagecustomizerschemacli/main.go
cwize1, line 28: You should move most of the logic into a helper function that returns a normal `error`. Then handle the `log.Fatalf` in a single place.

cwize1 (changes requested):

toolkit/tools/imagecustomizerschemacli/Makefile
romoh: Yes, I'm planning to but will do that in a separate change. The other change will include adding a github workflow to ensure this file is kept up to date with any change/PR.

toolkit/tools/imagecustomizerschemacli/go.mod
romoh: I thought to make this independent of the existing tools, but no strong opinion.

cwize1: Some of the Go tooling doesn't like multiple `go.mod` files.

toolkit/tools/imagecustomizerschemacli/main.go
romoh: binary name doesn't count

toolkit/tools/imagecustomizerapi/disksize.go
```diff
@@ -51,6 +52,13 @@ func (s *DiskSize) UnmarshalJSON(data []byte) error {
 	return parseAndSetDiskSize(stringValue, s)
 }
 
+func (DiskSize) JSONSchema() *jsonschema.Schema {
+	return &jsonschema.Schema{
+		Type:    "string",
+		Pattern: `^\d+[KMGT]?$`,
```
cwize1, line 58: `^\d+[KMGT]$`

The suffix is currently not optional.

toolkit/tools/imagecustomizerapi/partitionsize.go
```diff
@@ -66,6 +67,13 @@ func (s *PartitionSize) UnmarshalJSON(data []byte) error {
...
+	return &jsonschema.Schema{
+		Type:    "string",
+		Pattern: `^\d+[KMGT]?$|^grow`,
```
cwize1, line 73: Missing `$` at end of regex?

cwize1, line 73: Does `^(\d+[KMGT]?|grow)$` work?

toolkit/tools/imagecustomizerschemacli/Makefile
cwize1: Remove `GO_FILES`?

toolkit/tools/imagecustomizerschemacli/main.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	schema := reflector.Reflect(&imagecustomizerapi.Config{})
+	schemaJSON, err := json.MarshalIndent(schema, "", "  ")
+	if err != nil {
+		return fmt.Errorf("failed to marshal schema: %w", err)
+	}
+
+	// Write schema to file
+	if err := os.WriteFile(outputFile, schemaJSON, 0644); err != nil {
```
cwize1, line 42: nit: `0o644`

Just to be explicit that this is an octal.

toolkit/.gitignore
```diff
@@ -1,14 +0,0 @@
-/out/
```
cwize1, line 1: Removing this `.gitignore` file seems like an unrelated change.

toolkit/tools/go.mod
```diff
@@ -29,15 +29,26 @@ require (
 	gopkg.in/yaml.v3 v3.0.1
 )
 
+require (
```
cwize1, line 32: The changes to this `go.mod` file seem very odd. Typically, there are only 2 `require` sections: 1 for direct dependencies and 1 for indirect dependencies.

I recommend reverting the changes to this file and then running `go mod tidy` to pick up the new dependencies.

toolkit/tools/go.mod
```diff
@@ -29,15 +29,26 @@ require (
...
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.11.1 // indirect
 	github.com/Azure/azure-sdk-for-go/sdk/internal v1.8.0 // indirect
 	github.com/AzureAD/microsoft-authentication-library-for-go v1.2.2 // indirect
+	github.com/alecthomas/jsonschema v0.0.0-20220216202328-9eeeec9d044b
 	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751 // indirect
 	github.com/alecthomas/units v0.0.0-20190924025748-f65c72e2690d // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/gdamore/encoding v1.0.0 // indirect
 	github.com/golang-jwt/jwt/v5 v5.2.1 // indirect
+	github.com/invopop/jsonschema v0.13.0
```
cwize1, line 51: `jsonschema` is a direct dependency. So, it probably shouldn't be included in the list of indirect dependencies.

toolkit/tools/imagecustomizerschemacli/main.go
```diff
@@ -0,0 +1,47 @@
...
+
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/alecthomas/kingpin"
```
cwize1, line 12: `"gopkg.in/alecthomas/kingpin.v2"` for consistency with the other executables.

cwize1 (changes requested):

toolkit/.gitignore
romoh: Included by mistake. Will send a separate PR.

toolkit/tools/go.mod
romoh: I need to rebase as well, will clean it once.

romoh: weird.. I had to remove the whole section for it to regenerate properly.

cwize1 (approved):

PR #64, opened 2025-01-03
Wire verity signature parameters for the uki flow + update container build

gmileka:
---

### **Checklist**
- [ ] Tests added/updated


romoh (approved):
Since this is feature branch, it is ok
Before merging to main, let's discuss where to put samples. At least a common top dire under docs

PR #65, opened 2025-01-06
Remove script capability restrictions.

cwize1:
In change [d60055e](https://github.com/microsoft/azure-linux-image-tools/commit/d60055ea36b4131ecc6626c8bbaaca6a9381fe08), user scripts were given a restricted set of capabilities. However, it seems that the capabilities given were not enough for some legitimate scenarios. So, remove this restriction for now while we re-think the design.

---

### **Checklist**

- [x] Tests added/updated


romoh (changes requested):
You need to remove or comment out the tests otherwise they would block a full e2e build/test

romoh (approved):

PR #66, opened 2025-01-07
Better handle filesystem optional features.

cwize1:
Filesystems (e.g. ext4, xfs) sometimes receive new features. When these features are enabled, anything that loads the filesystem (e.g. kernel, grub, etc.) must support the new feature or it will refuse to load it.

When formatting a new partition, there are a few considerations for if a filesystem feature should be enabled:

- Does the version of mkfs support that feature?
- Does the build host kernel support that feature?
- Does the target OS support that feature?

This change ensure that all these considerations are handled correctly for the ext4 and xfs filesystem types.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/imagegen/diskutils/diskutils.go
+++ b/toolkit/tools/imagegen/diskutils/diskutils.go
@@ -800,7 +800,8 @@ func formatSinglePartition(targetOs targetos.TargetOs, partDevPath string, parti
 
 		mkfsOptions, err := getFileSystemOptions(targetOs, fsType)
 		if err != nil {
-			err = fmt.Errorf("failed to get mkfs args for filesystem type (%s):\n%w", fsType, err)
+			err = fmt.Errorf("failed to get mkfs args for filesystem type (%s) and target os (%s):\n%w", fsType,
+				targetOs, err)
 			return fsType, err
 		}
 
--- a/toolkit/tools/imagegen/diskutils/diskutils.go
+++ b/toolkit/tools/imagegen/diskutils/diskutils.go
@@ -20,21 +20,11 @@ import (
 	"github.com/microsoft/azurelinux/toolkit/tools/internal/logger"
 	"github.com/microsoft/azurelinux/toolkit/tools/internal/retry"
 	"github.com/microsoft/azurelinux/toolkit/tools/internal/shell"
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/targetos"
 	"github.com/sirupsen/logrus"
 )
 
 var (
-	// When calling mkfs, the default options change depending on the host OS you are running on and typically match
-	// what the distro has decided is best for their OS. For example, for ext2/3/4, the defaults are stored in
-	// /etc/mke2fs.conf.
-	// However, when building Azure Linux images, the defaults should be as consistent as possible and should only contain
-	// features that are supported on Azure Linux.
-	DefaultMkfsOptions = map[string][]string{
-		"ext2": {"-b", "4096", "-O", "none,sparse_super,large_file,filetype,resize_inode,dir_index,ext_attr"},
-		"ext3": {"-b", "4096", "-O", "none,sparse_super,large_file,filetype,resize_inode,dir_index,ext_attr,has_journal"},
-		"ext4": {"-b", "4096", "-O", "none,sparse_super,large_file,filetype,resize_inode,dir_index,ext_attr,has_journal,extent,huge_file,flex_bg,metadata_csum,64bit,dir_nlink,extra_isize"},
-	}
-
 	partedVersionRegex = regexp.MustCompile(`^parted \(GNU parted\) (\d+)\.(\d+)`)
 
 	// The default partition name used when the version of `parted` is too old (<3.5).
@@ -433,7 +423,7 @@ func WaitForDevicesToSettle() error {
 }
 
 // CreatePartitions creates partitions on the specified disk according to the disk config
-func CreatePartitions(diskDevPath string, disk configuration.Disk, rootEncryption configuration.RootEncryption,
+func CreatePartitions(targetOs targetos.TargetOs, diskDevPath string, disk configuration.Disk, rootEncryption configuration.RootEncryption,
 	diskKnownToBeEmpty bool,
 ) (partDevPathMap map[string]string, partIDToFsTypeMap map[string]string, encryptedRoot EncryptedRootDevice, err error) {
 	const timeoutInSeconds = "5"
@@ -492,7 +482,7 @@ func CreatePartitions(diskDevPath string, disk configuration.Disk, rootEncryptio
 			return partDevPathMap, partIDToFsTypeMap, encryptedRoot, err
 		}
 
-		partFsType, err := FormatSinglePartition(partDevPath, partition)
+		partFsType, err := formatSinglePartition(targetOs, partDevPath, partition)
 		if err != nil {
 			err = fmt.Errorf("failed to format partition:\n%w", err)
 			return partDevPathMap, partIDToFsTypeMap, encryptedRoot, err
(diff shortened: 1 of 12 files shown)


toolkit/tools/imagegen/diskutils/filesystem.go
```diff
@@ -0,0 +1,316 @@
...
+		Features: []string{"sparse_super", "large_file", "filetype", "resize_inode", "dir_index", "ext_attr",
+			"has_journal", "extent", "huge_file", "flex_bg", "metadata_csum", "64bit", "dir_nlink", "extra_isize",
+		},
+	}
+
+	azl3Ext4Options = ext4Options{
+		BlockSize: 4096,
+		Features: []string{"sparse_super", "large_file", "filetype", "resize_inode", "dir_index", "ext_attr",
+			"has_journal", "extent", "huge_file", "flex_bg", "metadata_csum", "64bit", "dir_nlink", "extra_isize",
+			"orphan_file",
```
liulanze, line 52: Question, why was `orphan_file` added to Azl3? Perhaps adding a clue comment explaining this diff between Azl2 and Azl3 for ext4.

cwize1: Fixed.

liulanze (approved):

toolkit/tools/internal/targetos/targetos.go
```diff
@@ -0,0 +1,69 @@
...
+			continue
+		}
+		name := split[0]
+		value := split[1]
+
+		value = strings.TrimPrefix(value, "\"")
+		value = strings.TrimSuffix(value, "\"")
+
+		fields[name] = value
+	}
```
PawelWMS, line 42: Just an idea: this looks like a great candidate to be extracted to a separate helper function for reading `/etc/os-release` into a map.

PawelWMS: You could also use this Go module, although it's not super fresh: https://github.com/acobaugh/osrelease.

There's also this one: https://pkg.go.dev/github.com/quay/claircore/osrelease. Seems fresher (last commit a week ago): https://github.com/quay/claircore/tree/main/osrelease.

toolkit/tools/internal/kernelversion/kernelversion.go
```diff
@@ -0,0 +1,56 @@
...
+	// Examples:
+	//   OS               Version
+	//   Fedora 40        6.11.6-200.fc40.x86_64
+	//   Ubuntu 22.04     6.8.0-48-generic
+	//   Azure Linux 2.0  5.15.153.1-2.cm2
+	//   Azure Linux 3.0  6.6.47.1-1.azl3
+	kernelVersionRegex = regexp.MustCompile(`^(\d+)\.(\d+)\.(\d+)([.\-][a-zA-Z0-9_.\-]*)?$`)
+)
+
+func GetBuildHostKernelVersion() (version.Version, error) {
```
PawelWMS, line 28: You could also use `GetKernelVersion` from this package: https://pkg.go.dev/github.com/moby/moby/pkg/parsers/kernel#GetKernelVersion.

toolkit/tools/imagegen/diskutils/filesystem.go
```diff
@@ -0,0 +1,324 @@
...
+			// Feature is not supported by version of mkfs.ext4.
+			logger.Log.Infof("mkfs.ext4 does not support ext4 feature (%s)", feature)
+			continue
+		}
+
+		features = append(features, feature)
+	}
+
+	// "none" requests no default options.
+	featuresArg := "none," + strings.Join(features, ",")
```
PawelWMS, line 216: Just checking: in the unlikely event that `features` is empty, will the trailing comma be an issue or will the tool just ignore it?

toolkit/tools/internal/kernelversion/kernelversion.go
cwize1: This is simple enough that I don't think it is worth taking an external dependency. Though, I did switch to using the `uname` function directly.

toolkit/tools/imagegen/diskutils/filesystem.go
cwize1: Fixed.

toolkit/tools/internal/targetos/targetos.go
cwize1: I split the `os-release` parsing into its own package.

I also figured I'd used the robust grub/Bash parser I've already written.

toolkit/tools/imagegen/diskutils/diskutils.go
```diff
@@ -433,7 +423,7 @@ func WaitForDevicesToSettle() error {
 }
 
 // CreatePartitions creates partitions on the specified disk according to the disk config
-func CreatePartitions(diskDevPath string, disk configuration.Disk, rootEncryption configuration.RootEncryption,
+func CreatePartitions(targetOs targetos.TargetOs, diskDevPath string, disk configuration.Disk, rootEncryption configuration.RootEncryption,
```
mfrw, line 426: An opinion here, since this is a new repo, there is a chance that as time goes on, we might end up adding more and more options here.
How about we have a struct with default values, and on top of that struct we implement a method `CreateParititions`.
This might help us in only modifying the code at a single place where we have the initial struct and the rest of the code might not need any modifications :)
Just an opinion

cwize1: Most of the options for this function are included in the `Disk` struct. So, it is probably unlikely that this function will expand over time. Also, this function is only called in 4 places. So, IMHO, this function is probably ok as it is.

toolkit/tools/imagegen/diskutils/diskutils.go
```diff
@@ -804,12 +794,16 @@ func FormatSinglePartition(partDevPath string, partition configuration.Partition
...
 	case "fat32", "fat16", "vfat", "ext2", "ext3", "ext4", "xfs":
-		mkfsOptions := DefaultMkfsOptions[fsType]
-
 		if fsType == "fat32" || fsType == "fat16" {
 			fsType = "vfat"
 		}
 
+		mkfsOptions, err := getFileSystemOptions(targetOs, fsType)
+		if err != nil {
+			err = fmt.Errorf("failed to get mkfs args for filesystem type (%s):\n%w", fsType, err)
```
romoh, line 803: nit - maybe it is worth logging the targetOS here too

toolkit/tools/internal/targetos/targetos.go
```diff
@@ -0,0 +1,51 @@
...
+	case "mariner":
+		switch versionId {
+		case "2.0":
+			return TargetOsAzureLinux2, nil
+
+		default:
+			return "", fmt.Errorf("unknown VERSION_ID (%s) for CBL-Mariner in /etc/os-release", versionId)
+		}
+
+	case "azurelinux":
```
romoh, line 39: hmm I have mixed feelings about this. Do we want to have a default behavior instead of erroring out of unsupported versions? (this can also happen at the caller, but here seems more central)
My concern is around maintainability. For a new OS version, someone will need to update this flow.

romoh (commented):
Few comments

toolkit/tools/imagegen/diskutils/filesystem.go
```diff
@@ -0,0 +1,324 @@
...
+		"rmapbt":     "metadata",
+		"sparse":     "inode",
+		"nrext64":    "inode",
+	}
+
+	// The maximum version of mkfs.xfs that is currently supported.
+	// This is used to prevent issues with newer versions of mkfs.xfs default enabling new features.
+	maxMkfsXfsVersion = version.Version{6, 9}
+
+	// The minmum supported kernel version. This helps avoid versions complexity for features that are old and therefore
```
amritakohli, line 136: nit: minimum*

toolkit/tools/internal/targetos/targetos.go
cwize1: I think erroring out is probably a good thing. When a new distro/version is released, we will want to add a new set of filesystem options for that distro/version.

I guess in theory we could have some sort of universal default. That could be more interesting, if and when we decide to support other distros. But I don't think it is needed for now.

romoh: My thinking is that a dev working on azl 4.0 upbringing will be blocked on this.

toolkit/tools/imagegen/diskutils/filesystem.go
cwize1: Fixed

toolkit/tools/imagegen/diskutils/diskutils.go
cwize1: Fixed.

romoh (approved):
I won't block on my comment for now, but I think we should revisit the default behavior logic in a separate change.

PR #78, opened 2025-01-09
Check-in and enforce API schema update

romoh:
The change checks-in the generated API schema and adds a github workflow to enforce its update.
A test workflow https://github.com/microsoft/azure-linux-image-tools/actions/runs/12700031766/job/35401914843?pr=78
 
Note this is a ⚠️ Breaking Change ⚠️ for dev flows. 
Documentation added for how to update the schema. For reference, run `make -C toolkit/tools/imagecustomizerschemacli/`

---

### **Checklist**


toolkit/tools/imagecustomizerschemacli/schema.json
```diff
@@ -0,0 +1,578 @@
+{
```
cwize1, line 1: Should this file live in the `imagecustomizerapi` directory?

docs/imagecustomizer/developer-guide.md
```diff
@@ -12,6 +12,9 @@ Run:
 sudo make -C ./toolkit go-imagecustomizer
 ```
 
+If you're updating the image customizer api as part of your change, do update
```
cwize1, line 15: I generally like to place commands to run in multiline codeblocks. IMHO, this makes it easier to find and copy the commands.

cwize1 (changes requested):

toolkit/tools/imagecustomizerschemacli/schema.json
romoh: I went with updating the makefile to keep the command simple.

cwize1 (approved):

PR #79, opened 2025-01-10
Use 'just-the-docs' theme for GitHub pages.

cwize1:
Replace the GitHub pages Jekyll theme with 'just-the-docs'. This theme has an inbuilt navigation bar. It also seems to have good browser width handling.

In addition, add a Gemfile and Makefile for the docs generation.

---

### **Checklist**

- [x] Tests added/updated


romoh (changes requested):
You need to fix the github workflow. 
I think you need to switch to different github actions to deploy the new theme.

romoh (approved):

docs/Gemfile.lock
```diff
@@ -0,0 +1,315 @@
+GEM
```
romoh, line 1: We don't really need to check this one in. The Gemfile is sufficient to generate it. We can add it to .gitignore

cwize1: In general, it is considered goodness to check-in package lock files since it means that everyone will be using exactly the same versions of all the dependencies. It also means that you can control when you take updates to packages.

romoh (approved):

PR #83, opened 2025-01-13
Image Customizer - Support modifying Verity protected base image.

liulanze:
---

### **Checklist**
- [x] Tests added/updated


toolkit/tools/imagecustomizerapi/extendedmountidentifiertype.go
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Microsoft Corporation.
```
cwize1, line 1: This is an internal type. So, it should live in `imagecustomizerlib` directory.

toolkit/tools/internal/file/file.go
```diff
@@ -235,6 +236,21 @@ func Append(data string, dst string) (err error) {
...
 
+// AppendLines appends multiple lines to the end of a file.
+func AppendLines(lines []string, dst string) error {
+	for _, line := range lines {
+		// Check if the line already ends with a newline character.
+		if !strings.HasSuffix(line, "\n") {
+			// Add newline if missing.
+			line += "\n"
+		}
+		if err := Append(line, dst); err != nil {
```
cwize1, line 247: `Append` opens and closes the file. So, this function is very inefficient. You should ensure the file is only opened once.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -318,6 +321,22 @@ func findSourcePartitionHelper(source string,
...
+
+	partition, partitionIndex, err := findExtendedPartition(mountIdType, mountId, partitions, buildDir)
+	if err != nil {
+		return imagecustomizerapi.ExtendedMountIdentifierTypeDefault, diskutils.PartitionInfo{}, 0, err
+	}
+
+	return mountIdType, partition, partitionIndex, nil
+}
+
 func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId string,
```
cwize1, line 340: Can `findPartition` be implemented using `findExtendedPartition` to avoid duplicate code?

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -318,6 +321,22 @@ func findSourcePartitionHelper(source string,
 	return mountIdType, partition, partitionIndex, nil
 }
 
+func findExtendedSourcePartitionHelper(source string, partitions []diskutils.PartitionInfo,
```
cwize1, line 324: Can `findSourcePartitionHelper` be implemented using `findExtendedSourcePartitionHelper` to reduce duplicate code?

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -352,6 +371,136 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
...
+		matches := false
+		switch mountIdType {
+		case imagecustomizerapi.ExtendedMountIdentifierTypeUuid:
+			matches = partition.Uuid == mountId
+		case imagecustomizerapi.ExtendedMountIdentifierTypePartUuid:
+			matches = partition.PartUuid == mountId
+		case imagecustomizerapi.ExtendedMountIdentifierTypePartLabel:
+			matches = partition.PartLabel == mountId
+		case imagecustomizerapi.ExtendedMountIdentifierTypeDev:
+			matches, err := findExtendedDevPartition(partition, partitions, buildDir)
```
cwize1, line 388: You should only do the conversion from /dev path to UUID a single time. That is, `findExtendedDevPartition` should be pulled outside of the loop.

And if you are feeling fancy, you could parse the value of `systemd.verity_root_data` and then feed it back into `findSourcePartitionHelper`.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -352,6 +371,136 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
...
+	err = bootPartitionMount.CleanClose()
+	if err != nil {
+		return false, fmt.Errorf("failed to close bootPartitionMount: %w", err)
+	}
+
+	return false, nil
+}
+
+func checkVerityRootPartUUID(partition diskutils.PartitionInfo, args string) (bool, error) {
+	argsParts := strings.Split(args, " ")
```
cwize1, line 494: Please use the `ParseCommandLineArgs` function for parsing args.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -352,6 +371,136 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
...
+			if matches {
+				matchedPartitionIndexes = append(matchedPartitionIndexes, i)
+			}
+		}
+		if matches {
+			matchedPartitionIndexes = append(matchedPartitionIndexes, i)
+		}
+	}
+
+	if len(matchedPartitionIndexes) < 1 {
```
jiria, line 401: nit: == 0 is more accurate, no?

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -352,6 +371,136 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
...
+	}
+	if len(matchedPartitionIndexes) > 1 {
+		err := fmt.Errorf("too many matches for partition found (%s=%s)", mountIdType, mountId)
+		return diskutils.PartitionInfo{}, 0, err
+	}
+
+	partitionIndex := matchedPartitionIndexes[0]
+	partition := partitions[partitionIndex]
+
+	return partition, partitionIndex, nil
```
jiria, line 413: Why do we need to return both?

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -352,6 +371,136 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
...
+}
+
+func findExtendedDevPartition(partition diskutils.PartitionInfo, partitions []diskutils.PartitionInfo,
+	buildDir string,
+) (bool, error) {
+	espPartition, err := findSystemBootPartition(partitions)
```
jiria, line 419: SystemBootPartition name is a bit confusing here

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -352,6 +371,136 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
...
+	if err != nil {
+		return false, err
+	}
+
+	bootPartition, err := findBootPartitionFromEsp(espPartition, partitions, buildDir)
```
jiria, line 424: Is this function not expected to work for UKIs/systemd-boot?

jiria: Or are we expecting that grub config is always going to be present?

jiria: And are we expecting the boot partition to be always present?

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -352,6 +371,136 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
...
+	if err != nil {
+		return false, err
+	}
+
+	// Temporarily mount the boot partition so that the grub config file can be read.
+	tmpDirBoot := filepath.Join(buildDir, tmpBootPartitionDirName)
+	bootPartitionMount, err := safemount.NewMount(bootPartition.Path, tmpDirBoot, bootPartition.FileSystemType, 0, "", true)
```
gmileka, line 431: This mount should be read-only. Mounting partitions non-read-only modifies it unintentionally.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -352,6 +371,136 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
...
+	if err != nil {
+		return false, fmt.Errorf("failed to mount boot partition (%s):\n%w", bootPartition.Path, err)
+	}
+	defer bootPartitionMount.Close()
+
+	// Mount ESP partition to check for UKI.
+	tmpDirEsp := filepath.Join(buildDir, tmpEspParitionDirName)
+	espPartitionMount, err := safemount.NewMount(espPartition.Path, tmpDirEsp, espPartition.FileSystemType, 0, "", true)
```
gmileka, line 439: same here, read-only is needed

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -352,6 +371,136 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
...
+	defer espPartitionMount.Close()
+
+	// Check if there is any UKI images.
+	espLinuxPath := filepath.Join(tmpDirEsp, UkiOutputDir)
+	ukiFiles, err := filepath.Glob(filepath.Join(espLinuxPath, "vmlinuz-*.efi"))
+	if err != nil {
+		return false, fmt.Errorf("failed to search for UKI images in ESP partition:\n%w", err)
+	}
+
+	if len(ukiFiles) > 0 {
```
gmileka, line 452: should this part be part of a bootloader abstraction?

gmileka: (it's bootloader specific logic, in a partition utility code)

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -352,6 +371,136 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
...
+		return diskutils.PartitionInfo{}, 0, err
+	}
+
+	partitionIndex := matchedPartitionIndexes[0]
+	partition := partitions[partitionIndex]
+
+	return partition, partitionIndex, nil
+}
+
+func findExtendedDevPartition(partition diskutils.PartitionInfo, partitions []diskutils.PartitionInfo,
```
gmileka, line 416: This function seems to be specific to finding verity partition (uki/non-uki ends up calling `checkVerityRootPartUUID`), should we rename to indicate its purpose? I'm not sure why it's named `ExtendedDevPartition`?)

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
fintelia, line 431: Shouldn't this only be mounted if the UKI check below doesn't pass? Seems like it isn't needed if the cmdline can be extracted from a UKI

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -352,6 +371,136 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
...
+
+	err = bootPartitionMount.CleanClose()
+	if err != nil {
+		return false, fmt.Errorf("failed to close bootPartitionMount: %w", err)
+	}
+
+	return false, nil
+}
+
+func checkVerityRootPartUUID(partition diskutils.PartitionInfo, args string) (bool, error) {
```
gmileka, line 493: nit: maybe rename to `hasVerityRootPartUUID` ? or `isEnabledVerityRoot`?

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -372,6 +521,30 @@ func parseSourcePartition(source string) (imagecustomizerapi.MountIdentifierType
 	return imagecustomizerapi.MountIdentifierTypeDefault, "", err
 }
 
+func parseExtendedSourcePartition(source string) (imagecustomizerapi.ExtendedMountIdentifierType, string, error) {
```
gmileka, line 524: can we update `parseSourcePartition` instead of creating a new function?
are there scenarios where the new `/dev` should not be handled and an error must be still returned?

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -222,7 +222,7 @@ func systemdFormatPartitionId(configDeviceId string, mountIdType imagecustomizer
 ) (string, error) {
 	partUuid := partIdToPartUuid[configDeviceId]
 
-	partition, _, err := findPartition(imagecustomizerapi.MountIdentifierTypePartUuid, partUuid, partitions)
+	partition, _, err := findPartition(imagecustomizerapi.MountIdentifierTypePartUuid, partUuid, partitions, "")
```
cwize1, line 225: Why is `buildDir` set to `""` here?

toolkit/tools/pkg/imagecustomizerlib/extendedmountidentifiertype.go
```diff
@@ -0,0 +1,41 @@
...
+	ExtendedMountIdentifierTypePartLabel ExtendedMountIdentifierType = "part-label"
+
+	// ExtendedMountIdentifierTypeDev mounts this partition via a device.
+	ExtendedMountIdentifierTypeDev ExtendedMountIdentifierType = "dev"
+
+	// ExtendedMountIdentifierTypeDefault uses the default type, which is PARTUUID.
+	ExtendedMountIdentifierTypeDefault ExtendedMountIdentifierType = ""
+)
+
+func (e ExtendedMountIdentifierType) IsValid() error {
```
cwize1, line 32: Is this `IsValid` function actually used?

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -362,15 +363,6 @@ func customizeOSContents(ic *ImageCustomizerParameters) error {
 		ic.config.OS = &imagecustomizerapi.OS{}
 	}
 
-	// Check if the partition is using DM_verity_hash file system type.
-	// The presence of this type indicates that dm-verity has been enabled on the base image. If dm-verity is not enabled,
-	// the verity hash device should not be assigned this type. We do not support customization on verity enabled base
-	// images at this time because such modifications would compromise the integrity and security mechanisms enforced by dm-verity.
-	err := checkDmVerityEnabled(ic.rawImageFile)
```
cwize1, line 369: Delete the `checkDmVerityEnabled` function?

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -293,44 +295,108 @@ func isSpecialPartition(fstabEntry diskutils.FstabEntry) bool {
...
 	}
 
 	return partition.Path, nil
 }
 
-func findSourcePartitionHelper(source string,
-	partitions []diskutils.PartitionInfo,
+func findSourcePartitionHelper(source string, partitions []diskutils.PartitionInfo,
+	buildDir string,
 ) (imagecustomizerapi.MountIdentifierType, diskutils.PartitionInfo, int, error) {
```
cwize1, line 311: Looking at the users of the `findSourcePartitionHelper` function, I think you can change this function to return `ExtendedMountIdentifierType` instead of `MountIdentifierType`. That way, you don't need to do the conversion.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -352,24 +418,138 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
...
+		return false, fmt.Errorf("failed to close bootPartitionMount: %w", err)
+	}
+
+	return false, nil
+}
+
+func checkVerityRootPartUUID(partition diskutils.PartitionInfo, args string) (bool, error) {
+	argsParts := strings.Split(args, " ")
+	for _, part := range argsParts {
+		if strings.HasPrefix(part, "systemd.verity_root_data=PARTUUID=") {
```
liulanze, line 501: TODO: remove the hardcoded PARTUUID.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
cwize1: Ideally, we should detect the verity settings in the base image and re-enable verity. Except when the user customizes the partitions, then we should throw away the verity settings in the base image.

toolkit/tools/pkg/imagecustomizerlib/extendedmountidentifiertype.go
cwize1: Why? Adding a test for code that isn't being used, seems a little silly.

The `IsValid` functions exist because we want to verify the values provided by the user in the config file. But if the type is only ever used internally, then we know that the values will always be valid.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -890,34 +891,34 @@ func customizeVerityImageHelper(buildDir string, baseConfigPath string, config *
...
-		return err
+		return false, err
 	}
 
 	for i := range diskPartitions {
 		diskPartition := diskPartitions[i]
 
 		if diskPartition.FileSystemType == "DM_verity_hash" {
-			return fmt.Errorf("cannot customize base image that has dm-verity enabled")
+			return true, nil
```
cwize1, line 912: You shouldn't early return here. You should ensure `loopback.CleanClose()` is called.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -293,43 +295,94 @@ func isSpecialPartition(fstabEntry diskutils.FstabEntry) bool {
...
 	}
 
 	return partition.Path, nil
 }
 
-func findSourcePartitionHelper(source string,
-	partitions []diskutils.PartitionInfo,
-) (imagecustomizerapi.MountIdentifierType, diskutils.PartitionInfo, int, error) {
-	mountIdType, mountId, err := parseSourcePartition(source)
+func findSourcePartitionHelper(source string, partitions []diskutils.PartitionInfo,
```
cwize1, line 309: `findSourcePartitionHelper` and `findExtendedSourcePartitionHelper` are the same function.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -352,24 +405,156 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
...
 }
 
+func convertDevToUuid(partitions []diskutils.PartitionInfo, buildDir string) (string, error) {
+	for _, partition := range partitions {
+		matches, err := checkExtendedDevPartition(partition, partitions, buildDir)
+		if err != nil {
+			return "", err
+		}
+		if matches {
+			return partition.Uuid, nil
```
cwize1, line 415: It would probably be better to use `partition.PartUuid` here, since that is always available (for GPT disks) regardless of the filesystem being used.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -352,24 +405,156 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
...
+	}
+	return "", fmt.Errorf("unable to find UUID for /dev path")
+}
+
+// checkExtendedDevPartition checks whether a given partition is associated with a verity-enabled root partition.
+// This function is specific to the case where the partition type is "dev" (device path, e.g., /dev/mapper/root).
+// It is an extension of the findExtendedPartition logic and specializes in handling verity partitions,
+// supporting both UKI and non-UKI configurations by validating against the kernel cmdline arguments or grub configuration.
+// Returns true if the partition matches the expected verity configuration, otherwise false.
+func checkExtendedDevPartition(partition diskutils.PartitionInfo, partitions []diskutils.PartitionInfo,
```
cwize1, line 426: This function should be split up. You should have a function that extracts the kernel command-line, correctly picking between pulling from either UKI and grub.cfg (and eventually systemd-boot). Then another function that extracts the verity info from the kernel command-line

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
liulanze: fixed.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
liulanze: Revised, could you please check it again? Thanks.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -412,103 +399,111 @@ func convertDevToUuid(partitions []diskutils.PartitionInfo, buildDir string) (st
...
+
+	err = bootPartitionMount.CleanClose()
+	if err != nil {
+		return "", fmt.Errorf("failed to close bootPartitionMount:\n%w", err)
 	}
 
 	err = espPartitionMount.CleanClose()
 	if err != nil {
-		return false, fmt.Errorf("failed to close bootPartitionMount: %w", err)
+		return "", fmt.Errorf("failed to close bootPartitionMount:\n%w", err)
```
cwize1, line 493: nit: `espPartitionMount`

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -352,24 +392,164 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
 	return partition, partitionIndex, nil
 }
 
+func convertDevToUuid(partitions []diskutils.PartitionInfo, buildDir string) (string, error) {
```
cwize1, line 395: The `extractVerityRootPartUUID` function is returning the partition's UUID. So, the for-loop inside the `convertDevToUuid` function is pointless.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -352,24 +392,164 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
...
+	defer bootPartitionMount.Close()
+
+	grubCfgPath := filepath.Join(tmpDirBoot, "/grub2/grub.cfg")
+	kernelToArgs, err := extractKernelToArgsFromGrub(grubCfgPath)
+	if err != nil {
+		return "", fmt.Errorf("failed to extract kernel arguments from grub.cfg:\n%w", err)
+	}
+
+	var combinedArgs []string
+	for _, args := range kernelToArgs {
```
cwize1, line 482: Shouldn't you only use the args from a single kernel version, instead of all of them?

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -352,24 +392,164 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
...
+	err = espPartitionMount.CleanClose()
+	if err != nil {
+		return "", fmt.Errorf("failed to close bootPartitionMount:\n%w", err)
+	}
+
+	return strings.Join(combinedArgs, " "), nil
+}
+
+func extractVerityRootPartUUID(cmdline string) (string, error) {
+	argsParts := strings.Split(cmdline, " ")
```
cwize1, line 500: You should be using the utilities in `grubcfgutils.go` to parse the `grub.cfg` file.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -352,24 +392,164 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
...
+	if err != nil {
+		return "", fmt.Errorf("failed to close bootPartitionMount:\n%w", err)
+	}
+
+	err = espPartitionMount.CleanClose()
+	if err != nil {
+		return "", fmt.Errorf("failed to close bootPartitionMount:\n%w", err)
+	}
+
+	return strings.Join(combinedArgs, " "), nil
```
cwize1, line 496: Use `GrubArgsToString` to combine grub args into a single string.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -352,24 +392,164 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
...
+}
+
+func extractVerityRootPartUUID(cmdline string) (string, error) {
+	argsParts := strings.Split(cmdline, " ")
+	for _, part := range argsParts {
+		if strings.HasPrefix(part, "systemd.verity_root_data=PARTUUID=") {
```
cwize1, line 502: You should extract the full value of `systemd.verity_root_data=`, instead of hard coding the `PARTUUID=` part. Then use the `parseExtendedSourcePartition` to parse the value.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -352,24 +392,164 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
...
+
+	tmpDirBoot := filepath.Join(buildDir, tmpBootPartitionDirName)
+	bootPartitionMount, err := safemount.NewMount(bootPartition.Path, tmpDirBoot, bootPartition.FileSystemType, unix.MS_RDONLY, "", true)
+	if err != nil {
+		return "", fmt.Errorf("failed to mount boot partition (%s):\n%w", bootPartition.Path, err)
+	}
+	defer bootPartitionMount.Close()
+
+	grubCfgPath := filepath.Join(tmpDirBoot, "/grub2/grub.cfg")
+	kernelToArgs, err := extractKernelToArgsFromGrub(grubCfgPath)
```
cwize1, line 476: You should be using the `getLinuxCommandLineArgs` and `findKernelCommandLineArgValue` functions here.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -352,24 +392,164 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
...
+		}
+
+		cmdlineContent, err := os.ReadFile(cmdlinePath)
+		if err != nil {
+			return "", fmt.Errorf("failed to read kernel cmdline args from dumped file:\n%w", err)
+		}
+
+		err = espPartitionMount.CleanClose()
+		if err != nil {
+			return "", fmt.Errorf("failed to close bootPartitionMount:\n%w", err)
```
cwize1, line 462: nit: `espPartitionMount`

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -352,24 +392,164 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
...
+		}
+
+		return string(cmdlineContent), nil
+	}
+
+	tmpDirBoot := filepath.Join(buildDir, tmpBootPartitionDirName)
```
cwize1, line 468: You should create a `extractKernelCmdlineFromGrub` function and call it from `extractKernelCmdline`. This will hopefull make the code a little easier to read. It might also make it easier to de-duplicate the call to `espPartitionMount.CleanClose()`.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -362,14 +363,14 @@ func customizeOSContents(ic *ImageCustomizerParameters) error {
...
-	// The presence of this type indicates that dm-verity has been enabled on the base image. If dm-verity is not enabled,
-	// the verity hash device should not be assigned this type. We do not support customization on verity enabled base
-	// images at this time because such modifications would compromise the integrity and security mechanisms enforced by dm-verity.
-	err := checkDmVerityEnabled(ic.rawImageFile)
+	// Check if dm-verity is enabled on the base image.
+	verityEnabled, err := isDmVerityEnabled(ic.rawImageFile)
 	if err != nil {
 		return err
 	}
+	if verityEnabled && len(ic.config.Storage.Verity) == 0 {
```
cwize1, line 371: I think this should be changed to: `verityEnabled  && !ic.customizeOSPartitions`

Currently, you can't specify verity without customizing partitions.

In the future, if the base image is verity-enabled and partitions aren't customized, then MIC should automatically re-apply the verity settings from the base image.

Also in the future, if the base image has the partitions to support verity but isn't enabled for verity, then we will add the ability to enable verity.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -347,24 +389,207 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
 	return partition, partitionIndex, nil
 }
 
+func convertDevToPartUuid(partitions []diskutils.PartitionInfo, buildDir string) (string, error) {
+	for _, partition := range partitions {
```
cwize1, line 393: You do not need this for-loop. Just pass the results of `extractVerityRootPartitionId` up to `findExtendedPartition` and let `findExtendedPartition`'s for-loop find the partition.

Finding the same partition twice in a row is silly.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -347,24 +389,207 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to read kernel cmdline args from dumped file:\n%w", err)
+	}
+
+	tokens, err := grub.TokenizeConfig(string(cmdlineContent))
+	if err != nil {
+		return nil, fmt.Errorf("failed to tokenize kernel command-line from UKI: %w", err)
+	}
+
+	args, err := ParseCommandLineArgs(tokens)
```
cwize1, line 507: You should use `getLinuxCommandLineArgs` instead of `ParseCommandLineArgs`.

`getLinuxCommandLineArgs` searches for the `linux` command and then parse the kernel args. `ParseCommandLineArgs` only parses the kernel args.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -372,14 +373,14 @@ func customizeOSContents(ic *ImageCustomizerParameters) error {
...
-	// the verity hash device should not be assigned this type. We do not support customization on verity enabled base
-	// images at this time because such modifications would compromise the integrity and security mechanisms enforced by dm-verity.
-	err := checkDmVerityEnabled(ic.rawImageFile)
+	// Check if dm-verity is enabled on the base image.
+	verityEnabled, err := isDmVerityEnabled(ic.rawImageFile)
 	if err != nil {
 		return err
 	}
+	if verityEnabled && !ic.customizeOSPartitions {
+		return fmt.Errorf("dm-verity is enabled on the base image. To customize a verity-enabled base image, the verity section must be reconfigured.")
```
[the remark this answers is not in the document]
liulanze: Fixed.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -372,14 +373,14 @@ func customizeOSContents(ic *ImageCustomizerParameters) error {
...
-	// The presence of this type indicates that dm-verity has been enabled on the base image. If dm-verity is not enabled,
-	// the verity hash device should not be assigned this type. We do not support customization on verity enabled base
-	// images at this time because such modifications would compromise the integrity and security mechanisms enforced by dm-verity.
-	err := checkDmVerityEnabled(ic.rawImageFile)
+	// Check if dm-verity is enabled on the base image.
+	verityEnabled, err := isDmVerityEnabled(ic.rawImageFile)
 	if err != nil {
 		return err
 	}
+	if verityEnabled && !ic.customizeOSPartitions {
```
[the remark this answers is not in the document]
liulanze: No, it won't have the verity, do you want me to involve this logic into this PR?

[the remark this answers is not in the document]
cwize1: No, that's the behavior I expect. Thanks.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -347,24 +389,207 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
...
+
+	cmdline, err := extractKernelCmdlineFromUki(tmpDirEsp, buildDir)
+	if err == nil {
+		return cmdline, nil
+	} else if !errors.Is(err, os.ErrNotExist) {
+		return nil, err
+	}
+
+	tmpDirBoot := filepath.Join(buildDir, tmpBootPartitionDirName)
+	bootPartitionMount, err := safemount.NewMount(bootPartition.Path, tmpDirBoot, bootPartition.FileSystemType, unix.MS_RDONLY, "", true)
```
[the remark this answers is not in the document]
liulanze: Fixed.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -347,24 +389,207 @@ func findPartition(mountIdType imagecustomizerapi.MountIdentifierType, mountId s
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to mount ESP partition (%s):\n%w", espPartition.Path, err)
+	}
+	defer espPartitionMount.Close()
+
+	cmdline, err := extractKernelCmdlineFromUki(tmpDirEsp, buildDir)
+	if err == nil {
+		return cmdline, nil
+	} else if !errors.Is(err, os.ErrNotExist) {
+		return nil, err
```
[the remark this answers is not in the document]
liulanze: Fixed.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
liulanze: There will not be able to search `linux <kernel>` from UKI, due to trim the first two strings (linux and its kernel) when building the UKI.

cwize1: Oh, sorry. I misread the code.

You are already using `getLinuxCommandLineArgs` in `extractKernelCmdlineFromGrub`. And you are correctly using `ParseCommandLineArgs` in `extractKernelCmdlineFromUki`.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
cwize1: Looks good.

cwize1 (approved):

PR #85, opened 2025-01-14
Add Phase 2 COSI Support

elainezhao96:
PR summary:

This is a phase 2 COSI support, we should now be able to generate cosi file from a **_non-verity_** base image with a verity config, and the verity information will be included in the cosi metadata file, so it will align with expected sample format

local test:
tested with command `sudo ./imagecustomizer --log-level debug  --build-dir ./build  --image-file core-3.0.20241220.vhdx --output-image-file ./output-image-test.cosi --output-image-format cosi --config-file ../pkg/imagecustomizerlib/testdata/verity-config.yaml`
unpacked the .cosi file
image
and obtained output/metadata.json

```
{
  "version": "1.0",
  "osArch": "amd64",
  "images": [
    {
      "image": {
        "path": "images/output-image-test_1.raw.zst",
        "compressedSize": 988946,
        "uncompressedSize": 8388608,
        "sha384": "6c2a78a9bc84624d90fdf2b0be7e33dd4abd305f8ae2467a4868bdf3ed8114cbe72c8512e5836e0a806e932732f79585"
      },
      "mountPoint": "",
      "fsType": "vfat",
      "fsUuid": "1163-42B5",
      "partType": "c12a7328-f81f-11d2-ba4b-00a0c93ec93b",
      "verity": null
    },
    {
      "image": {
        "path": "images/output-image-test_2.raw.zst",
        "compressedSize": 51437979,
        "uncompressedSize": 1064304640,
        "sha384": "f6594fa455d3feffd002dddd32afc843856e09bb850c6a3a6ec84366a00c8dde92e28f68a464ca62e8160bb5ad6a50e2"
      },
      "mountPoint": "",
      "fsType": "ext4",
      "fsUuid": "14e039a8-61d2-4d43-85b2-f26c92326c03",
      "partType": "0fc63daf-8483-4772-8e79-3d69d8477de4",
      "verity": null
    },
    {
      "image": {
        "path": "images/output-image-test_3.raw.zst",
        "compressedSize": 115309290,
        "uncompressedSize": 2147483648,
        "sha384": "29d8663dbd8f320ad9bba8169173837213eb973842f52ec774af8840f130c596b118cef88c2a4c38bc8820cfba482637"
      },
      "mountPoint": "",
      "fsType": "ext4",
      "fsUuid": "393703a2-b8a9-4574-8fd0-d961674c3efa",
      "partType": "0fc63daf-8483-4772-8e79-3d69d8477de4",
      "verity": {
        "image": {
          "path": "images/output-image-test_4.raw.zst",
          "compressedSize": 2908417,
          "uncompressedSize": 134217728,
          "sha384": "d4c8dd1357868b9aab271af8f74c52f05c8e4378d53637644cec43fc7a20adbb0aa1f0b01d701555cfda0490205d11ab"
        },
        "hash": "d5ff43f8d03a738b1e25f4eeb1b1f71441fa456daf60a664f64a2e95316ec65f"
      }
    },
    {
      "image": {
        "path": "images/output-image-test_5.raw.zst",
        "compressedSize": 28775661,
        "uncompressedSize": 2012217344,
        "sha384": "9a69d82c2e1c0b6681191cbcc6847ba8db2b54f15671b0b9a00e1649ce753038a8f2cb2b6d7e1bc36bdcc9cc8d0dbbf1"
      },
      "mountPoint": "",
      "fsType": "ext4",
      "fsUuid": "8fee3677-b8ba-432b-8add-3f336d11575f",
      "partType": "0fc63daf-8483-4772-8e79-3d69d8477de4",
      "verity": null
    }
  ],
  "osRelease": "",
  "id": "85e18600-1160-1b53-bcea-d93a3f0d848b"
```

Note: 
A verity base image currently is not supported, will wait customizations on verity images to be enabled. Also, will have a follow-up pr to fix the missing mountpoints.

---

### **Checklist**
- [ ] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
@@ -3,7 +3,6 @@ package imagecustomizerlib
 import (
 	"archive/tar"
 	"crypto/sha512"
-	_ "embed"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -17,9 +16,10 @@ import (
 )
 
 type ImageBuildData struct {
-	Source    string
-	KnownInfo outputPartitionMetadata
-	Metadata  *Image
+	Source       string
+	KnownInfo    outputPartitionMetadata
+	Metadata     *Image
+	VeritySource string
 }
 
 func convertToCosi(ic *ImageCustomizerParameters) error {
@@ -41,9 +41,9 @@ func convertToCosi(ic *ImageCustomizerParameters) error {
 		return err
 	}
 
-	err = buildCosiFile(outputDir, ic.outputImageFile, partitionMetadataOutput, ic.imageUuidStr)
+	err = buildCosiFile(outputDir, ic.outputImageFile, partitionMetadataOutput, ic.verityMetadata, ic.imageUuidStr)
 	if err != nil {
-		return fmt.Errorf("failed to build COSI:\n%w", err)
+		return fmt.Errorf("failed to build COSI file:\n%w", err)
 	}
 
 	logger.Log.Infof("Successfully converted to COSI: %s", ic.outputImageFile)
@@ -56,45 +56,90 @@ func convertToCosi(ic *ImageCustomizerParameters) error {
 	return nil
 }
 
-func buildCosiFile(sourceDir string, outputFile string, expectedImages []outputPartitionMetadata, imageUuidStr string) error {
+func buildCosiFile(sourceDir string, outputFile string, partitions []outputPartitionMetadata, verityMetadata []verityDeviceMetadata, imageUuidStr string) error {
+
+	// Pre-compute a map for quick lookup of partition metadata by UUID
+	partUuidToMetadata := make(map[string]outputPartitionMetadata)
+	for _, partition := range partitions {
+		partUuidToMetadata[partition.PartUuid] = partition
+	}
+
+	// Pre-compute a set of verity hash UUIDs for quick lookup
+	verityHashUuids := make(map[string]struct{})
+	for _, verity := range verityMetadata {
+		verityHashUuids[verity.hashPartUuid] = struct{}{}
+	}
+
+	imageData := []ImageBuildData{}
+
+	for _, partition := range partitions {
+		// Skip verity hash partitions as their metadata will be assigned to the corresponding data partitions
+		if _, isVerityHash := verityHashUuids[partition.PartUuid]; isVerityHash {
+			continue
+		}
+
+		metadataImage := Image{
+			Image: ImageFile{
+				Path:             path.Join("images", partition.PartitionFilename),
+				UncompressedSize: partition.UncompressedSize,
+			},
+			PartType:   partition.PartitionTypeUuid,
+			MountPoint: partition.Mountpoint,
+			FsType:     partition.FileSystemType,
+			FsUuid:     partition.Uuid,
+		}
+
+		imageDataEntry := ImageBuildData{
+			Source:    path.Join(sourceDir, partition.PartitionFilename),
+			Metadata:  &metadataImage,
+			KnownInfo: partition,
+		}
+
+		// Add Verity metadata if the partition has a matching entry in verityMetadata
+		for _, verity := range verityMetadata {
+			if partition.PartUuid == verity.dataPartUuid {
+				hashPartition, exists := partUuidToMetadata[verity.hashPartUuid]
+				if !exists {
+					return fmt.Errorf("missing metadata for hash partition UUID:\n%s", verity.hashPartUuid)
+				}
+
+				metadataImage.Verity = &Verity{
+					Hash: verity.hash,
+					Image: ImageFile{
+						Path:             path.Join("images", hashPartition.PartitionFilename),
+						UncompressedSize: hashPartition.UncompressedSize,
+					},
+				}
+
+				veritySourcePath := path.Join(sourceDir, hashPartition.PartitionFilename)
+				imageDataEntry.VeritySource = veritySourcePath
+				break
+			}
+		}
+
+		imageData = append(imageData, imageDataEntry)
+	}
+
+	// Populate metadata for each image
+	for i := range imageData {
+		err := populateMetadata(&imageData[i])
+		if err != nil {
+			return fmt.Errorf("failed to populate metadata for %s:\n%w", imageData[i].Source, err)
+		}
+
+		logger.Log.Infof("Populated metadata for image %s", imageData[i].Source)
+	}
+
 	metadata := MetadataJson{
 		Version: "1.0",
 		OsArch:  runtime.GOARCH,
 		Id:      imageUuidStr,
-		Images:  make([]Image, len(expectedImages)),
+		Images:  make([]Image, len(imageData)),
 	}
 
-	if len(expectedImages) == 0 {
-		return fmt.Errorf("no images to build")
-	}
-
-	// Create an interim metadata struct to combine the known data with the metadata
-	imageData := make([]ImageBuildData, len(expectedImages))
-	for i, image := range expectedImages {
-		metadata := &metadata.Images[i]
-		imageData[i] = ImageBuildData{
-			Source:    path.Join(sourceDir, image.PartitionFilename),
-			Metadata:  metadata,
-			KnownInfo: image,
-		}
-
-		metadata.Image.Path = path.Join("images", image.PartitionFilename)
-		metadata.PartType = image.PartitionTypeUuid
-		metadata.MountPoint = image.Mountpoint
-		metadata.FsType = image.FileSystemType
-		metadata.FsUuid = image.Uuid
-		metadata.UncompressedSize = image.UncompressedSize
-	}
-
-	// Populate metadata for each image
-	for _, data := range imageData {
-		logger.Log.Infof("Processing image %s", data.Source)
-		err := populateMetadata(data)
-		if err != nil {
-			return fmt.Errorf("failed to populate metadata for %s:\n%w", data.Source, err)
-		}
-
-		logger.Log.Infof("Populated metadata for image %s", data.Source)
+	// Copy updated metadata
+	for i, data := range imageData {
+		metadata.Images[i] = *data.Metadata
 	}
 
 	// Marshal metadata.json
@@ -133,26 +178,43 @@ func buildCosiFile(sourceDir string, outputFile string, expectedImages []outputP
 }
 
 func addToCosi(data ImageBuildData, tw *tar.Writer) error {
-	imageFile, err := os.Open(data.Source)
+	err := addFileToCosi(tw, data.Source, data.Metadata.Image)
 	if err != nil {
-		return fmt.Errorf("failed to open image file:\n%w", err)
+		return fmt.Errorf("failed to add image file to COSI:\n%w", err)
 	}
-	defer imageFile.Close()
+
+	if data.VeritySource != "" && data.Metadata.Verity != nil {
+		err := addFileToCosi(tw, data.VeritySource, data.Metadata.Verity.Image)
+		if err != nil {
+			return fmt.Errorf("failed to add verity file to COSI:\n%w", err)
+		}
+	}
+
+	return nil
+}
+
+func addFileToCosi(tw *tar.Writer, source string, image ImageFile) error {
+	file, err := os.Open(source)
+	if err != nil {
+		return fmt.Errorf("failed to open file :\n%w", err)
+	}
+	defer file.Close()
 
 	err = tw.WriteHeader(&tar.Header{
 		Typeflag: tar.TypeReg,
-		Name:     data.Metadata.Image.Path,
-		Size:     int64(data.Metadata.Image.CompressedSize),
+		Name:     image.Path,
+		Size:     int64(image.CompressedSize),
 		Mode:     0o400,
 		Format:   tar.FormatPAX,
 	})
+
 	if err != nil {
-		return fmt.Errorf("failed to write tar header:\n%w", err)
+		return fmt.Errorf("failed to write tar header for file '%s':\n%w", image.Path, err)
 	}
 
-	_, err = io.Copy(tw, imageFile)
+	_, err = io.Copy(tw, file)
 	if err != nil {
-		return fmt.Errorf("failed to write image to COSI:\n%w", err)
+		return fmt.Errorf("failed to write image '%s' to COSI:\n%w", image.Path, err)
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -240,3 +292,13 @@ func writePartitionMetadataJson(outDir string, name string, output *[]outputPart
 	logger.Log.Infof("Partition metadata file created: %s", fullPath)
 	return nil
 }
+
+func isBootPartition(partition diskutils.PartitionInfo) bool {
+	if partition.PartLabel == "boot" ||
+		partition.PartitionTypeUuid == "c12a7328-f81f-11d2-ba4b-00a0c93ec93b" ||
```
fintelia, line 298: This is the UUID for an EFI System Partition, which is typically mounted at /boot/efi

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -188,5 +202,41 @@ func populateMetadata(data ImageBuildData) error {
...
+	verityFile.CompressedSize = uint64(verityStat.Size())
+
+	veritySha384, err := sha384sum(*source)
+	if err != nil {
+		return fmt.Errorf("failed to calculate sha384 of verity source: %w", err)
+	}
+
+	verityFile.Sha384 = veritySha384
+
+	verityFile.UncompressedSize = uint64(verityStat.Size())
```
frhuelsz, line 239: Isn't this the same as the compressed size?

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
elainezhao96: removed the function as it was not needed

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -188,5 +217,232 @@ func populateMetadata(data ImageBuildData) error {
...
+
+	return nil
+}
+
+func extractPartitionsForCosi(diskPartitions []diskutils.PartitionInfo, outDir string, basename string, partitionFormat string, imageUuid [UuidSize]byte, roothash string, rootPartUUID string, roothashFilepath string) ([]outputPartitionMetadata, error) {
+	// Stores the output partition metadata that will be written to JSON file
+	var partitionMetadataOutput []outputPartitionMetadata
+
+	// Extract partitions to files
+	for _, partition := range diskPartitions {
```
cwize1, line 263: Remove duplication between the `extractPartitionsForCosi` and `extractPartitions` functions.

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -188,5 +217,232 @@ func populateMetadata(data ImageBuildData) error {
...
+	}
+
+	verityFile.Sha384 = veritySha384
+
+	verityFile.UncompressedSize = uint64(verityStat.Size())
+
+	return nil
+}
+
+func extractPartitionsForCosi(diskPartitions []diskutils.PartitionInfo, outDir string, basename string, partitionFormat string, imageUuid [UuidSize]byte, roothash string, rootPartUUID string, roothashFilepath string) ([]outputPartitionMetadata, error) {
```
cwize1, line 258: Please line wrap at 120.

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -188,5 +217,232 @@ func populateMetadata(data ImageBuildData) error {
...
+		// Construct metadata
+		partitionMetadata, err := constructOutputPartitionMetadata(partition, partitionNum, partitionFilepath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to construct partition metadata:\n%w", err)
+		}
+		partitionMetadata.UncompressedSize = uncompressedPartitionFileSize
+
+		// Assign verity information to the root partition
+		if roothash != "" && rootPartUUID != "" && roothashFilepath != "" && partition.PartUuid == rootPartUUID {
+			roothashFilepath, err = extractRawZstPartition(roothashFilepath, imageUuid, partitionFilename, outDir)
```
cwize1, line 328: This code seems odd. Why are you extracting the verity hash partition here? Why not do this in 2 phases: 1. Extract all the partitions and their metadata, 2. Convert the flat list of partition metadata into the tree structure wanted by COSI?

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -36,7 +40,18 @@ func convertToCosi(ic *ImageCustomizerParameters) error {
...
 	defer imageLoopback.Close()
 
-	partitionMetadataOutput, err := extractPartitions(imageLoopback.DevicePath(), outputDir, ic.outputImageBase, "raw-zst", ic.imageUuid)
+	// Get partition info
+	diskPartitions, err := diskutils.GetDiskPartitions(imageLoopback.DevicePath())
+	if err != nil {
+		return err
+	}
+
+	roothash, rootPartUUID, roothashFilepath, err := locateVerityRootInfo(diskPartitions, outputDir, ic.outputImageBase)
```
cwize1, line 49: Instead of trying to extract the verity info from the final image, please plumb the verity info from the `customizeVerityImageHelper` function (which is where the verity info is generated) to the `convertToCosi` function.

Extracting the verity info from the image correctly is a bit of a pain. This is also needed by PR #83. However, PR #83 currently handles extracting the info from UKIs, which this change is missing. Therefore, it is probably easier to just take the verity info from the place that it is being generated.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
elainezhao96: removed the func

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
elainezhao96: removed func extractPartitionsForCosi

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -15,15 +15,16 @@ import (
...
+	PartitionNum      int     `json:"partitionnum"`     // Example: 1
+	PartitionFilename string  `json:"filename"`         // Example: image_1.raw.zst
+	PartLabel         string  `json:"partlabel"`        // Example: boot
+	FileSystemType    string  `json:"fstype"`           // Example: vfat
+	PartitionTypeUuid string  `json:"parttype"`         // Example: c12a7328-f81f-11d2-ba4b-00a0c93ec93b
+	Uuid              string  `json:"uuid"`             // Example: 4BD9-3A78
+	PartUuid          string  `json:"partuuid"`         // Example: 7b1367a6-5845-43f2-99b1-a742d873f590
+	Mountpoint        string  `json:"mountpoint"`       // Example: /mnt/os/boot
+	UncompressedSize  uint64  `json:"uncompressedsize"` // Example: 104857600
+	Verity            *Verity `json:"verity,omitempty"`
```
cwize1, line 27: Please don't add the `Verity` field to this struct. This is the type used by `extractPartitions` to output its metadata. It shouldn't have fields that are populated by other code (e.g. the COSI code).

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -807,85 +812,90 @@ func customizeVerityImageHelper(buildDir string, baseConfigPath string, config *
 	// Extract the partition block device path.
 	dataPartition, err := idToPartitionBlockDevicePath(rootfsVerity.DataDeviceId, diskPartitions, partIdToPartUuid)
 	if err != nil {
-		return err
+		return "", "", err
 	}
 	hashPartition, err := idToPartitionBlockDevicePath(rootfsVerity.HashDeviceId, diskPartitions, partIdToPartUuid)
```
cwize1, line 817: The `customizeVerityImageHelper` function also knows the identity of the hash partition (`rootfsVerity.HashDeviceId`). So, you should plumb its PartUuid to `convertToCosi` as well, instead of having the `buildCosiData` search for the hash partition (`DM_verity_hash`) again.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -82,6 +82,9 @@ type ImageCustomizerParameters struct {
 
 	imageUuid    [UuidSize]byte
 	imageUuidStr string
+
+	rootHash     string
```
cwize1, line 86: At some point we will support multiple verity partitions. So, if possible, it might be good to write this code in anticipation of that.

For example:

```golang
type verityDeviceMetadata struct {
    hash string
    dataPartUuid string
    hashPartUuid string
}

type ImageCustomizerParameters struct {
    ...
    verityMetadata []verityDeviceMetadata
}
```

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
cwize1: +1. You need to pull the `UncompressedSize` from the `[]outputPartitionMetadata` list.

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -188,5 +207,81 @@ func populateMetadata(data ImageBuildData) error {
...
 	data.Metadata.Image.Sha384 = sha384
+
+	err = populateVerityMetadata(data.VeritySource, data.Metadata.Verity)
+	if err != nil {
+		return fmt.Errorf("failed to populate verity metadata: %w", err)
+	}
+	return nil
+}
+
+func populateVerityMetadata(source *string, verity *Verity) error {
```
cwize1, line 218: `populateMetadata` and `populateVerityMetadata` are both populating a `ImageFile` type. You should see if there is a way to deduplicate the code.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
elainezhao96: I feel maybe `func buildCosiData` still need to search for the hash partition because it needs root partition file path (which is generated in `func extractpartitions` and rootfsVerity does not have that info. Correct me if I'm wrong

cwize1: Currently, the `buildCosiData` function identifies the hash partition by `FileSystemType` equaling `DM_verity_hash`. While this works for now, it won't work if there are multiple verity partitions. So, instead you should search for the hash partition by PartUuid.

elainezhao96: right, that is what i plan to do

toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
elainezhao96: addressed by creating a cosiPartitionMetadata struct with verity as a list

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -807,85 +822,102 @@ func customizeVerityImageHelper(buildDir string, baseConfigPath string, config *
...
+
+	// Identify the data partition UUID
+	dataPartUuid, ok := partIdToPartUuid[rootfsVerity.DataDeviceId]
+	if !ok {
+		return "", "", "", fmt.Errorf("failed to determine root partition UUID for DataDeviceId: %s", rootfsVerity.DataDeviceId)
+	}
+
+	// Identify the hash partition UUID
+	var hashPartUuid string
+	for _, partition := range diskPartitions {
```
cwize1, line 864: Why do you have a for-loop here? Why not just use the value `partIdToPartUuid[rootfsVerity.HashDeviceId]` directly?

toolkit/tools/pkg/imagecustomizerlib/cosimetadata.go
```diff
@@ -9,18 +9,17 @@ type MetadataJson struct {
...
-	FsUuid           string    `json:"fsUuid"`
-	PartType         string    `json:"partType"`
-	Verity           *Verity   `json:"verity"`
-	UncompressedSize uint64    `json:"uncompressedSize"`
+	Image      ImageFile `json:"image"`
+	MountPoint string    `json:"mountPoint"`
+	FsType     string    `json:"fsType"`
+	FsUuid     string    `json:"fsUuid"`
+	PartType   string    `json:"partType"`
+	Verity     []*Verity `json:"verity"`
```
cwize1, line 17: `Verity` is not a list. There can be multiple entries in `Images` that have a populated `Verity` field. But there is only one verity entry per image.

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -172,21 +211,119 @@ func sha384sum(path string) (string, error) {
...
+				verityEntry := &Verity{
+					Hash: verity.hash,
+					Image: ImageFile{
+						Path:             metadata.Filename,
+						UncompressedSize: metadata.UncompressedSize,
+					},
+				}
+
+				if cosiPartition.Verity == nil {
+					cosiPartition.Verity = []*Verity{}
```
cwize1, line 319: In Go, `nil` is a valid value for an array. It represents an empty array. So, you don't need to do this.

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -16,10 +15,24 @@ import (
...
+}
+
 type ImageBuildData struct {
-	Source    string
-	KnownInfo outputPartitionMetadata
-	Metadata  *Image
+	Source       string
+	KnownInfo    cosiPartitionMetadata
+	Metadata     *Image
+	VeritySource *string
```
cwize1, line 35: If the empty string (`""`) isn't a valid value, then you should use the empty string to represent that the value is missing, instead of using `*string` and `nil`.

This helps reduce the complexity of the code.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -68,27 +86,48 @@ func buildCosiFile(sourceDir string, outputFile string, expectedImages []outputP
 		return fmt.Errorf("no images to build")
 	}
 
-	// Create an interim metadata struct to combine the known data with the metadata
 	imageData := make([]ImageBuildData, len(expectedImages))
 	for i, image := range expectedImages {
```
cwize1, line 90: Could the code be simplified by merging the logic in `buildCosiData` into `buildCosiFile`? That is, do the matching of verity partitions in `buildCosiFile`? That way, you could go directly from `outputPartitionMetadata` to `ImageBuildData` and the `cosiPartitionMetadata` type wouldn't be needed.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
elainezhao96: fixed

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
elainezhao96: fixed.

toolkit/tools/pkg/imagecustomizerlib/cosimetadata.go
elainezhao96: I see. fixed.

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -139,20 +183,35 @@ func addToCosi(data ImageBuildData, tw *tar.Writer) error {
 	}
 	defer imageFile.Close()
 
-	err = tw.WriteHeader(&tar.Header{
+	addFileToCosi(tw, data.Metadata.Image.Path, int64(data.Metadata.Image.CompressedSize), imageFile)
+
+	if data.VeritySource != "" && data.Metadata.Verity != nil {
+		verityFile, err := os.Open(data.VeritySource)
```
cwize1, line 189: Move the `os.Open` call to `addFileToCosi` so that you can de-duplicate the code.

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -139,20 +183,35 @@ func addToCosi(data ImageBuildData, tw *tar.Writer) error {
...
+			return fmt.Errorf("failed to open verity file:\n%w", err)
+		}
+		defer verityFile.Close()
+		addFileToCosi(tw, data.Metadata.Verity.Image.Path, int64(data.Metadata.Verity.Image.CompressedSize), verityFile)
+	}
+
+	return nil
+}
+
+func addFileToCosi(tw *tar.Writer, name string, size int64, file *os.File) error {
```
cwize1, line 200: You could pass `ImageFile` to `addFileToCosi` instead of `name` and `size`.

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -56,45 +56,89 @@ func convertToCosi(ic *ImageCustomizerParameters) error {
...
+				metadataImage.Verity = &Verity{
+					Hash: verity.hash,
+					Image: ImageFile{
+						Path:             path.Join("images", hashPartition.PartitionFilename),
+						UncompressedSize: hashPartition.UncompressedSize,
+					},
+				}
+
+				veritySourcePath := path.Join(sourceDir, hashPartition.PartitionFilename)
+				imageDataEntry.VeritySource = veritySourcePath
```
cwize1, line 115: Add a `break` here?

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
elainezhao96: revised.

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
elainezhao96: revised

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -133,26 +178,40 @@ func buildCosiFile(sourceDir string, outputFile string, expectedImages []outputP
 }
 
 func addToCosi(data ImageBuildData, tw *tar.Writer) error {
-	imageFile, err := os.Open(data.Source)
+	addFileToCosi(tw, data.Source, data.Metadata.Image)
```
cwize1, line 181: Check error code.

cwize1 (changes requested):

cwize1 (approved):

PR #86, opened 2025-01-14
Split up root README content.

cwize1:
Move the 'Quick Start' guide and the 'Things to Avoid' guide to their own pages.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/docs/imagecustomizer/README.md
+++ b/docs/imagecustomizer/README.md
@@ -5,6 +5,8 @@ nav_order: 2
 
 # Image Customizer
 
+## Overview
+
 The Image Customizer is a tool that can take an existing generic Azure Linux
 image and modify it to be suited for particular scenario.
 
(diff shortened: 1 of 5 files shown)


docs/imagecustomizer/how-to/quick-start.md
```diff
@@ -0,0 +1,70 @@
...
+
+   - For Ubuntu 22.04 images, run:
+
+     ```bash
+     sudo apt -y install qemu-utils rpm coreutils util-linux mount fdisk udev openssl \
+        sed createrepo-c squashfs-tools genisoimage parted e2fsprogs dosfstools \
+        xfsprogs zstd cryptsetup-bin grub2-common
+     ```
+
+   - For Mariner 2.0, run:
```
romoh, line 37: update to `Azure Linux 2.0`

romoh: Should we update to `3.0` actually?

docs/imagecustomizer/README.md
```diff
@@ -31,99 +33,10 @@ Disadvantages:
   So, some customizations aren't possible using the Image Customizer.
```
romoh, line 33: nit - doesn't seem like a good way to start the docs with what can't be done. 
Maybe instead call out what is supported and refer to the docs. and maybe rephrase this to state the fact that `customizations are done offline without booting the image ... `

romoh (changes requested):

docs/imagecustomizer/how-to/quick-start.md
cwize1: The list works for both. So, added both to doc.

romoh (approved):

docs/imagecustomizer/README.md
cwize1: I do want to be somewhat frank with what can and can't be done, as this might help reduce customer surprises/pain.

That being said, I did rephrase this to be less negative while still conveying the same information.

PR #92, opened 2025-01-22
Switch from 'parted' to 'sfdisk'.

cwize1:
Use `sfdisk` and `blockdev` instead of `parted` and `partprobe`.

This is being done for a number of reasons:

1. There is already a dependency on `sfdisk`. So, this will remove a dependency.

2. The API of `sfdisk` is a little better structured (IMHO).

3. Feature support for some GPT features (e.g. partition types) came to `sfdisk` way sooner than they did for `parted`. So, we don't need version workarounds to use slightly older versions of `sfdisk`.

4. Ubuntu has a "fun" patch added to `parted` that adds in a `udevadm settle` call to all `parted` operations. This can cause deadlocks when `parted` is called under `flock`.

In addition, this change updates the partition customization tests so that they also verify the partition sizes.

Also, this change adds `flock` around the `mkfs` call. This will prevent udev from processing the partition while it is being formatted.

In addition, add a wait loop that ensures the kernel's view of the partitions have been actually updated.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/docs/imagecustomizer/how-to/quick-start.md
+++ b/docs/imagecustomizer/how-to/quick-start.md
@@ -21,8 +21,8 @@ nav_order: 6
    [Supported configuration](../api/configuration.md)
 
 3. Install prerequisites: `qemu-img`, `rpm`, `dd`, `lsblk`, `losetup`, `sfdisk`,
-   `udevadm`, `flock`, `blkid`, `openssl`, `sed`, `createrepo`, `mksquashfs`,
-   `genisoimage`, `parted`, `mkfs`, `mkfs.ext4`, `mkfs.vfat`, `mkfs.xfs`, `fsck`,
+   `blockdev` `udevadm`, `flock`, `blkid`, `openssl`, `sed`, `createrepo`, `mksquashfs`,
+   `genisoimage`, `mkfs`, `mkfs.ext4`, `mkfs.vfat`, `mkfs.xfs`, `fsck`,
    `e2fsck`, `xfs_repair`, `resize2fs`, `tune2fs`, `xfs_admin`, `fatlabel`, `zstd`,
    `veritysetup`, `grub2-install` (or `grub-install`).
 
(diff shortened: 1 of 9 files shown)


reubeno:
Since you're already proposing switching to a different tool, did you evaluate the pros/cons of directly calling into `libfdisk` or similar? I'd expect it's what `sfdisk` is using underneath anyway, and the prospect of being able to use an API vs. a command-line interface for programmatic disk/partition management is compelling. (Alternatively I see there are a handful of golang packages available for disk and partition manipulation, but it's not clear how mature/complete they are.)

cwize1:
> Since you're already proposing switching to a different tool, did you evaluate the pros/cons of directly calling into `libfdisk` or similar? I'd expect it's what `sfdisk` is using underneath anyway, and the prospect of being able to use an API vs. a command-line interface for programmatic disk/partition management is compelling. (Alternatively I see there are a handful of golang packages available for disk and partition manipulation, but it's not clear how mature/complete they are.)

Calling dynamic libraries in Go is a bit of a pain.

When you use CGO, you also start dynamically linking in the C stdlib. This can make it a lot harder to build an exe once and run it everywhere since you now have to care about the version of stdlib on the host OS. And to make matters worse, there isn't an easy way to cross-compile against an old version of the stdlib. So, you have to use an old OS as your build machine to ensure maximum compatibility.

I figure it is easier to just use `sfdisk` since we are already using a bunch of host tools anyway.

reubeno:
> This can make it a lot harder to build an exe once and run it everywhere since you now have to care about the version of stdlib on the host OS.

True... but `glibc` is great at backwards compatibility. The Rust ecosystem counts on this, as Rust binaries by default dynamically link against libc.

If the push is also for Prism to primarily be run from within a container with dependencies already present, doesn't this concern also become moot?

> So, you have to use an old OS as your build machine to ensure maximum compatibility.

Well, as your build *container*; the underlying machine or VM can be running anything it pleases.

PR #96, opened 2025-01-28
Fix handling of VHDs.

cwize1:
The VHD file format is a little bit of mess, since Hyper-V changed how the file is interpreted from how Microsoft Virtual PC interpreted it. `qemu-img` tries its best to distinguish between the two seemingly identical formats. But sometimes it gets it wrong. So, this change forces `qemu-img` to always use the Hyper-V variant, since Virtual PC is pretty dead at this point.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/imagefileinfo.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/imagefileinfo.go
@@ -1,0 +1,31 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package imagecustomizerlib
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/shell"
+)
+
+type ImageFileInfo struct {
+	Format      string `json:"format"`
+	VirtualSize int64  `json:"virtual-size"`
+}
+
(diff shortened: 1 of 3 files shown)


toolkit/tools/pkg/imagecustomizerlib/imagefileinfo.go
```diff
@@ -0,0 +1,31 @@
...
+}
+
+func getImageFileInfo(inputImageFile string) (ImageFileInfo, error) {
+	stdout, _, err := shell.Execute("qemu-img", "info", "--output", "json", inputImageFile)
+	if err != nil {
+		return ImageFileInfo{}, fmt.Errorf("failed to check image file's disk format:\n%w", err)
+	}
+
+	info := ImageFileInfo{}
+	err = json.Unmarshal([]byte(stdout), &info)
```
liulanze, line 25: `err = json.Unmarshal(stdout, &info)`? Is `stdout` already a string?

liulanze (approved):
Approve with a nit Unmarshal comment.

toolkit/tools/pkg/imagecustomizerlib/imagefileinfo.go
cwize1: `stdout` is a `string`. But `json.Unmarshal` wants a byte array.

PR #98, opened 2025-01-30
Doc: Create HTTP app in VM in Azure.

cwize1:
Add a how-to doc that: downloads a marketplace image, uses Image Customizer to add a HTTP app to the OS image, uploads the new OS image to Azure, and then boots it in a VM.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/docs/imagecustomizer/how-to/azure-vm.md
+++ b/docs/imagecustomizer/how-to/azure-vm.md
@@ -1,0 +1,229 @@
+---
+title: Create Azure VM
+parent: How To
+nav_order: 1
+---
+
+# Create a customized image and deploy it as an Azure VM
+
+This guide shows you how to customize a marketplace Azure Linux image to include a
+simple HTTP application written in Python and then deploy it to Azure as a VM.
+
+## Words of caution
+
+This is intended as an example only.
+
+In general, for web applications, it is better to use a dedicated hosting service like
+[Azure App Service](https://learn.microsoft.com/en-us/azure/app-service/overview),
+[Azure Container Apps](https://learn.microsoft.com/en-us/azure/container-apps/), or
+[Azure Kubernetes Service](https://learn.microsoft.com/en-us/azure/aks/what-is-aks)
+than managing the VMs directly.
+
+In addition, it is a good idea to stick both a load balancer (e.g.
+[Azure Application Gateway](https://learn.microsoft.com/en-us/azure/application-gateway/overview))
+and a CDN (Content Delivery Network) (e.g.
+[Azure Front Door](https://learn.microsoft.com/en-us/azure/frontdoor/front-door-overview))
+in front of any HTTP endpoints.
+
+## Steps
+
+1. Create a directory to stage all the build artifacts:
+
+   ```bash
+   STAGE_DIR="<staging-directory>"
+   mkdir -p "$STAGE_DIR"
+   ```
+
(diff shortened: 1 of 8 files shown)


docs/imagecustomizer/how-to/azure-vm.md
```diff
@@ -0,0 +1,229 @@
...
+     - content: |
+         set -eux
+         python3 -m venv /home/myapp/venv
+         /home/myapp/venv/bin/pip3 install Flask waitress
+   ```
+
+5. Run Image Customizer to create the new image:
+
+   ```bash
+   IMG_CUSTOMIZER_TAG="mcr.microsoft.com/azurelinux/imagecustomizer:0.10.0"
```
gmileka, line 154: we don't 'latest'?

cwize1: We currently don't publish a `latest`.

docs/imagecustomizer/how-to/azure-vm.md
```diff
@@ -0,0 +1,229 @@
...
+         -A INPUT -p icmp --icmp-type 11 -j ACCEPT
+         # Allot ICMP Destination Unreachable - Used for MTU negotiation
+         -A INPUT -p icmp --icmp-type 3 -j ACCEPT
+         # Open port 80.
+         -A INPUT -p tcp -m tcp --dport 80 -j ACCEPT
+         -A OUTPUT -j ACCEPT
+         COMMIT
+
+     packages:
+       install:
```
gmileka, line 128: Is there enough space on the downloaded image hold these new packages/dependencies? I sometimes run out of space if I don't include a storage section with larger partitions.

docs/imagecustomizer/how-to/azure-vm.md
```diff
@@ -0,0 +1,229 @@
...
+       - python3
+       - python3-pip
+
+     services:
+       enable:
+       - nginx
+       - myapp
+
+     users:
+     - name: myapp
```
gmileka, line 139: should we include a log-in mechanism?

toolkit/tools/pkg/imagecustomizerlib/testdata/http-app.yaml
```diff
@@ -0,0 +1,99 @@
+os:
```
gmileka, line 1: do we need this content in both this file and the .md page?

gmileka (approved):

docs/imagecustomizer/how-to/azure-vm.md
cwize1: The OS partition has a size of 4.5 GiB, only 586 MiB is used. So, plenty of margin.

toolkit/tools/pkg/imagecustomizerlib/testdata/http-app.yaml
cwize1: Having a copy in a test YAML is handy for manual testing.

docs/imagecustomizer/how-to/azure-vm.md
cwize1: Probably not. HTTP servers usually have zero state. So, if something goes wrong, just delete it and create a new one. So, there generally isn't any reason to try to login to one,

PR #102, opened 2025-02-04
Enable Image Customizer VM tests.

gmileka:
This change updates the VMTests framework to support:
- ADO pipelines.
- iso images.
- Azl2.0 hosts.

---

### **Checklist**
- [x] Tests added/updated

### DIFF
--- a/test/vmtests/Makefile
+++ b/test/vmtests/Makefile
@@ -16,7 +16,7 @@ TOOLKIT_DIR=../../toolkit
 TOOLS_BIN_DIR=${TOOLKIT_DIR}/out/tools
 IMAGE_CUSTOMIZER_BIN=${TOOLS_BIN_DIR}/imagecustomizer
 
-IMAGE_CUSTOMIZER_CONTAINER_TAG=imagecustomizer:dev
+IMAGE_CUSTOMIZER_CONTAINER_TAG?=imagecustomizer:dev
 
 SSH_PRIVATE_KEY_FILE ?= ${HOME}/.ssh/id_ed25519
 
@@ -62,11 +62,25 @@ ${IMAGE_CUSTOMIZER_BIN}:
 image-customizer-container: ${IMAGE_CUSTOMIZER_BIN}
 	${TOOLKIT_DIR}/tools/imagecustomizer/container/build-mic-container.sh -t ${IMAGE_CUSTOMIZER_CONTAINER_TAG}
 
+.PHONY: test
+test:
+	${PYTEST} \
+		--image-customizer-container-url="${IMAGE_CUSTOMIZER_CONTAINER_TAG}" \
+		--input-image="${INPUT_IMAGE}" \
+		--output-format="${OUTPUT_FORMAT}" \
+		--ssh-private-key="${SSH_PRIVATE_KEY_FILE}" \
+		$(if $(filter y,$(KEEP_ENVIRONMENT)),--keep-environment) \
+		--log-cli-level=DEBUG \
+		--show-capture=all \
+		--tb=short \
+		./vmtests
+
 .PHONY: run
 run: image-customizer-container
 	${PYTEST} \
 		--image-customizer-container-url="${IMAGE_CUSTOMIZER_CONTAINER_TAG}" \
-		--core-efi-azl2="${CORE_EFI_AZL2}" \
+		--input-image="${INPUT_IMAGE}" \
+		--output-format="${OUTPUT_FORMAT}" \
 		--ssh-private-key="${SSH_PRIVATE_KEY_FILE}" \
 		$(if $(filter y,$(KEEP_ENVIRONMENT)),--keep-environment) \
 		--log-cli-level=DEBUG \
--- a/test/vmtests/vmtests/test_no_change.py
+++ b/test/vmtests/vmtests/test_no_change.py
@@ -3,7 +3,10 @@
 
 import os
 from getpass import getuser
+import logging
 from pathlib import Path
+import shlex
+import time
 from typing import List, Tuple
 
 import libvirt  # type: ignore
@@ -21,7 +24,8 @@ from .utils.ssh_client import SshClient
 def test_no_change(
     docker_client: DockerClient,
     image_customizer_container_url: str,
-    core_efi_azl2: Path,
+    input_image: Path,
+    output_format: str,
     ssh_key: Tuple[str, Path],
     test_temp_dir: Path,
     test_instance_name: str,
@@ -30,36 +34,50 @@ def test_no_change(
 ) -> None:
     (ssh_public_key, ssh_private_key_path) = ssh_key
 
-    config_path = TEST_CONFIGS_DIR.joinpath("nochange-config.yaml")
-    output_image_path = test_temp_dir.joinpath("image.qcow2")
-    diff_image_path = test_temp_dir.joinpath("image-diff.qcow2")
+    if output_format != "iso":
+        config_path = TEST_CONFIGS_DIR.joinpath("nochange-config.yaml")
+    else:
+        config_path = TEST_CONFIGS_DIR.joinpath("nochange-iso-config.yaml")
+
+    output_image_path = test_temp_dir.joinpath("image." + output_format)
+
+    boot_type = "efi"
+    if Path(input_image).suffix.lower() == ".vhd" and output_format != "iso":
+        boot_type = "legacy"
 
     username = getuser()
 
     run_image_customizer(
         docker_client,
         image_customizer_container_url,
-        core_efi_azl2,
+        input_image,
         config_path,
         username,
         ssh_public_key,
-        "qcow2",
+        output_format,
         output_image_path,
         close_list,
     )
 
-    # Create a differencing disk for the VM.
-    # This will make it easier to manually debug what is in the image itself and what was set during first boot.
-    local_client.run(
-        ["qemu-img", "create", "-F", "qcow2", "-f", "qcow2", "-b", str(output_image_path), str(diff_image_path)],
-    ).check_exit_code()
+    vm_image = output_image_path
+    if output_format != "iso":
+        diff_image_path = test_temp_dir.joinpath("image-diff.qcow2")
 
-    # Ensure VM can write to the disk file.
-    os.chmod(diff_image_path, 0o666)
+        # Create a differencing disk for the VM.
+        # This will make it easier to manually debug what is in the image itself and what was set during first boot.
+        local_client.run(
+            ["qemu-img", "create", "-F", "qcow2", "-f", "qcow2", "-b", str(output_image_path), str(diff_image_path)],
+        ).check_exit_code()
+
+        # Ensure VM can write to the disk file.
+        os.chmod(diff_image_path, 0o666)
+
+        vm_image = diff_image_path
 
     # Create VM.
     vm_name = test_instance_name
-    domain_xml = create_libvirt_domain_xml(VmSpec(vm_name, 4096, 4, diff_image_path))
+
+    domain_xml = create_libvirt_domain_xml(VmSpec(vm_name, 4096, 4, vm_image), get_host_os(), boot_type)
 
     vm = LibvirtVm(vm_name, domain_xml, libvirt_conn)
     close_list.append(vm)
@@ -69,6 +87,8 @@ def test_no_change(
 
     # Wait for VM to boot by waiting for it to request an IP address from the DHCP server.
     vm_ip_address = vm.get_vm_ip_address(timeout=30)
+    # iso booting takes longer due to the copying of artifacts to memory
+    time.sleep(30)
 
     # Connect to VM using SSH.
     ssh_known_hosts_path = test_temp_dir.joinpath("known_hosts")
@@ -83,5 +103,17 @@ def test_no_change(
         with open(os_release_path, "r") as os_release_fd:
             os_release_text = os_release_fd.read()
 
-            assert "ID=mariner" in os_release_text
-            assert 'VERSION_ID="2.0"' in os_release_text
+            assert ("ID=azurelinux" in os_release_text) or ("ID=mariner" in os_release_text)
+            assert ('VERSION_ID="3.0"' in os_release_text) or ('VERSION_ID="2.0"' in os_release_text)
+
+
+def get_host_os () -> str:
+    with open("/etc/os-release", "r") as f:
+        for line in f:
+            key, _, value = line.partition("=")
+            if key == "NAME":
+                os_name = shlex.split(value)[0]  # Safely handle quoted values
+                print(f"OS Name: {os_name}")
+                break
+
+    return os_name
--- a/test/vmtests/vmtests/utils/libvirt_utils.py
+++ b/test/vmtests/vmtests/utils/libvirt_utils.py
@@ -2,6 +2,7 @@
 # Licensed under the MIT License.
 
 import xml.etree.ElementTree as ET  # noqa: N817
+import os
 from pathlib import Path
 from typing import Dict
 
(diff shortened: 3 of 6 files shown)


test/vmtests/vmtests/test_no_change.py
```diff
@@ -69,6 +87,8 @@ def test_no_change(
 
     # Wait for VM to boot by waiting for it to request an IP address from the DHCP server.
     vm_ip_address = vm.get_vm_ip_address(timeout=30)
+    # iso booting takes longer due to the copying of artifacts to memory
+    time.sleep(30)
```
cwize1, line 91: Do not add arbitrary sleep calls. Add a wait loop of some kind. In this case, you can wait until the SSH port is open.

If you spin on the `SshClient` call, then make sure you are checking for a specific error. Alternatively, you could use Python's [socket](https://docs.python.org/3/library/socket.html) API to check if the TCP port is open yet.

test/vmtests/vmtests/test_no_change.py
```diff
@@ -83,5 +103,17 @@ def test_no_change(
         with open(os_release_path, "r") as os_release_fd:
             os_release_text = os_release_fd.read()
 
-            assert "ID=mariner" in os_release_text
-            assert 'VERSION_ID="2.0"' in os_release_text
+            assert ("ID=azurelinux" in os_release_text) or ("ID=mariner" in os_release_text)
```
cwize1, line 106: The test should know which image it is testing against and therefore should know exactly which string is correct.

test/vmtests/vmtests/utils/libvirt_utils.py
```diff
@@ -136,21 +155,57 @@ def _add_disk_xml(
...
-            # unlikely that someone will ever need more than 26 disks. So, keep it simple
-            # for now.
-            if disk_index < 0 or disk_index > 25:
-                raise Exception(f"Unsupported disk index: {disk_index}.")
-            suffix = chr(ord("a") + disk_index)
-            return f"{prefix}{suffix}"
-
-        case _:
-            return f"{prefix}{disk_index}"
+    # cannot use python's 'match' because Azure Linux 2.0 has an older version
```
cwize1, line 199: It might be good to explicitly specify the minimum Python version we are targeting.

test/vmtests/vmtests/utils/libvirt_utils.py
```diff
@@ -136,21 +155,57 @@ def _add_disk_xml(
...
+    disk_target.attrib["bus"] = bus_type
+
+    disk_source = ET.SubElement(disk, "source")
+    disk_source.attrib["file"] = file_path
+    disk_source.attrib["index"] = "1"
+
+    disk_readonly = ET.SubElement(disk, "readonly")
+
+    disk_alias = ET.SubElement(disk, "alias")
+    disk_alias.attrib["name"] = "ide0-0-0"
```
cwize1, line 186: Make sure you shut down a libvirt VM before copying its XML.

These aren't needed:

- `source['index']`
- `alias`

test/vmtests/vmtests/utils/libvirt_utils.py
```diff
@@ -97,14 +106,24 @@ def create_libvirt_domain_xml(vm_spec: VmSpec) -> str:
...
+            "disk",
+            "qcow2",
+            "virtio",
+            next_disk_indexes,
+        )
+    else:
+        _add_iso_xml(
+            devices,
+            str(vm_spec.os_disk_path),
+            "ide",
```
cwize1, line 124: Why not `sata` instead of `ide`?

test/vmtests/vmtests/utils/libvirt_utils.py
```diff
@@ -136,21 +155,57 @@ def _add_disk_xml(
...
+    disk_source = ET.SubElement(disk, "source")
+    disk_source.attrib["file"] = file_path
+    disk_source.attrib["index"] = "1"
+
+    disk_readonly = ET.SubElement(disk, "readonly")
+
+    disk_alias = ET.SubElement(disk, "alias")
+    disk_alias.attrib["name"] = "ide0-0-0"
+
+    disk_address = ET.SubElement(disk, "address")
```
cwize1, line 188: I think `address` needs to be added for all devices with bus type of sata/ide. So, it could make sense to add this logic to `_add_disk_xml` instead of having a separate `_add_iso_xml` function.

toolkit/tools/pkg/imagecustomizerlib/testdata/nochange-config.yaml
```diff
@@ -1 +1,5 @@
 os:
```
cwize1, line 1: `nochange-config.yaml` is intended to be completely empty. If you need something different, then add a new config file.

cwize1 (changes requested):

test/vmtests/Makefile
```diff
@@ -62,11 +62,25 @@ ${IMAGE_CUSTOMIZER_BIN}:
 image-customizer-container: ${IMAGE_CUSTOMIZER_BIN}
 	${TOOLKIT_DIR}/tools/imagecustomizer/container/build-mic-container.sh -t ${IMAGE_CUSTOMIZER_CONTAINER_TAG}
 
+.PHONY: test
+test:
```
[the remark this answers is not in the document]
gmileka: run builds from source and runs the test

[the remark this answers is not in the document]
cwize1: Ah, fair. Though I wonder if we could change the behavior based on whether or not the user provides a `IMAGE_CUSTOMIZER_CONTAINER_TAG` value.

test/vmtests/Makefile
```diff
@@ -62,11 +62,25 @@ ${IMAGE_CUSTOMIZER_BIN}:
...
+	${PYTEST} \
+		--image-customizer-container-url="${IMAGE_CUSTOMIZER_CONTAINER_TAG}" \
+		--input-image="${INPUT_IMAGE}" \
+		--output-format="${OUTPUT_FORMAT}" \
```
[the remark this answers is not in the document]
gmileka: It's a parameter passed to the Makefile, like in `make test OUTPUT_FORMAT=qcow2`.

test/vmtests/vmtests/utils/libvirt_utils.py
```diff
@@ -30,14 +31,25 @@ def create_libvirt_domain_xml(vm_spec: VmSpec) -> str:
...
+        os_tag.attrib["firmware"] = "efi"
 
     os_type = ET.SubElement(os_tag, "type")
     os_type.text = "hvm"
 
-    firmware = ET.SubElement(domain, "firmware")
-    firmware.attrib["secure-boot"] = "yes"
-    firmware.attrib["enrolled-keys"] = "yes"
+    if boot_type == "efi" and host_os == "Ubuntu":
+        firmware = ET.SubElement(domain, "firmware")
```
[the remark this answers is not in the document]
gmileka: They confuse libvirt on azl2, the image does not boot. I am not sure why `firmware` is not working on azl2. I can open a bug to revisit these `if` statements when azl3 is available. For now, we wanted azl2 to be part of our metrics.

test/vmtests/vmtests/utils/libvirt_utils.py
```diff
@@ -30,14 +31,25 @@ def create_libvirt_domain_xml(vm_spec: VmSpec) -> str:
...
+        firmware.attrib["enrolled-keys"] = "yes"
+
+    nvram = ET.SubElement(os_tag, "nvram")
+
+    if boot_type == "efi":
+        loader = ET.SubElement(os_tag, "loader")
+        loader.attrib["readonly"] = "yes"
+        loader.attrib["secure"] = "no"
+        loader.attrib["type"] = "pflash"
+        loader.text = "/usr/share/OVMF/OVMF_CODE.fd"
```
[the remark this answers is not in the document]
gmileka: This path works for our current test matrix though, why change it/make it more complicated?

test/vmtests/vmtests/utils/libvirt_utils.py
gmileka: You mean in a comment? or a runtime check?

test/vmtests/vmtests/utils/libvirt_utils.py
gmileka: pending

gmileka: Which part are you referring to re. 'copying its XML'?

cwize1: I am guessing that to see how to provision a CDROM device, you created a VM using virt-manager or virt-install and then looked at the VM's xml. However, if the VM is running, the XML will contain runtime metadata/status which isn't needed to create the VM.

test/vmtests/vmtests/utils/libvirt_utils.py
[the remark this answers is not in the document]
cwize1: The `firmware` feature was added in libvirt v7.2.0. And Mariner 2.0 has libvirt v7.10. So, in theory this should work. I wonder why it doesn't. :-(

test/vmtests/vmtests/utils/libvirt_utils.py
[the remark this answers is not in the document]
cwize1: I use Fedora as my dev box. Also, doing it properly makes the code robust the distro version updates.

test/vmtests/vmtests/utils/libvirt_utils.py
cwize1: As a comment and/or in the README would be a good start.

test/vmtests/vmtests/utils/libvirt_utils.py
```diff
@@ -30,14 +125,25 @@ def create_libvirt_domain_xml(vm_spec: VmSpec) -> str:
...
+        os_tag.attrib["firmware"] = "efi"
 
     os_type = ET.SubElement(os_tag, "type")
     os_type.text = "hvm"
 
-    firmware = ET.SubElement(domain, "firmware")
-    firmware.attrib["secure-boot"] = "yes"
-    firmware.attrib["enrolled-keys"] = "yes"
+    if boot_type == "efi" and host_os == "Ubuntu":
+        firmware = ET.SubElement(domain, "firmware")
```
cwize1, line 135: The `firmware` tag is a way to auto select the values for the `loader` tag. If you are manually specifying `loader`, then you don't need `firmware`.

test/vmtests/vmtests/utils/libvirt_utils.py
```diff
@@ -14,8 +18,99 @@ def __init__(self, name: str, memory_mib: int, core_count: int, os_disk_path: Pa
...
+
+
+def get_libvirt_firmware_config(
+        libvirt_conn: libvirt.virConnect,
+        machine_type: str,
+        enable_secure_boot: bool,
+    ) -> Dict[str, Any]:
+        # Resolve the machine type to its full name.
+        domain_caps_str = libvirt_conn.getDomainCapabilities(
+            machine=machine_type, virttype="kvm"
```
cwize1, line 36: Just hardcode `machine="q35"`.

`pc-i440fx` is the old firmware that doesn't support things like secure-boot.

test/vmtests/vmtests/utils/libvirt_utils.py
```diff
@@ -30,14 +125,25 @@ def create_libvirt_domain_xml(vm_spec: VmSpec) -> str:
...
+        firmware = ET.SubElement(domain, "firmware")
+        firmware.attrib["secure-boot"] = "yes"
+        firmware.attrib["enrolled-keys"] = "yes"
+
+    nvram = ET.SubElement(os_tag, "nvram")
+
+    if boot_type == "efi":
+        loader = ET.SubElement(os_tag, "loader")
+        loader.attrib["readonly"] = "yes"
+        loader.attrib["secure"] = "no"
```
cwize1, line 144: You should have the value of `"secure"` match the value of `enable_secure_boot` passed to `get_libvirt_firmware_config`.

test/vmtests/vmtests/utils/libvirt_utils.py
```diff
@@ -14,8 +18,99 @@ def __init__(self, name: str, memory_mib: int, core_count: int, os_disk_path: Pa
...
+            raise LisaException(
+                f"Could not find matching firmware for machine-type={machine_type} "
+                f"({full_machine_type}) and secure-boot={enable_secure_boot}."
+            )
+
+        return firmware_config
+
 # Create XML definition for a VM.
-def create_libvirt_domain_xml(vm_spec: VmSpec) -> str:
+def create_libvirt_domain_xml(vm_spec: VmSpec, host_os: str, boot_type: str, uefi_firmware_binary: str) -> str:
```
cwize1, line 113: I would call `get_libvirt_firmware_config` in `create_libvirt_domain_xml` to avoid adding complexity to the individual tests.

test/vmtests/vmtests/utils/libvirt_utils.py
```diff
@@ -117,9 +232,10 @@ def _add_disk_xml(
     device_type: str,
     image_type: str,
     bus_type: str,
+    device_prefix: str,
```
cwize1, line 235: You can probably derive `device_prefix` from `bus_type`.

vd => virtio device
sd => sata device

test/vmtests/vmtests/utils/libvirt_utils.py
```diff
@@ -136,21 +252,49 @@ def _add_disk_xml(
     disk_source = ET.SubElement(disk, "source")
     disk_source.attrib["file"] = file_path
 
+# Adds a disk to a libvirt domain XML document.
+def _add_iso_xml(
```
cwize1, line 256: If you add a `readonly` param to `_add_disk_xml`, you could implement `_add_iso_xml` using `_add_disk_xml`.

test/vmtests/vmtests/utils/ssh_client.py
```diff
@@ -187,7 +189,19 @@ def __init__(
...
         # Open SSH connection.
-        self.ssh_client.connect(hostname=hostname, port=port, username=username, key_filename=key_filename, sock=sock)
+        start_time = datetime.now()
+        while True:
+            try:
+                self.ssh_client.connect(hostname=hostname, port=port, username=username, key_filename=key_filename, sock=sock)
+                break
+            except (NoValidConnectionsError, SSHException) as e:
+                delta_time = datetime.now() - start_time
+                if delta_time.total_seconds() > 60:
```
cwize1, line 199: Use `time_out_in_seconds` instead of `60`?

test/vmtests/vmtests/utils/libvirt_utils.py
```diff
@@ -30,14 +125,25 @@ def create_libvirt_domain_xml(vm_spec: VmSpec) -> str:
     vcpu.text = str(vm_spec.core_count)
 
     os_tag = ET.SubElement(domain, "os")
-    os_tag.attrib["firmware"] = "efi"
```
cwize1, line 33: Can you double check that this causes problems on AZL2?

test/vmtests/vmtests/test_min_change.py
```diff
@@ -0,0 +1,365 @@
...
+    image_customizer_container_url: str,
+    core_efi_azl2: Path,
+    ssh_key: Tuple[str, Path],
+    test_temp_dir: Path,
+    test_instance_name: str,
+    libvirt_conn: libvirt.virConnect,
+    close_list: List[Closeable],
+) -> None:
+    azl_release = 2
+    config_path = TEST_CONFIGS_DIR.joinpath("os-vm-config.yaml")
```
cwize1, line 148: It would probably be bad if the image can't boot without needing changes. So, please use `"nochange-config.yaml"` here. This should work since `cloud-init` is default enabled and will auto-configure the network.

Also, AZL2 has a default network configuration (outside of what cloud-init provides). This was removed in AZL3.

cwize1 (changes requested):

test/vmtests/vmtests/utils/libvirt_utils.py
cwize1: Out of interest, what problem does setting `fimware="efi"` cause?

cwize1 (approved):

PR #107, opened 2025-02-07
Fix ARM64 unit tests failures

himaja-kesari:
The PR focuses on fixing the unit test failures on arm64

it fixes the following tests by adding support for arm64 in
https://github.com/microsoft/azure-linux-image-tools/blob/41394c19204254c99d08c9ef217d6e0bc5e94546/toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go#L29 
image

---

### **Checklist**
- [ ] Tests added/updated


toolkit/tools/internal/logger/log.go
himaja-kesari, line 1: change to info level before merging

toolkit/tools/pkg/imagecustomizerlib/customizepartitions_test.go
```diff
@@ -189,10 +190,17 @@ func TestCustomizeImagePartitionsSizeOnly(t *testing.T) {
 
 func TestCustomizeImagePartitionsEfiToLegacy(t *testing.T) {
 	for _, version := range supportedAzureLinuxVersions {
-		t.Run(string(version), func(t *testing.T) {
-			testCustomizeImagePartitionsToLegacy(t, "TestCustomizeImagePartitionsEfiToLegacy"+string(version),
-				baseImageTypeCoreEfi, version)
-		})
+		//skip the tests for arm64
+		if runtime.GOARCH == "arm64" {
+			t.Skip("Skipping legacy test for arm64")
```
cwize1, line 195: `t.Skip` stops execution. So, no need for the `else` block.

toolkit/tools/pkg/imagecustomizerlib/customizepartitions_test.go
```diff
@@ -189,10 +190,17 @@ func TestCustomizeImagePartitionsSizeOnly(t *testing.T) {
 
 func TestCustomizeImagePartitionsEfiToLegacy(t *testing.T) {
 	for _, version := range supportedAzureLinuxVersions {
-		t.Run(string(version), func(t *testing.T) {
-			testCustomizeImagePartitionsToLegacy(t, "TestCustomizeImagePartitionsEfiToLegacy"+string(version),
-				baseImageTypeCoreEfi, version)
-		})
+		//skip the tests for arm64
+		if runtime.GOARCH == "arm64" {
```
cwize1, line 194: `runtime.GOARCH` is a constant. So, pull it out of the `for` loop.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -227,18 +239,26 @@ func (b *LiveOSIsoBuilder) stageIsoMakerInitrdArtifacts(writeableRootfsDir, isoM
 		return fmt.Errorf("failed to create %s\n%w", targetBootloadersDir, err)
 	}
 
-	sourceBoot64EfiPath := b.artifacts.bootx64EfiPath
-	targetBoot64EfiPath := filepath.Join(targetBootloadersDir, bootx64Binary)
-	err = file.Copy(sourceBoot64EfiPath, targetBoot64EfiPath)
+	sourceBootEfiPath := b.artifacts.bootx64EfiPath
+	targetBootEfiPath := filepath.Join(targetBootloadersDir, bootx64Binary)
+	if runtime.GOARCH == "arm64" {
```
romoh, line 244: Should we move the logic to a common `getSecureBootEfiPath`? Same for the rest of the code. That way we don't have to spread the logic around the codebase

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -110,7 +120,9 @@ type IsoArtifacts struct {
 	dracutPackageInfo        *PackageVersionInformation
 	selinuxPolicyPackageInfo *PackageVersionInformation
 	bootx64EfiPath           string
+	bootAA64EfiPath          string
```
cwize1, line 123: The code as written will be a bit of a pain to maintain. If changes are ever made to this code, we will need to test both AMD64 and ARM64 to be confident about its correctness. It would be good if the two architectures shared their code instead of duplicating it.

If you want suggestions about how to go about this, feel free to ping me.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -719,18 +753,35 @@ func (b *LiveOSIsoBuilder) extractBootDirFiles(writeableRootfsDir string) error
 		}
 	}
 
-	if b.artifacts.bootx64EfiPath == "" {
-		return fmt.Errorf("failed to find the boot efi file (%s):\n"+
-			"this file is provided by the (shim) package",
-			bootx64Binary)
-	}
+	if runtime.GOARCH == "amd64" {
```
cwize1, line 756: Use switch statement for `GOARCH`. Here and everywhere else.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder_test.go
```diff
@@ -268,12 +269,17 @@ func TestCustomizeImageLiveCdIsoNoShimEfi(t *testing.T) {
 
 	buildDir := filepath.Join(tmpDir, "TestCustomizeImageLiveCdIso")
 	outImageFilePath := filepath.Join(buildDir, "image.iso")
+	shimPackage := "shim"
+	if runtime.GOARCH == "arm64" {
+		shimPackage = "shim-unsigned"
```
cwize1, line 274: Since there is a difference between AZL2 and AZL3 here, we should run this test on both versions. Follow the pattern demonstrated by `TestCustomizeImagePartitionsEfiToLegacy`.

cwize1 (changes requested):

toolkit/tools/internal/logger/log.go
```diff
@@ -58,7 +58,7 @@ const (
 	ColorFlagHelp = "Color setting for log terminal output."
 
 	defaultLogFileLevel   = logrus.DebugLevel
-	defaultStderrLogLevel = logrus.InfoLevel
+	defaultStderrLogLevel = logrus.DebugLevel
```
liulanze, line 61: Why does the log level here change to debug? Wouldn't the info level enough for testing?

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder_test.go
himaja-kesari: created a task for that ..need to debug a bit more https://dev.azure.com/mariner-org/ECF/_workitems/edit/10977/

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
liulanze, line 244: This way requires reading through the logic to see that the initial assignments might be changed later. Using a switch statement for `runtime.GOARCH` could make the architecture dependent assignment explicit.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -580,7 +600,9 @@ func generatePxeGrubCfg(inputContentString string, pxeIsoImageBaseUrl string, px
 //     false otherwise.
 func containsGrubNoPrefix(filePaths []string) bool {
 	for _, filePath := range filePaths {
-		if filepath.Base(filePath) == grubx64NoPrefixBinary {
+		if runtime.GOARCH == "arm64" && filepath.Base(filePath) == grubAA64NoPrefixBinary {
```
liulanze, line 603: The function currently checks `runtime.GOARCH` on every loop iteration, which isn't necessary. I think `runtime.GOARCH` checking can be pulled out of the loop and just check file path in the loop, since the `filePath` is what iterated in the loop.

toolkit/tools/internal/logger/log.go
himaja-kesari: added a comment to change this before merging

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder_test.go
himaja-kesari: updated

toolkit/tools/pkg/imagecustomizerlib/customizepartitions_test.go
cwize1: Pull the `if` check out of the `for` loop.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -253,6 +262,44 @@ func (b *LiveOSIsoBuilder) stageIsoMakerInitrdArtifacts(writeableRootfsDir, isoM
 	return nil
 }
 
+// getGrubBinary
+//
+// returns the name of the grub binary for the current architecture.
+func getGrubBinary() (string, error) {
+
+	archBootBinaries := map[string]string{
```
cwize1, line 270: Move this mapping to the top of the file. (Make it a global.)

Similar comments below.

cwize1: It might be beneficial to use a settings struct.

e.g.

```golang
type BootFilesArchConfig struct {
    grubBinary string,
    grubNoPrefixBinary string,
    bootBinary string,
    …
}

var (
    bootloaderFilesConfig = map[string]BootFilesArchConfig{ … }
)

File: toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go (hunk 4 above)
Comment (himaja-kesari) [reply toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go]: updated the code to having a map of archs

File: toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go [hunk 10]
```diff
@@ -717,16 +853,34 @@ func (b *LiveOSIsoBuilder) extractBootDirFiles(writeableRootfsDir string) error
...
 
-	if b.artifacts.bootx64EfiPath == "" {
-		return fmt.Errorf("failed to find the boot efi file (%s):\n"+
-			"this file is provided by the (shim) package",
-			bootx64Binary)
-	}
+	switch runtime.GOARCH {
+	case "amd64":
+
+		if b.artifacts.bootEfiPath == "" {
```
cwize1, line 859: This change still has a ton of code that is duplicated for each arch. Please remove all the duplication.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
himaja-kesari: updated the code, If we want to add support to new architectures we need to add the supported architecture to the getArchitecture function and add the respective constants and variable

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -227,18 +275,24 @@ func (b *LiveOSIsoBuilder) stageIsoMakerInitrdArtifacts(writeableRootfsDir, isoM
...
+	arch, err := getArchitecture()
+	if err != nil {
+		return err
+	}
+
+	sourceBootEfiPath := b.artifacts.bootEfiPath
+	targetBootEfiPath := filepath.Join(targetBootloadersDir, bootloaderFilesConfig[arch].bootBinary)
+	err = file.Copy(sourceBootEfiPath, targetBootEfiPath)
 	if err != nil {
 		return fmt.Errorf("failed to stage bootloader file (bootx64.efi):\n%w", err)
```
cwize1, line 287: Replace `bootx64.efi` with `bootloaderFilesConfig[arch].bootBinary`?

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -227,18 +275,24 @@ func (b *LiveOSIsoBuilder) stageIsoMakerInitrdArtifacts(writeableRootfsDir, isoM
...
-	sourceGrub64EfiPath := b.artifacts.grubx64EfiPath
-	targetGrub64EfiPath := filepath.Join(targetBootloadersDir, grubx64Binary)
-	err = file.Copy(sourceGrub64EfiPath, targetGrub64EfiPath)
+	sourceGrubEfiPath := b.artifacts.grubEfiPath
+	targetGrubEfiPath := filepath.Join(targetBootloadersDir, bootloaderFilesConfig[arch].grubBinary)
+
+	err = file.Copy(sourceGrubEfiPath, targetGrubEfiPath)
 	if err != nil {
-		return fmt.Errorf("failed to stage bootloader file (grubx64.efi):\n%w", err)
+		return fmt.Errorf("failed to stage bootloader file :\n%w", err)
```
cwize1, line 295: It might be nice to include `bootloaderFilesConfig[arch].grubBinary` in the error message.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -126,6 +161,19 @@ type LiveOSIsoBuilder struct {
 	cleanupDirs []string
 }
 
+// getArchitecture
+//
+//	Returns the architecture of the current machine.
+//	Supported architectures are amd64 and arm64.
+func getArchitecture() (string, error) {
+
```
cwize1, line 169: Don't add blank lines after a `{`.

Same comment above.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -126,6 +161,19 @@ type LiveOSIsoBuilder struct {
 	cleanupDirs []string
 }
 
+// getArchitecture
+//
+//	Returns the architecture of the current machine.
+//	Supported architectures are amd64 and arm64.
+func getArchitecture() (string, error) {
```
cwize1, line 168: A lot of the code that uses `getArchitecture` wants a copy of the `BootFilesArchConfig` for the arch. So, it might be worth adding this as an additional return value.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder_test.go
```diff
@@ -264,16 +265,34 @@ func TestCustomizeImageLiveCd2(t *testing.T) {
...
+	baseImage := checkSkipForCustomizeImage(t, baseImageTypeCoreEfi, version)
+
+	buildDir := filepath.Join(tmpDir, testName)
 	outImageFilePath := filepath.Join(buildDir, "image.iso")
+	shimPackage := "shim"
+
+	// For arm64 and baseImageVersionAzl2, the shim package is shim-unsigned.
+	if runtime.GOARCH == "arm64" && version == baseImageVersionAzl2 {
+		shimPackage = "shim-unsigned"
+
```
cwize1, line 288: Remove empty line.

cwize1 (changes requested):

cwize1 (approved):

cwize1 (approved):

PR #109, opened 2025-02-08
Add multi-kernel support.

cwize1:
Remove the arbitary restrictions to allow the kernel args to be updated when multiple kernels are installed.

In addition, remove the logic that sets `"GRUB_DISABLE_RECOVERY"` when verity is enabled, since the logic can now handle multiple boot entries in the grub.cfg file when a single kernel is installed.

Also, remove the requirement for `$kernelopts` to exist in the grub.cfg file, since this logic just complicates the code.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
@@ -92,10 +92,6 @@ func prepareGrubConfigForVerity(verityList []imagecustomizerapi.Verity, imageChr
 			if err := bootCustomizer.PrepareForVerity(); err != nil {
 				return err
 			}
-		} else if mountPath == "/usr" {
-			if err := bootCustomizer.PrepareForUsrVerity(); err != nil {
-				return err
-			}
 		}
 	}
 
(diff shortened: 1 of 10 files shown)


toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -92,10 +92,6 @@ func prepareGrubConfigForVerity(verityList []imagecustomizerapi.Verity, imageChr
 			if err := bootCustomizer.PrepareForVerity(); err != nil {
 				return err
 			}
-		} else if mountPath == "/usr" {
-			if err := bootCustomizer.PrepareForUsrVerity(); err != nil {
```
gmileka, line 96: Why is this being removed? seems unrelated to this pr?

gmileka: ignore - building context...

toolkit/tools/pkg/imagecustomizerlib/testdata/multikernel-azl2.yaml
```diff
@@ -0,0 +1,10 @@
+os:
+  packages:
+    install:
+    - kernel-5.15.167.1-2.cm2
+    - kernel-5.15.173.1-1.cm2
+
+  kernelCommandLine:
+    extraCommandLine:
+    - console=tty0
```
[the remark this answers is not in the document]
gmileka: It doesn't matter - these are just for the sake of matching in the test...

toolkit/tools/pkg/imagecustomizerlib/testdata/multikernel-azl2.yaml
```diff
@@ -0,0 +1,10 @@
+os:
+  packages:
+    install:
+    - kernel-5.15.167.1-2.cm2
```
gmileka, line 4: curious, does grub regenerates grub.cfg when new kernels are installed? or do we trigger that in prism?

gmileka (approved):

toolkit/tools/pkg/imagecustomizerlib/testdata/multikernel-azl2.yaml
cwize1: Both. The kernel package install will trigger `grub2-mkconfig`. And Prism/MIC will also run it to handle the `extraCommandLine`.

PR #110, opened 2025-02-10
Verity usr functional test.

liulanze:
---

### **Checklist**
- [ ] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/testdata/verity-usr-config.yaml
+++ b/toolkit/tools/pkg/imagecustomizerlib/testdata/verity-usr-config.yaml
@@ -1,0 +1,83 @@
+storage:
+  bootType: efi
+  disks:
+  - partitionTableType: gpt
+    maxSize: 5120M
+    partitions:
+    - id: esp
+      type: esp
+      start: 1M
+      end: 9M
+
+    - id: boot
+      start: 9M
+      end: 1024M
+
+    - id: root
+      start: 1024M
+      end: 3072M
+
+    - id: usrhash
+      start: 3072M
+      end: 3200M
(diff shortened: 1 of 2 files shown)


toolkit/tools/pkg/imagecustomizerlib/testdata/verity-usr-config.yaml
```diff
@@ -0,0 +1,83 @@
...
+    - id: esp
+      type: esp
+      start: 1M
+      end: 9M
+
+    - id: boot
+      start: 9M
+      end: 1024M
+
+    - id: root
```
cwize1, line 16: I am thinking it might make sense to place the root partition after the usr partition, since the usr partition has a fixed size but the root partition can theoretically be expanded.

toolkit/tools/pkg/imagecustomizerlib/testdata/verity-usr-config.yaml
```diff
@@ -0,0 +1,83 @@
+storage:
+  bootType: efi
+  disks:
+  - partitionTableType: gpt
+    maxSize: 5120M
+    partitions:
+    - id: esp
+      type: esp
+      start: 1M
```
cwize1, line 9: I would use `size` for partitions and omit `maxSize`. This should make the config a little simpler.

toolkit/tools/pkg/imagecustomizerlib/testdata/verity-usr-config.yaml
liulanze: Fixed.

cwize1 (approved):

PR #111, opened 2025-02-10
Remove --output-split-partitions-format API

romoh:
Deprecate in favor of COSI.

---

### **Checklist**
- [x] Tests added/updated

### DIFF
--- a/docs/imagecustomizer/api/cli.md
+++ b/docs/imagecustomizer/api/cli.md
@@ -45,12 +45,11 @@ Added in v0.3.
 
 ## --output-image-format=FORMAT
 
+Required
+
 The image format of the the final customized image.
 
 Options: vhd, vhd-fixed, vhdx, qcow2, raw, iso, and [cosi](./cosi.md).
-
-At least one of `--output-image-format` and `--output-split-partitions-format` is
-required.
 
 The vhd-fixed option outputs a fixed size VHD image. This is the required format for
 VMs in Azure.
@@ -58,29 +57,6 @@ VMs in Azure.
 When the output image format is set to iso, the generated image is a LiveOS
 iso image. For more details on this format, see:
 [Image Customizer ISO Support](../concepts/iso.md).
-
-Added in v0.3.
-
-## --output-split-partitions-format=FORMAT
-
-Format of partition files. If specified, disk partitions will be extracted as separate
-files and a json file with partition metadata will be produced. For more details on
-the json file format, see: [Partition Metadata JSON Format](./partitionmetadatajson.md).
-
-Options: raw, raw-zst.
-
-Added in v0.3.
-
-## --shrink-filesystems
-
-Enable shrinking of partition filesystems to their minimum size.
-
-Currently only supports ext2/ext3/ext4 filesystems.
-
-Can only be specified if `--output-split-partitions-format` is, and
-cannot be specified with `--output-image-format`.
-
-Added in v0.3.
 
 ## --config-file=FILE-PATH
 
--- a/toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
@@ -11,7 +11,6 @@ import (
 	"path/filepath"
 	"runtime"
 
-	"github.com/klauspost/compress/zstd"
 	"github.com/microsoft/azurelinux/toolkit/tools/imagegen/diskutils"
 	"github.com/microsoft/azurelinux/toolkit/tools/internal/logger"
 	"github.com/microsoft/azurelinux/toolkit/tools/internal/safeloopback"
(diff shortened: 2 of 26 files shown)


toolkit/tools/pkg/imagecustomizerlib/customizeverity_test.go
```diff
@@ -86,81 +83,6 @@ func testCustomizeImageVerityHelper(t *testing.T, testName string, imageType bas
 		"PARTUUID="+partitions[4].PartUuid, "root")
 }
 
-func TestCustomizeImageVerityShrinkExtract(t *testing.T) {
```
cwize1, line 89: All these tests that you have removed are still valid tests. But they need to be updated to use COSI.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/customizeverity_test.go
```diff
@@ -120,8 +120,8 @@ func testCustomizeImageVerityShrinkExtractHelper(t *testing.T, testName string,
 	config.Storage.FileSystems[hashPartitionNum-1].Type = "ext4"
 
 	// Customize image, shrink partitions, and split the partitions into individual files.
-	err = CustomizeImage(buildDir, testDir, &config, baseImage, nil, outImageFilePath, "", "raw",
-		"" /*outputPXEArtifactsDir*/, true /*useBaseImageRpmRepos*/, true /*enableShrinkFilesystems*/)
+	err = CustomizeImage(buildDir, testDir, &config, baseImage, nil, outImageFilePath, "cosi",
+		"" /*outputPXEArtifactsDir*/, true /*useBaseImageRpmRepos*/)
 	if !assert.NoError(t, err) {
 		return
 	}
```
cwize1, line 127: The `image_XX.raw` file no longer exist when the output is COSI. You need to extract them from the COSI file.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions_test.go
```diff
@@ -141,147 +139,3 @@ func verifySkippableFrameMetadataFromFile(partitionFilepath string, magicNumber
 
 	return nil
 }
-
-// Tests partition extracting with partition resize enabled, but where the partition resize is a no-op.
-func TestCustomizeImageNopShrink(t *testing.T) {
```
cwize1, line 146: This is also a test that should be refactored to use cosi.

cwize1 (changes requested):

docs/imagecustomizer/api/cli.md
```diff
@@ -35,37 +35,19 @@ The file path to write the final customized image to.
...
 
 When the output image format is set to iso, the generated image is a LiveOS
 iso image. For more details on this format, see:
 [Image Customizer ISO Support](../concepts/iso.md).
 
-## --output-split-partitions-format=FORMAT
-
-Format of partition files. If specified, disk partitions will be extracted as separate
-files and a json file with partition metadata will be produced. For more details on
-the json file format, see: [Partition Metadata JSON Format](./partitionmetadatajson.md).
```
cwize1, line 56: You should delete the `partitionmetadatajson.md` file.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions_test.go
romoh: The test doesn't make sense for cosi since we no longer have a reference to unshrunk partitions to compare to.

cwize1: When the image is exported as COSI, the partitions/filesystems are shrunk. And if the config doesn't leave any free space on one of the partitions eligible to be shrunk, then the filesystem shrink operation might not do anything. And in this case, we need to make sure we don't crash.

So, this test is still valid.

himaja-kesari: refactored the tests to use cosi

cwize1:
This change is sitting on top of a commit that is a few weeks old and there are merge conflicts. It would be good to rebase.

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -284,3 +284,86 @@ func populateVerityMetadata(source string, verity *Verity) error {
 
 	return nil
 }
+
+// extractPartitionsFromCosi extracts the partition files from a COSI file
+// and returns a list of their paths.
+// It skips directories, metadata.json, and files that are not .zst.
+// It also decompresses the .zst files and writes them to the output directory.
+func extractPartitionsFromCosi(cosiFilePath, outputDir string) ([]string, error) {
```
cwize1, line 292: This function is only being used in test code. So, you can probably move it into a `*_test.go` file.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions_test.go
```diff
@@ -247,14 +243,20 @@ func TestCustomizeImageExtractEmptyPartition(t *testing.T) {
...
-	espPartitionFilePath := filepath.Join(buildDir, "image_1.raw")
-	rootfsPartitionFilePath := filepath.Join(buildDir, "image_2.raw")
+	// Attach partition files.
+	partitionsPaths, err := extractPartitionsFromCosi(outImageFilePath, buildDir)
+
+	if !assert.NoError(t, err) || !assert.Len(t, partitionsPaths, 3) {
+		return
+	}
+
+	espPartitionFilePath := partitionsPaths[0]
```
cwize1, line 258: I am not sure it is a good idea for the tests to depend on order of files in the tar file. For the `zst` files in `TestCustomizeImageNopShrink`, you reference them by filename. Would it make sense to do the same here?

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/extractpartitions_test.go
himaja-kesari: referenced by filename

toolkit/tools/pkg/imagecustomizerlib/customizeverity_test.go
```diff
@@ -125,8 +125,7 @@ func testCustomizeImageVerityShrinkExtractHelper(t *testing.T, testName string,
 	baseImage := checkSkipForCustomizeImage(t, imageType, imageVersion)
```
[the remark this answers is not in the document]
himaja-kesari: updated the test names

cwize1 (approved):

cwize1 (approved):

PR #133, opened 2025-02-21
Docs: Fix example config in quick start guide.

cwize1:
Example config is using an old API.

---

### **Checklist**

- [x] Tests added/updated


Britel (approved):

romoh (approved):
I wonder if we should run the json schema validation against our samples somehow.

PR #140, opened 2025-02-27
Ensure partition metadata is populated in kernel.

cwize1:
The `udevadm settle` command is commonly used to wait for udev to read a disk and populate the disk's metadata in the kernel. While this works most of the time, it is well known `udevadm settle` isn't quite enough to guarantee that the metadata population has finished. In theory, `udevadm wait` is new command that does provide that guarantee. But that command isn't widely available yet.

This change works around this problem by adding a wait loop that checks the kernel's cached disk's metadata and compares it to the expected values read directly from disk.

This change also replaces the usage of `partprobe` with a call to the `BLKRRPART` (block device read partition) IOCTL syscall. This was done because `partprobe` manually handles the partition node updates. Whereas, calling `BLKRRPART` allows the kernel to handle the update, which is more efficient and reliable.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/imagegen/diskutils/diskutils.go
+++ b/toolkit/tools/imagegen/diskutils/diskutils.go
@@ -445,7 +445,7 @@ func WaitForLoopbackToDetach(devicePath string, diskPath string) error {
 }
 
 func WaitForDiskDevice(diskDevPath string) error {
-	err := WaitForDevicesToSettle()
+	err := waitForDevicesToSettle()
 	if err != nil {
 		return err
 	}
@@ -534,9 +534,9 @@ func waitForDiskToPopulate(diskDevPath string) error {
 	return nil
 }
 
-// WaitForDevicesToSettle waits for all udev events to be processed on the system.
+// waitForDevicesToSettle waits for all udev events to be processed on the system.
 // This can be used to wait for partitions to be discovered after mounting a disk.
-func WaitForDevicesToSettle() error {
+func waitForDevicesToSettle() error {
 	logger.Log.Debugf("Waiting for devices to settle")
 	_, _, err := shell.Execute("udevadm", "settle")
 	if err != nil {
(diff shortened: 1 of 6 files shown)


toolkit/tools/imagegen/diskutils/diskutils.go
```diff
@@ -920,6 +1051,66 @@ func GetDiskPartitions(diskDevPath string) ([]PartitionInfo, error) {
...
+
+	if output.PartitionTable == nil {
+		// Disk is empty.
+		return nil, nil
+	}
+
+	partitionTable := output.PartitionTable
+
+	if partitionTable.Unit != "sectors" {
+		return nil, fmt.Errorf("sfdisk returned unepexted unit size '%s': expecting 'sectors'", partitionTable.Unit)
```
romoh, line 1086: nit - typo `expected`

toolkit/tools/imagegen/diskutils/diskutils.go
```diff
@@ -411,6 +444,96 @@ func WaitForLoopbackToDetach(devicePath string, diskPath string) error {
...
+	if err != nil {
+		return fmt.Errorf("timed out waiting for disk (%s) info to be populated:\n%w", diskDevPath, err)
+	}
+
+	return nil
+}
+
 // WaitForDevicesToSettle waits for all udev events to be processed on the system.
 // This can be used to wait for partitions to be discovered after mounting a disk.
 func WaitForDevicesToSettle() error {
```
romoh, line 539: Do we see other usages for `WaitForDevicesToSettle()` or should we remove it to prevent accidental usage in scenarios where it is not behaving accurately?
I still see a reference in https://github.com/microsoft/azure-linux-image-tools/blob/9d4d908f0cf2a969f1ae2d3857d1523714e06115/toolkit/tools/pkg/imagecustomizerlib/customizepartitionsuuids.go#L68

romoh (changes requested):
Few comments

toolkit/tools/imagegen/diskutils/diskutils.go
cwize1: Fixed.

toolkit/tools/imagegen/diskutils/diskutils.go
cwize1: Good call.

PR #141, opened 2025-02-27
2-phase verity.

cwize1:
Add support for enabling verity in an image in 2 phases:

1. Run MIC/Prism to create the partitions layout.
2. Run MIC/Prism again to enable verity.

This will be useful for scenarios where a base image is created that contains a set of common configurations and then that base image is recustomized into multiple variant images.

This change also includes a functional test for this feature.

---

### **Checklist**

- [x] Tests added/updated


toolkit/tools/imagecustomizerapi/Identifiedpartitiontype.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+func (i IdentifiedPartitionType) IsValid() error {
+	switch i {
+	case IdentifiedPartitionTypePartLabel:
+		// All good.
+		return nil
+
+	default:
+		return fmt.Errorf("invalid value (%v)", i)
```
liulanze, line 23: nit - `(%s)`?

toolkit/tools/imagecustomizerapi/storage.go
```diff
@@ -161,29 +201,49 @@ func (s *Storage) IsValid() error {
...
+		if err != nil {
+			return err
 		}
+	}
 
-		expectedVerityName, validMount := verityMountMap[filesystem.MountPoint.Path]
+	return nil
+}
+
+func ValidateVerityMounts(vertiyDevices []Verity, verityDeviceMountPoint map[*Verity]*MountPoint) error {
```
liulanze, line 225: typo here.

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -308,3 +322,81 @@ func updateUkiKernelArgsForVerity(verityMetadata map[string]verityDeviceMetadata
...
+	if err != nil {
+		return err
+	}
+
+	logger.Log.Debugf("HELLO: %s", fmt.Sprintf("%#v", config.Storage.Verity))
+
+	return nil
+}
+
+func findIdentitiedPartition(partitions []diskutils.PartitionInfo, ref imagecustomizerapi.IdentifiedPartition,
```
liulanze, line 387: typo on the function name.

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -308,3 +322,81 @@ func updateUkiKernelArgsForVerity(verityMetadata map[string]verityDeviceMetadata
...
+		}
+		verityDeviceMountPoint[verity] = mountPoint
+	}
+
+	err = imagecustomizerapi.ValidateVerityMounts(config.Storage.Verity, verityDeviceMountPoint)
+	if err != nil {
+		return err
+	}
+
+	logger.Log.Debugf("HELLO: %s", fmt.Sprintf("%#v", config.Storage.Verity))
```
liulanze, line 382: Please delete the logger here.

liulanze (changes requested):
Based on the PR, the Verity section is needed but the disk section can be avoided. In the future do we want to avoid the Verity section potentially? For example Prism can auto detect the base image's Verity configuration, and its storage layout, then just carry over the verity after the customization.

cwize1:
> Based on the PR, the Verity section is needed but the disk section can be avoided. In the future do we want to avoid the Verity section potentially? For example Prism can auto detect the base image's Verity configuration, and its storage layout, then just carry over the verity after the customization.

Yes, we might add support for that in the future, if there is an ask for it.

toolkit/tools/imagecustomizerapi/Identifiedpartitiontype.go
cwize1: Fixed

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
cwize1: Fixed

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
cwize1: Fixed. Thanks!

toolkit/tools/imagecustomizerapi/storage.go
cwize1: Fixed

liulanze (approved):

PR #147, opened 2025-03-04
Add partition size checks to tests.

cwize1:
In the CustomizeImage tests that customize partitions, verify the final size of the partitions.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/imagegen/diskutils/diskutils.go
+++ b/toolkit/tools/imagegen/diskutils/diskutils.go
@@ -67,7 +67,7 @@ type PartitionInfo struct {
 	Mountpoint        string `json:"mountpoint"` // Example: /mnt/os/boot
 	PartLabel         string `json:"partlabel"`  // Example: boot
 	Type              string `json:"type"`       // Example: part
-	Size              uint64 `json:"size"`       // Example: 4096
+	SizeInBytes       uint64 `json:"size"`       // Example: 4096
 }
 
 type loopbackListOutput struct {
--- a/toolkit/tools/imagegen/diskutils/diskutils.go
+++ b/toolkit/tools/imagegen/diskutils/diskutils.go
@@ -67,6 +67,7 @@ type PartitionInfo struct {
 	Mountpoint        string `json:"mountpoint"` // Example: /mnt/os/boot
 	PartLabel         string `json:"partlabel"`  // Example: boot
 	Type              string `json:"type"`       // Example: part
+	Size              uint64 `json:"size"`       // Example: 4096
 }
 
 type loopbackListOutput struct {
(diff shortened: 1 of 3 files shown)


toolkit/tools/imagegen/diskutils/diskutils.go
```diff
@@ -67,6 +67,7 @@ type PartitionInfo struct {
 	Mountpoint        string `json:"mountpoint"` // Example: /mnt/os/boot
 	PartLabel         string `json:"partlabel"`  // Example: boot
 	Type              string `json:"type"`       // Example: part
+	Size              uint64 `json:"size"`       // Example: 4096
```
gmileka, line 70: nit: `SizeInBytes`?

toolkit/tools/pkg/imagecustomizerlib/customizeverity_test.go
```diff
@@ -144,11 +148,38 @@ func testCustomizeImageVerityShrinkExtractHelper(t *testing.T, testName string,
...
+	rootStat, err := os.Stat(rootPartitionPath)
+	assert.NoError(t, err)
+
+	hashStat, err := os.Stat(hashPartitionPath)
+	assert.NoError(t, err)
 
+	varStat, err := os.Stat(varPartitionPath)
+	assert.NoError(t, err)
+
+	// Check partition sizes.
```
gmileka, line 172: is the goal here to ensure that those partitions that should remain unchanged remain unchanged?

toolkit/tools/pkg/imagecustomizerlib/customizeverity_test.go
```diff
@@ -144,11 +148,38 @@ func testCustomizeImageVerityShrinkExtractHelper(t *testing.T, testName string,
...
+	assert.Equal(t, int64(8*diskutils.MiB), espStat.Size())
+	assert.Equal(t, int64(100*diskutils.MiB), hashStat.Size())
+
+	// These partitions are shrunk.
+	// So, their size will vary slightly.
+	assert.Greater(t, int64(100*diskutils.MiB), bootStat.Size())
```
gmileka, line 178: where do the 100, 650, 150 numbers come from?

toolkit/tools/pkg/imagecustomizerlib/customizeverity_test.go
cwize1: The goal is to ensure that the partition specification provided by the user is provisioned exactly.

toolkit/tools/imagegen/diskutils/diskutils.go
cwize1: Fixed.

toolkit/tools/pkg/imagecustomizerlib/customizeverity_test.go
cwize1: They are somewhat arbitrary values that make the test work. I've added a comment.

romoh (approved):

toolkit/tools/pkg/imagecustomizerlib/customizepartitions_test.go
```diff
@@ -119,6 +119,12 @@ func testCustomizeImagePartitionsToEfi(t *testing.T, testName string, imageType
 		assert.Equal(t, "4f68bce3-e8cd-4db1-96e7-fbcaf984b709", partitions[3].PartitionTypeUuid) // root (x64)
 		assert.Equal(t, "4d21b016-b534-45c2-a9fb-5c16e091fd2d", partitions[4].PartitionTypeUuid) // var
 	}
+
+	// Check the partition sizes.
+	assert.Equal(t, uint64(8*diskutils.MiB), partitions[1].SizeInBytes)
+	assert.Equal(t, uint64(99*diskutils.MiB), partitions[2].SizeInBytes)
+	assert.Equal(t, uint64(1940*diskutils.MiB), partitions[3].SizeInBytes)
```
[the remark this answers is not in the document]
Britel: Understood - and not a blocking comment - but I would suggest considering if there is a better way to have one source of truth. Either a framework where the test config is generated from a data set that is also used for verification, or one where we parse the yaml when checking what kernel did.

PR #151, opened 2025-03-05
Change shrink filesystems to use `sfdisk` instead of `parted`.

cwize1:
The API for `sfdisk` is a little nicer and `sfdisk` supported various features earlier than `parted`.

Also, when changing a partition size, `sfdisk` only needs the `size`. Whereas, `parted` needs both the start and end of the partition. So, using `sfdisk` simplifies the code a lot.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/shrinkfilesystems.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/shrinkfilesystems.go
@@ -6,38 +6,19 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/microsoft/azurelinux/toolkit/tools/imagecustomizerapi"
 	"github.com/microsoft/azurelinux/toolkit/tools/imagegen/diskutils"
 	"github.com/microsoft/azurelinux/toolkit/tools/internal/logger"
 	"github.com/microsoft/azurelinux/toolkit/tools/internal/shell"
+	"github.com/sirupsen/logrus"
 )
 
-var (
-	// Parsing output of: fdisk --list <device>
-	//
-	// Example:
-	//   Device     Start     End Sectors Size Type
-	//   /dev/vda1   2048   18431   16384   8M EFI System
-	//   /dev/vda2  18432 8386559 8368128   4G Linux filesystem
-	fdiskPartitionsTableHeaderRegexp = regexp.MustCompile(`(?m)^Device[\t ]+Start[\t ]+`)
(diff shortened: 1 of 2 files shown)


toolkit/tools/pkg/imagecustomizerlib/shrinkfilesystems.go
```diff
@@ -69,34 +45,39 @@ func shrinkFilesystems(imageLoopDevice string, verity []imagecustomizerapi.Verit
...
 		}
 
-		if end == "" {
+		if filesystemSizeInSectors < 0 {
 			// Filesystem wasn't resized. So, there is no need to resize the partition.
 			logger.Log.Infof("Filesystem is already at its min size (%s)", partitionLoopDevice)
 			continue
 		}
 
 		// Resize the partition with parted resizepart
```
romoh, line 70: nit - update comment

romoh (changes requested):

toolkit/tools/pkg/imagecustomizerlib/shrinkfilesystems.go
cwize1: Fixed.

romoh (approved):

toolkit/tools/pkg/imagecustomizerlib/shrinkfilesystems.go
```diff
@@ -69,34 +45,39 @@ func shrinkFilesystems(imageLoopDevice string, verity []imagecustomizerapi.Verit
...
 		}
 
-		// Resize the partition with parted resizepart
-		_, stderr, err = shell.ExecuteWithStdin("yes" /*stdin*/, "flock", "--timeout", "5", imageLoopDevice,
-			"parted", "---pretend-input-tty", imageLoopDevice, "resizepart",
-			strconv.Itoa(partitionNumber), end)
+		// Resize the partition.
+		sfdiskScript := fmt.Sprintf("unit: sectors\nsize=%d", filesystemSizeInSectors)
+
+		err = shell.NewExecBuilder("flock", "--timeout", "5", imageLoopDevice, "sfdisk", "--lock=no",
```
gmileka, line 73: are we setting `lock` to `no` because we rely on `flock`?

cwize1: Yeah, `sfdisk` will by default take the lock but it doesn't have a timeout mechanism. So, I use `flock` instead.

gmileka (approved):

PR #153, opened 2025-03-05
Rename OsArch from amd64 to x86_64

elainezhao1:
Change OsArch from amd64 to x86_64 for consistent naming based on trident's cosi def
doc already reflected this change.

---

### **Checklist**


toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -287,3 +287,10 @@ func populateVerityMetadata(source string, verity *Verity) error {
 
 	return nil
 }
+
+func getArchitecture() string {
```
cwize1, line 291: nit: Rename to `getArchitectureForCosi` since all the functions in the same directory are in the same namespace.

cwize1 (changes requested):

cwize1 (approved):

PR #159, opened 2025-03-06
Add output.image.format

vinceaperri:
Add output.image.format as the default value for --output-image-format. Also anticipate the removal of --output-split-partitions-forma and --shrink-partitions.

Also fix missing vhd-fixed in help for --output-image-format.

### DIFF
--- a/toolkit/tools/imagecustomizerapi/imageFormatType.go
+++ b/toolkit/tools/imagecustomizerapi/imageFormatType.go
@@ -33,9 +33,9 @@ var supportedImageFormatTypes = []string{
 	string(ImageFormatTypeCosi),
 }
 
-func (ft *ImageFormatType) IsValid() error {
-	if *ft != ImageFormatTypeNone && !slices.Contains(SupportedImageFormatTypes(), string(*ft)) {
-		return fmt.Errorf("invalid image format type (%s)", *ft)
+func (ft ImageFormatType) IsValid() error {
+	if ft != ImageFormatTypeNone && !slices.Contains(SupportedImageFormatTypes(), string(ft)) {
+		return fmt.Errorf("invalid image format type (%s)", ft)
 	}
 
 	return nil


docs/imagecustomizer/api/cli.md
```diff
@@ -39,26 +39,22 @@ provided in the configuration file. If both `output.image.path` and
...
+is used.
 
-The vhd-fixed option outputs a fixed size VHD image. This is the required format for
-VMs in Azure.
+The image format of the the final customized image.
 
-When the output image format is set to iso, the generated image is a LiveOS
-iso image. For more details on this format, see:
-[Image Customizer ISO Support](../concepts/iso.md).
+See [output.image.format](./configuration/outputImage.md#format-string)
```
cwize1, line 56: Forcing a user to jump between pages to find information is a little annoying for the user. So, I would duplicate the list of supported formats in both docs.

toolkit/tools/imagecustomizerapi/output_image.go
```diff
@@ -4,7 +4,8 @@
 package imagecustomizerapi
```
cwize1, line 4: Sorry, I missed this before, but this file should be named `outputimage.go`.

toolkit/tools/imagecustomizerapi/output_image.go
```diff
@@ -4,7 +4,8 @@
...
 
 type OutputImage struct {
-	Path string `yaml:"path" json:"path,omitempty"`
+	Path   string `yaml:"path" json:"path,omitempty"`
+	Format string `yaml:"format" json:"format,omitempty"`
```
cwize1, line 8: You should verify that format contains a valid value.

See, `FileSystemType` for an example.

cwize1 (changes requested):

toolkit/tools/imagecustomizerapi/imageFormat.go
```diff
@@ -0,0 +1,50 @@
...
+		return nil
+	}
+
+	return fmt.Errorf("failed to parse image format: %s. Supported formats: %s",
+		val, strings.Join(SupportedImageFormats(), ", "))
+}
+
+// SupportedImageFormats returns all valid image formats.
+func SupportedImageFormats() []string {
+	return []string{
```
cwize1, line 41: nit: Move into a `var` at the top of the file.

toolkit/tools/imagecustomizer/main.go
```diff
@@ -19,8 +21,8 @@ var (
 
 	buildDir                 = app.Flag("build-dir", "Directory to run build out of.").Required().String()
 	imageFile                = app.Flag("image-file", "Path of the base Azure Linux image which the customization will be applied to.").Required().String()
-	outputImageFile          = app.Flag("output-image-file", "Path to write the customized image to.").Required().String()
-	outputImageFormat        = app.Flag("output-image-format", "Format of output image. Supported: vhd, vhdx, qcow2, raw, iso, cosi.").Enum("vhd", "vhd-fixed", "vhdx", "qcow2", "raw", "iso", "cosi")
+	outputImageFile          = app.Flag("output-image-file", "Path to write the customized image to.").String()
+	outputImageFormat        = app.Flag("output-image-format", fmt.Sprintf("Format of output image. Supported: %s.", imagecustomizerapi.SupportedImageFormats())).Enum(imagecustomizerapi.SupportedImageFormats()...)
```
cwize1, line 25: Do you want to use `strings.Join(` here as well?

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -162,13 +158,6 @@ func createImageCustomizerParameters(buildDir string,
 	ic.outputImageDir = filepath.Dir(ic.outputImageFile)
 	ic.outputPXEArtifactsDir = outputPXEArtifactsDir
 
-	if ic.outputImageFormat != "" && !ic.outputIsIso {
```
cwize1, line 165: Why was this validation removed? I think this exists so that we fail early in the run instead of at the very end.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -475,21 +464,21 @@ func convertWriteableFormatToOutputImage(ic *ImageCustomizerParameters, inputIso
 
 	// Create final output image file if requested.
 	switch ic.outputImageFormat {
-	case ImageFormatVhd, ImageFormatVhdFixed, ImageFormatVhdx, ImageFormatQCow2, ImageFormatRaw:
+	case imagecustomizerapi.ImageFormatVhd, imagecustomizerapi.ImageFormatVhdFixed, imagecustomizerapi.ImageFormatVhdx, imagecustomizerapi.ImageFormatQcow2, imagecustomizerapi.ImageFormatRaw:
```
cwize1, line 467: This line is too long. Please line wrap at 120.

cwize1 (changes requested):

toolkit/tools/imagecustomizerapi/imageFormat.go
```diff
@@ -0,0 +1,53 @@
...
+	string(ImageFormatVhd),
+	string(ImageFormatVhdFixed),
+	string(ImageFormatVhdx),
+	string(ImageFormatQcow2),
+	string(ImageFormatRaw),
+	string(ImageFormatIso),
+	string(ImageFormatCosi),
+}
+
+func (f *ImageFormat) UnmarshalYAML(unmarshal func(interface{}) error) error {
```
cwize1, line 35: Please follow the existing patterns for string enum types, like `FileSystemType`. Keeping code consistent makes it easier to maintain.

Also, having the `IsValid` function be separate from the unmarshaling code was an intentional design choice. The imager tool handled verification and unmarshalling together and I found it made the code pretty annoying.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
cwize1: The `CustomizerImage` function is called directly by the tests. So, it can be handy to include/duplicate all the verification checks to make writing tests easier.

cwize1 (changes requested):

toolkit/tools/imagecustomizerapi/imageFormat.go
cwize1: The enum types are mostly an internal implementation detail and don't show up in the API. Hence, there is less scrutiny paid to their naming.

toolkit/tools/imagecustomizerapi/imageFormatType.go
```diff
@@ -0,0 +1,47 @@
...
+	string(ImageFormatTypeVhd),
+	string(ImageFormatVhdTypeFixed),
+	string(ImageFormatTypeVhdx),
+	string(ImageFormatTypeQcow2),
+	string(ImageFormatTypeRaw),
+	string(ImageFormatTypeIso),
+	string(ImageFormatTypeCosi),
+}
+
+func (ft *ImageFormatType) IsValid() error {
```
cwize1, line 36: `string` (`ImageFormatType`) is already a reference type. So, you don't need to capture it by pointer.

cwize1 (dismissed):

PR #176, opened 2025-03-13
Switch to Kong CLI parser.

cwize1:
Switch from the kingpin CLI parser to using kong instead. Kong is the successor of kingpin and is more actively maintained. This means that some of the warts have been fixed and there are more features available. In particular, kong has an inbuilt option to print usage on error.

This change also introduces a `build` sub-command. If no command is specified, `build` is the default command. This will allow us to easily add other sub-commands in the future. For example, `validate-config`.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/imagecustomizer/main.go
+++ b/toolkit/tools/imagecustomizer/main.go
@@ -17,7 +17,7 @@ import (
 	"github.com/microsoft/azurelinux/toolkit/tools/pkg/imagecustomizerlib"
 )
 
-type BuildCmd struct {
+type CustomizeCmd struct {
 	BuildDir                 string   `name:"build-dir" help:"Directory to run build out of." required:""`
 	InputImageFile           string   `name:"image-file" help:"Path of the base Azure Linux image which the customization will be applied to."`
 	OutputImageFile          string   `name:"output-image-file" help:"Path to write the customized image to."`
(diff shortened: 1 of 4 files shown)


toolkit/tools/imagecustomizer/main.go
```diff
@@ -4,57 +4,77 @@
...
-	configFile               = app.Flag("config-file", "Path of the image customization config file.").Required().String()
-	rpmSources               = app.Flag("rpm-source", "Path to a RPM repo config file or a directory containing RPMs.").Strings()
-	disableBaseImageRpmRepos = app.Flag("disable-base-image-rpm-repos", "Disable the base image's RPM repos as an RPM source").Bool()
-	outputPXEArtifactsDir    = app.Flag("output-pxe-artifacts-dir", "Create a directory with customized image PXE booting artifacts. '--output-image-format' must be set to 'iso'.").String()
-	logFlags                 = exe.SetupLogFlags(app)
-	timestampFile            = app.Flag("timestamp-file", "File that stores timestamps for this program.").String()
-)
+type CustomizeCmd struct {
+	BuildDir                 string   `name:"build-dir" help:"Directory to run build out of." required:""`
+	InputImageFile           string   `name:"image-file" help:"Path of the base Azure Linux image which the customization will be applied to."`
```
frhuelsz, line 22: tip: for all these files you are expecting should exist you can add `type:existingFile` and kong will check that for you.
(https://github.com/alecthomas/kong?tab=readme-ov-file#custom-named-decoders)

frhuelsz (approved):
lgtm

toolkit/tools/imagecustomizer/main.go
cwize1: The `existingFile` parser has some additional behavior that I don't want. For example, on Linux, `~` should be interpreted by the shell not the program.

PR #177, opened 2025-03-13
Prism Output Input API Design.

liulanze:
This PR will not be merged, since the logic code is not started, but it can be reviewed for the API design purpose.


docs/imagecustomizer/api/configuration/input.md
```diff
@@ -4,18 +4,36 @@ parent: Configuration
...
 
 ```yaml
 input:
   image:
     path: ./base.vhdx
+  verityHashes:
+    path: /input/verityhashes
+  ukis:
+    path: /input/ukis
+  bootloader:
```
gmileka, line 19: is this the bootloader itself or the bootloader configuration? I'd expect it's the configuration, correct?

docs/imagecustomizer/api/configuration/input.md
```diff
@@ -4,18 +4,36 @@ parent: Configuration
...
+Specifies the configuration for the input image and associated files required for customizing.
 
 Example:
 
 ```yaml
 input:
   image:
     path: ./base.vhdx
+  verityHashes:
+    path: /input/verityhashes
```
gmileka, line 16: how do the contents of this path map to individual partitions?

docs/imagecustomizer/api/configuration/input.md
```diff
@@ -4,18 +4,36 @@ parent: Configuration
...
 
 ## image [[inputImage](./inputImage.md)]
 
 Specifies the configuration for the input image.
 
 Added in v0.13.0.
+
+## verityHashes [[inputVerityHashes](./inputVerityHashes.md)]
+
+Specifies the configuration for inputting signed Verity hash files into the image.
```
gmileka, line 31: nit: `embedding signed verity hash files into the image` ?

docs/imagecustomizer/api/configuration/input.md
```diff
@@ -4,18 +4,36 @@ parent: Configuration
...
+
```
gmileka, line 32: we probably should document the `path` field for each of these?

gmileka: oh - there are separate files... never mind..

docs/imagecustomizer/api/configuration/inputBootloader.md
```diff
@@ -0,0 +1,37 @@
...
+The specified directory must contain the following pre-signed bootloader components:
+
+- bootx64.efi – The signed shim executable.
+- systemd-bootx64.efi – The signed systemd-boot executable.
+
+### Signed Bootloader Components
+
+The files in path should be pre-signed bootloader executables.
+
+- Prism does not perform signing. It is recommended to use ESRP for signing bootloader components.
```
gmileka, line 33: We can't use the term `ESRP` because it is an internal term/service.

docs/imagecustomizer/api/configuration/inputBootloader.md
```diff
@@ -0,0 +1,37 @@
...
+
+# inputBootloader type
+
+Specifies the configuration for inputting signed bootloader components, including shim and systemd-boot, into the image.
+
+Example:
+
+```yaml
+bootloader:
+  path: /input/bootloader
```
gmileka, line 13: Should we keep this as a folder with 'expected contents' or break down the field into individual field for each of the individual components?

docs/imagecustomizer/api/configuration/inputUkis.md
```diff
@@ -0,0 +1,33 @@
...
+
+Required.
+
+Specifies the directory path that contains all signed UKI PE images.
+
+### Expected Files in the Directory
+
+The specified directory must contain signed UKI PE images. Typical files found in this directory include:
+
+- vmlinuz-<version>.signed.efi – A signed UKI PE image.
```
gmileka, line 26: is this specific naming format required?

docs/imagecustomizer/api/configuration/inputVerityHashes.md
```diff
@@ -0,0 +1,34 @@
...
+
+Required.
+
+Specifies the directory path that contains all signed Verity hash files.
+
+### Expected Files in the Directory
+
+The specified directory must contain signed Verity hash files. Typical files found in this directory include:
+
+- root.hash.sig – The signed Verity hash for the root filesystem.
```
gmileka, line 26: same comment here as that for bootloader... should we break those down?

docs/imagecustomizer/api/configuration/input.md
```diff
@@ -4,18 +4,36 @@ parent: Configuration
...
 Example:
 
 ```yaml
 input:
   image:
     path: ./base.vhdx
+  verityHashes:
+    path: /input/verityhashes
+  ukis:
+    path: /input/ukis
```
jiria, line 18: This is applicable to the other paths as well, if these are meant as directory paths, how do you decide to what include? E.g., how do you tell what file represents a verity hash, which one is a UKI and which one is something to ignore?

docs/imagecustomizer/api/configuration/input.md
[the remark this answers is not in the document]
romoh: I wonder if this should actually call out what verity hash is this for? `root` or `usr` ?

docs/imagecustomizer/api/configuration/outputBootloader.md
```diff
@@ -0,0 +1,36 @@
...
+
+Required.
+
+Specifies the directory path where Prism will output the generated bootloader components.
+
+### Expected Files in the Directory
+
+After the image customization process, this directory will contain the following unsigned bootloader components:
+
+- bootx64.efi – The generated shim executable.
```
romoh, line 26: For `arm` this will be different, right? so maybe this should be arch agnostic?

romoh (commented):
Since uki is a preview feature, will those be preview feature as well?

cwize1:
> Since uki is a preview feature, will those be preview feature as well?

All new APIs need to sit under a preview flag for 2+ months.

docs/imagecustomizer/api/configuration/output.md
```diff
@@ -13,10 +13,28 @@ output:
   image:
     path: ./out/image.vhdx
     format: vhdx
+  verityHashes:
+    path: /output/verityhashes
+  ukis:
+    path: /output/ukis
+  bootloader:
+    path: /output/bootloader
```
fintelia, line 21: Are users likely to want to individually control these? An alternative would be to pass a single directory and have Prism create the three subdirectories under it

docs/imagecustomizer/api/configuration/input.md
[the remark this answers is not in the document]
liulanze: Make sense. And this field is designed to take a directory, and to automatically search the root hash files, either for `root.hash.sig` or `usr.hash.sig`, or both, thoughts?

liulanze: The path is for the host, where Prism should look for the signed files. Then the input path is kind of default, like we always want to input UKI to esp partition.

docs/imagecustomizer/api/configuration/outputBootloader.md
liulanze: Will revise, thanks.

PR #178, opened 2025-03-13
Add build docker validation step

romoh:
PR workflow improvement


toolkit/tools/imagecustomizer/container/build-container.sh
```diff
@@ -52,22 +52,22 @@ function cleanUp() {
...
 
-micLocalFile="$enlistmentRoot/toolkit/out/tools/imagecustomizer"
-micLicensesDir="$enlistmentRoot/toolkit/out/tools/LICENSES/imagecustomizer"
+LocalFile="$enlistmentRoot/toolkit/out/tools/imagecustomizer"
+LicensesDir="$enlistmentRoot/toolkit/out/tools/LICENSES/imagecustomizer"
 
 stagingBinDir="${containerStagingFolder}/usr/local/bin"
 stagingLicensesDir="${containerStagingFolder}/usr/local/share/licenses"
 
-dockerFile="$scriptDir/Dockerfile.mic-container"
```
cwize1, line 61: There should still be some prefix added to the `Dockerfile`. This is a way of signaling that the `Dockerfile` is not usable directly.

toolkit/tools/imagecustomizer/container/build-container.sh
```diff
@@ -52,22 +52,22 @@ function cleanUp() {
 }
 trap 'cleanUp' ERR
 
-micLocalFile="$enlistmentRoot/toolkit/out/tools/imagecustomizer"
-micLicensesDir="$enlistmentRoot/toolkit/out/tools/LICENSES/imagecustomizer"
+LocalFile="$enlistmentRoot/toolkit/out/tools/imagecustomizer"
```
cwize1, line 55: nit: Keep naming scheme consistent with the rest of the code.

cwize1, line 55: `LocalFile` is a little too general. Perhaps something like `exeFile`?

.github/workflows/build-and-test.yml
```diff
@@ -87,5 +87,9 @@ jobs:
           echo "Schema is up-to-date!"
         fi
 
+    - name: Build docker image
+      run: |
+        docker build toolkit/tools/imagecustomizer/container/
```
cwize1, line 92: You should use the `build-container.sh` script here.

cwize1 (changes requested):

cwize1 (approved):

PR #190, opened 2025-04-01
Fix quick start tdnf commands for Azure Linux environments

vinceaperri:


docs/imagecustomizer/how-to/quick-start.md
```diff
@@ -35,26 +35,52 @@ nav_order: 1
         xfsprogs zstd cryptsetup-bin grub2-common binutils lsof
      ```
 
-   - For Azure Linux 2.0, run:
+   - For Azure Linux 2.0 on x86_64, run:
```
cwize1, line 38: I think this might be getting a little spammy. Would it make sense to provide a base set of instructions for Azure Linux, and then provide additional instructions for both x86_64 and AZL3?

docs/imagecustomizer/how-to/quick-start.md
```diff
@@ -35,26 +35,52 @@ nav_order: 1
...
+        sed createrepo_c squashfs-tools cdrkit e2fsprogs dosfstools \
+        xfsprogs zstd veritysetup grub2 binutils lsof
+     ```
+
+      Note: This command is identical to the command for x86_64 machines,
+      except that `grub2-pc` is not installed, as above with Azure Linux
+      2.0 on arm64. `systemd-ukify` is also not installed, as explained
+      below.
+
+   Note: The `ukify` tool is only available in Azure Linux 3.0 on x86_64 machines. So, you
```
cwize1, line 80: `ukify` isn't available on ARM64? We should probably fix that.

romoh (dismissed):

cwize1 (approved):

cwize1 (approved):

PR #193, opened 2025-04-02
Support older lsof versions.

cwize1:
Older versions of lsof don't support the `-Q` argument. So, detect when an older version of `lsof` is being used and apply a workaround.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/internal/processes/processes.go
+++ b/toolkit/tools/internal/processes/processes.go
@@ -7,12 +7,19 @@ import (
 	"bufio"
 	"fmt"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 
 	"github.com/microsoft/azurelinux/toolkit/tools/internal/logger"
 	"github.com/microsoft/azurelinux/toolkit/tools/internal/shell"
 	"github.com/sirupsen/logrus"
+)
+
+var (
+	// Example:
+	//     revision: 4.93.2
+	lsofVersionRegexp = regexp.MustCompile(`(?m)^\s*revision:\s+(\d+)\.(\d+)\.\d+\s*$`)
 )
 
 type ProcessRecord struct {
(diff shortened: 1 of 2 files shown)


toolkit/tools/internal/processes/processes.go
```diff
@@ -23,10 +30,30 @@ type ProcessRecord struct {
...
+	args = append(args, "-F", "pcn", "--", path)
+
+	stdout, _, err := shell.NewExecBuilder("lsof", args...).
 		LogLevel(logrus.TraceLevel, logrus.DebugLevel).
 		ExecuteCaptureOuput()
 	if err != nil {
+		if !qArgAvailable {
+			// The -Q arg isn't available. So, this error could just mean there are no results.
+			// So, just hope that this is in fact the case, to avoid unexpected errors.
+			return nil, nil
```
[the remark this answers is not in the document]
cwize1: Unfortunately, it isn't unusual for there to be spurious warnings reported by `lsof`. So, I don't think it will be easy to parse the stderr. Also, in theory, there shouldn't be any other error except for no-results. So, hopefully this should be fine.

toolkit/tools/internal/processes/processes.go
```diff
@@ -79,6 +106,28 @@ func GetProcessesUsingPath(path string) ([]ProcessRecord, error) {
...
+
+	match := lsofVersionRegexp.FindStringSubmatch(stderr)
+	if match == nil {
+		return 0, 0, fmt.Errorf("failed to parse lsof version string")
+	}
+
+	majorStr := match[1]
+	minorStr := match[2]
+
+	major, _ := strconv.Atoi(majorStr)
```
liulanze, line 125: Do we wanted have an error handling? Like:

```
major, err := strconv.Atoi(majorStr)
if err != nil {
...
}
```

toolkit/tools/internal/processes/processes.go
```diff
@@ -79,6 +106,28 @@ func GetProcessesUsingPath(path string) ([]ProcessRecord, error) {
 	return records, nil
 }
 
+func getLsofVersion() (int, int, error) {
+	_, stderr, err := shell.NewExecBuilder("lsof", "-v").
+		LogLevel(logrus.TraceLevel, logrus.TraceLevel).
+		ExecuteCaptureOuput()
```
liulanze, line 112: I guess this function initially had a wrong name typo - `ExecuteCaptureOu**t**put()`, but could you help to fix it?

liulanze (approved):

toolkit/tools/internal/processes/processes.go
cwize1: I've opened a bug.

PR #199, opened 2025-04-14
Use 64 bit integers when shrinking partitions.

cwize1:
Currently, the shrink partitions logic uses 32 bit integers. So, it can't handle disk sizes above 4 GiB. This change fixes this.

Also, refactor the shrinking logic somewhat. This is being done in preparation of shrinking verity hash partitions.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/shrinkfilesystems.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/shrinkfilesystems.go
@@ -21,6 +21,11 @@ func shrinkFilesystems(imageLoopDevice string) error {
 		return err
 	}
 
+	sectorSize, _, err := diskutils.GetSectorSize(imageLoopDevice)
+	if err != nil {
+		return fmt.Errorf("failed to get disk (%s) sector size:\n%w", imageLoopDevice, err)
+	}
+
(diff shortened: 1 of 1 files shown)


toolkit/tools/pkg/imagecustomizerlib/shrinkfilesystems.go
```diff
@@ -113,41 +144,37 @@ func getFilesystemSizeInSectors(resize2fsStdout string, resize2fsStderr string,
 			resize2fsStderr)
 	}
 
-	blockCount, err := strconv.Atoi(match[1]) // Example: 21015
+	blockCount, err := strconv.ParseUint(match[1], 10, 64) // Example: 21015
```
romoh, line 147: Is the switch from Atoi to allow handling of larger filesystem sizes?

romoh (approved):

toolkit/tools/pkg/imagecustomizerlib/shrinkfilesystems.go
cwize1: Yeah. There isn't a direct equivalent for `Atoi` that is 64-bit. You have to use `ParseUint` instead.

PR #200, opened 2025-04-15
Refactor ISO generation logic.

cwize1:
Currently, the Prism ISO generation uses the `IsoMaker` class. However, the `IsoMaker` class has a lot of logic that isn't relevant to Prism and instead is only relevant to generating an installer ISO.

This change splits up the ISO generation logic to pull out the common ISO logic into a separate class that Prism can use instead of `IsoMaker`.

This has the additional benefit of removing some unneeded artifacts from Prism's ISOs.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/internal/isogenerator/isogenerator.go
+++ b/toolkit/tools/internal/isogenerator/isogenerator.go
@@ -32,19 +32,19 @@ type IsoGenConfig struct {
 	// Directory where to stage ISO files.
 	// If the directory exists, any existing files will be included in the ISO.
 	StagingDirPath string
+	// The path of the initrd file to use in the ISO.
+	InitrdPath string
 	// Enable legacy boot.
 	// Note: This isn't useful unless some additional assets are included in 'StagingDirPath'.
 	EnableBiosBoot bool
-	// The path where the ISO file will be written.
-	OutputFilePath string
 	// The directory in the ISO where the following files will be written to:
 	// - initrd.img
 	// - vmlinuz
 	// - isolinux.bin (for BIOS boot)
 	// - boot.cat (for BIOS boot)
 	IsoOsFilesDirPath string
-	// The path of the initrd file to use in the ISO.
-	InitrdPath string
+	// The path where the ISO file will be written.
+	OutputFilePath string
 }
 
 type isoGenInfo struct {
--- a/toolkit/tools/internal/isogenerator/isogenerator.go
+++ b/toolkit/tools/internal/isogenerator/isogenerator.go
@@ -1,0 +1,338 @@
+// Copyright Microsoft Corporation.
+// Licensed under the MIT License.
+
+package isogenerator
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+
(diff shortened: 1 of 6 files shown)


toolkit/tools/internal/isogenerator/isogenerator.go
```diff
@@ -0,0 +1,338 @@
...
+// copyInitrd copies a pre-built initrd into the isolinux folder.
+func copyInitrd(info isoGenInfo) error {
+	initrdDestinationPath := filepath.Join(info.Config.StagingDirPath, info.Config.IsoOsFilesDirPath, "initrd.img")
+
+	logger.Log.Debugf("Copying initrd from '%s'.", info.Config.InitrdPath)
+
+	return file.Copy(info.Config.InitrdPath, initrdDestinationPath)
+}
+
+func setUpIsoGrub2Bootloader(info isoGenInfo) (err error) {
```
gmileka, line 145: nit: maybe 'createIsoBootloaderImage()'?

toolkit/tools/internal/isogenerator/isogenerator.go
```diff
@@ -0,0 +1,338 @@
...
+	// Directory where temporary files can be stored.
+	BuildDirPath string
+	// Directory where to stage ISO files.
+	// If the directory exists, any existing files will be included in the ISO.
+	StagingDirPath string
+	// Enable legacy boot.
+	// Note: This isn't useful unless some additional assets are included in 'StagingDirPath'.
+	EnableBiosBoot bool
+	// The path where the ISO file will be written.
+	OutputFilePath string
```
gmileka, line 39: Chris, this is the output parameter, so maybe we keep it at the end?
- { build }
- { inputs }
- { outputs }

toolkit/tools/internal/isogenerator/isogenerator.go
```diff
@@ -0,0 +1,338 @@
+// Copyright Microsoft Corporation.
+// Licensed under the MIT License.
+
+package isogenerator
```
gmileka, line 4: why is this under internal and not under pkg?

gmileka (dismissed):

toolkit/tools/internal/isogenerator/isogenerator.go
cwize1: `pkg` is for official APIs, particularly those that have an equivalent CLI too. `internal` is for internal implementation details. `internal` made more sense to me for this class.

toolkit/tools/internal/isogenerator/isogenerator.go
cwize1: `setUpIsoGrub2Bootloader` is the name used by `IsoMaker`. I'd rather keep the name the same to make the history clearer to follow.

toolkit/tools/internal/isogenerator/isogenerator.go
cwize1: Fixed.

gmileka (dismissed):

romoh (dismissed):
Can we add some additional tests for `isogenerator`?

romoh (approved):

PR #201, opened 2025-04-15
Shrink verity hash partition.

cwize1:
When partitions are shrunk (e.g. when outputting COSI files), also shrink the verity hash partitons. This ensures that COSI files don't contain any unnecessary data.

This change might also be useful in the future, if and when we ever decide to implement an API that automatically calulates the size of verity hash partitions, instead of requiring the user to manually specify the size.

---

### **Checklist**

- [x] Tests added/updated


liulanze (approved):
LGTM!

PR #103, opened 2025-02-04
Support usr Verity.

liulanze:
---

### **Checklist**
- [x] Tests added/updated
- [x] Documentation updated (if needed)
- [x] Code conforms to style guidelines

The code has been verified locally with these test cases:

- core -> root + grub
- core -> root + uki
- core -> usr + grub
- core -> usr + uki
- core -> root + usr + grub
- core -> root + usr + uki
- usr + grub -> usr + grub
- usr + grub -> usr + uki

Example output:

```
root [ /home/test ]# dmsetup info
Name:              root
State:             ACTIVE (READ-ONLY)
Read Ahead:        256
Tables present:    LIVE
Open count:        1
Event number:      0
Major, minor:      254, 0
Number of targets: 1
UUID: CRYPT-VERITY-bb1ebee9b36648059288e7ca77c55d32-root

Name:              usr
State:             ACTIVE (READ-ONLY)
Read Ahead:        256
Tables present:    LIVE
Open count:        1
Event number:      0
Major, minor:      254, 1
Number of targets: 1
UUID: CRYPT-VERITY-801076ddc8c7437686f02f8d7620e4f1-usr
```

PR #206, opened 2025-04-18
Fix verity corruption option.

cwize1:
PR #103 accidentally removed the logic that handles 'corruptionOption'. So, fix that. Also, change the existing verity tests to also verify 'corruptionOption'.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/testdata/verity-usr-config.yaml
+++ b/toolkit/tools/pkg/imagecustomizerlib/testdata/verity-usr-config.yaml
@@ -24,7 +24,7 @@ storage:
     name: usr
     dataDeviceId: usr
     hashDeviceId: usrhash
-    corruptionOption: panic
+    corruptionOption: io-error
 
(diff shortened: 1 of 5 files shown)


romoh (approved):

toolkit/tools/pkg/imagecustomizerlib/testdata/verity-usr-config.yaml
```diff
@@ -24,7 +24,7 @@ storage:
     name: usr
     dataDeviceId: usr
     hashDeviceId: usrhash
-    corruptionOption: panic
+    corruptionOption: io-error
```
liulanze, line 27: Does `io-error` has specific meaning for usr verity?

cwize1: No, just changing one of the verity configs so that the different tests can check for different values.

PR #209, opened 2025-04-22
Verity hash output support.

liulanze:
---

### **Checklist**
- [ ] Tests added/updated


liulanze:
WIP limiting `verity-hashSignatureInjection` to be under ESP partition. But please have an initial review @cwize1

toolkit/tools/pkg/imagecustomizerlib/artifactsinputoutput.go
```diff
@@ -140,6 +142,26 @@ func outputArtifacts(items []imagecustomizerapi.OutputArtifactsItemType,
...
+			err = file.Write(verity.rootHash, destPath)
+			if err != nil {
+				return fmt.Errorf("failed to dump root hash for (%s) to (%s):\n%w", verity.name, destPath, err)
+			}
+
+			signedHashFile := replaceSuffix(unsignedHashFile, ".hash", ".hash.sig")
+			outputArtifactsMetadata = append(outputArtifactsMetadata, imagecustomizerapi.InjectArtifactMetadata{
+				Partition:      partition,
+				Source:         "./" + signedHashFile,
+				Destination:    filepath.Join("/", SystemdBootDir, signedHashFile),
```
liulanze, line 159: Need to correct

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -806,6 +813,57 @@ func validateOutput(baseConfigPath string, output imagecustomizerapi.Output, out
...
+
+	err := validateVeritySignatureInjection(storage.Verity, storage.FileSystems)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func validateVeritySignatureInjection(verityList []imagecustomizerapi.Verity,
```
cwize1, line 829: This should be handled in `Storage.IsValid`.

Actually, more specifically it should be handled `ValidateVerityMounts`. `ValidateVerityMounts` is called by both `Storage.IsValid` and `validateVerityMountPaths`, `validateVerityMountPaths` is used when verity is initialized on pre-created partitions.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -806,6 +813,57 @@ func validateOutput(baseConfigPath string, output imagecustomizerapi.Output, out
...
+	fileSystems []imagecustomizerapi.FileSystem,
+) error {
+	if len(verityList) == 0 {
+		return nil
+	}
+
+	var espMount string
+
+	for _, fs := range fileSystems {
```
cwize1, line 838: This isn't how you would find the ESP. Instead, you should use partition type UUID. `Storage.IsValid` function already has this logic.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -806,6 +813,57 @@ func validateOutput(baseConfigPath string, output imagecustomizerapi.Output, out
...
+	}
+
+	espMountClean := filepath.Clean(espMount)
+
+	for _, verity := range verityList {
+		if verity.HashSignatureInjection == "" {
+			continue
+		}
+
+		sigPath := filepath.Clean(verity.HashSignatureInjection)
```
cwize1, line 856: I would add an API validity check that ensures that the path provided by the user is "clean".

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -806,6 +813,57 @@ func validateOutput(baseConfigPath string, output imagecustomizerapi.Output, out
...
+
+		// Ensure it's under the ESP mount point
+		if !strings.HasPrefix(sigPath, espMountClean+"/") {
```
cwize1, line 859: It's worth noting that `Storage.IsValid` has a check that ensures the ESP is mounted at `/boot/efi`.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -970,12 +1028,14 @@ func customizeVerityImageHelper(buildDir string, config *imagecustomizerapi.Conf
 		}
 
 		verityMetadata[verityConfig.MountPath] = verityDeviceMetadata{
-			rootHash:              rootHash,
-			dataPartUuid:          dataPartition.PartUuid,
-			hashPartUuid:          hashPartition.PartUuid,
-			dataDeviceMountIdType: verityConfig.DataDeviceMountIdType,
-			hashDeviceMountIdType: verityConfig.HashDeviceMountIdType,
-			corruptionOption:      verityConfig.CorruptionOption,
+			name:                   imagecustomizerapi.VerityMountMap[verityConfig.MountPath],
```
cwize1, line 1031: `name` is available in `verityConfig.Name`.

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -38,6 +40,37 @@ func enableVerityPartition(verity []imagecustomizerapi.Verity, imageChroot *safe
...
+			diskPartitions, err := diskutils.GetDiskPartitions(imageConnection.loopback.DevicePath())
+			if err != nil {
+				return false, err
+			}
+			systemBootPartition, err := findSystemBootPartition(diskPartitions)
+			if err != nil {
+				return false, err
+			}
+
+			fstabDracutConfigFile := filepath.Join(imageChroot.RootDir(), dracutConfigDir, "fstab.conf")
```
cwize1, line 54: The `createNewImageHelper` function has logic that overrides the `/etc/fstab` file with the new partitions. We should delete the `fstab.conf` file (if it exists) at the same time.

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -200,11 +234,24 @@ func constructVerityKernelCmdlineArgs(verityMetadata map[string]verityDeviceMeta
...
 			fmt.Sprintf("%s=%s", dataArg, formattedDataPartition),
 			fmt.Sprintf("%s=%s", hashKey, formattedHashPartition),
-			fmt.Sprintf("%s=%s", optionsArg, formattedCorruptionOption),
+			fmt.Sprintf("%s=%s", optionsArg, options),
+		)
+	}
+
+	if hasSignatureInjection {
+		newArgs = append(newArgs,
+			"dm_verity.require_signatures=1",
```
cwize1, line 253: What if the user specifies two verity devices, one with a signature file and one without? Should we just let the user add this manually if they want it?

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -38,6 +40,37 @@ func enableVerityPartition(verity []imagecustomizerapi.Verity, imageChroot *safe
 		return false, fmt.Errorf("failed to add dracut modules for verity:\n%w", err)
 	}
 
+	for _, v := range verity {
```
cwize1, line 43: Use `slices.ContainsFunc(c.Storage.Verity, func (verity Verity) { return verity.HashSignatureInjection != "" })` instead of a for-loop.

toolkit/resources/dracutmodules/90mountesppartition/mountesppartition-generator.sh
```diff
@@ -0,0 +1,77 @@
...
+function updateVeritySetupUnit () {
+    systemdDropInDir=/etc/systemd/system
+    verityDropInDir=$systemdDropInDir/systemd-veritysetup@root.service.d
+
+    mkdir -p $verityDropInDir
+    verityConfiguration=$verityDropInDir/verity-azl-extension.conf
+
+    cat <<EOF > $verityConfiguration
+[Unit]
+After=espmountmonitor.service
```
cwize1, line 19: You can use `Before` and `RequiredBy` / `WantedBy` instead of `After` and `Requires` to specify the same dependencies but from the opposite service/mount file.

toolkit/resources/dracutmodules/90mountesppartition/mountesppartition-generator.sh
```diff
@@ -0,0 +1,77 @@
...
+set -x
+set -e
+
+echo "Running mountesppartition-generator.sh" > /dev/kmsg
+
+# type getarg > /dev/null 2>&1 || . /lib/dracut-lib.sh
+
+function updateVeritySetupUnit () {
+    systemdDropInDir=/etc/systemd/system
+    verityDropInDir=$systemdDropInDir/systemd-veritysetup@root.service.d
```
cwize1, line 12: We shouldn't hardcode this to root verity. Use `veritysetup-pre.target` instead of `systemd-veritysetup@root.service`.

toolkit/resources/dracutmodules/90mountesppartition/mountesppartition-generator.sh
```diff
@@ -0,0 +1,77 @@
...
+# -----------------------------------------------------------------------------
+function createEspPartitionMonitorUnit() {
+    local espPartitionMonitorCmd=$1
+
+    espMountMonitorName="espmountmonitor.service"
+    systemdDropInDir=/etc/systemd/system
+    espMountMonitorDir=$systemdDropInDir
+    espMountMonitorUnitFile=$espMountMonitorDir/$espMountMonitorName
+
+    cat <<EOF > $espMountMonitorUnitFile
```
cwize1, line 51: Create a `.mount` file instead of creating a `.service` file that calls `mount`.

cwize1 (changes requested):

toolkit/resources/dracutmodules/90mountesppartition/10-mountesppartition.conf
```diff
@@ -0,0 +1 @@
+add_dracutmodules+=" mountesppartition "
```
cwize1, line 1: There is also the [toolkit/tools/internal/resources](https://github.com/microsoft/azure-linux-image-tools/tree/main/toolkit/tools/internal/resources) directory, which embeds the files into the binary.

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
liulanze: Sure, removed, will add the instruction to the doc.

PR #211, opened 2025-04-24
Remove deadcode

romoh:
Title says it all

---

### **Checklist**
- [x] Tests added/updated


toolkit/tools/internal/network/network.go
```diff
@@ -4,166 +4,15 @@
 package network
 
 import (
-	"context"
```
cwize1, line 7: It might be worth running `go mod tidy` to see if any dependencies are now gone.

cwize1 (dismissed):

toolkit/tools/internal/network/network.go
romoh: no changes

cwize1 (approved):

PR #215, opened 2025-04-29
MIC: De-duplicate files in LiveOS ISOs.

gmileka:
- Smaller initrd
  - Stop embedding kernel+shim+bootloader into initrd
    - This was a work around to be able to use isomaker as-is. The fix is stop using isomaker and create the boot image and call mkisofs directly.
- Smaller rootfs
  - Remove the re-created initrd from the squashfs.
    - The initrd is always generated under chroot and is created within the customized rootfs. It was then copied to the iso media. The fix is to delete (move) it from the customized rootfs as it is not needed to be there.
  - Remove the boot folder entirely.
    - When the input image with all its partitions is mounted, we ended up copying all the contents including the files at /boot. However, those files are already on the iso media and do not need to remain in the squashfs image. The fix is to delete that folder entirely.

The size gains are:
|            | old   | new   |                                    |
|------------|-------|-------|------------------------------------|
| initrd.img |  47MB |  32MB | removal of kernel+shim+bootloader  |
| rootfs.img | 173MB | 138MB | removal of /boot and initrd.img    |
| Test iso   | 250MB | 200MB |                                    |

---

### **Checklist**
- [ ] Tests added/updated

### DIFF
--- a/toolkit/tools/internal/isogenerator/isogenerator.go
+++ b/toolkit/tools/internal/isogenerator/isogenerator.go
@@ -224,12 +224,17 @@ func setUpIsoGrub2Bootloader(info isoGenInfo) (err error) {
 	}
 	defer os.RemoveAll(extractedShimDir)
 
-	shimFileName := "bootx64.efi"
-	grubFileName := "grubx64.efi"
-
-	if runtime.GOARCH == "arm64" {
+	shimFileName := ""
+	grubFileName := ""
+	switch runtime.GOARCH {
+	case "arm64":
 		shimFileName = "bootaa64.efi"
 		grubFileName = "grubaa64.efi"
+	case "amd64":
+		shimFileName = "bootx64.efi"
+		grubFileName = "grubx64.efi"
+	default:
+		return fmt.Errorf("failed to determine shim/grub efi file names. Unsupported host architecture (%s)", runtime.GOARCH)
 	}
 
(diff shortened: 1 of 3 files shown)


toolkit/tools/internal/isogenerator/isogenerator.go
```diff
@@ -214,32 +216,63 @@ func setUpIsoGrub2Bootloader(info isoGenInfo) (err error) {
...
 	}
+	defer os.RemoveAll(extractedShimDir)
 
-	initrdGrubBootloaderFilePath := filepath.Join(initrdEFIBootDirectoryPath, grubBootloaderFile)
-	buildDirGrubEFIFilePath := filepath.Join(bootDirPath, grubBootloaderFile)
-	err = extractFromInitrdAndCopy(info, initrdGrubBootloaderFilePath, buildDirGrubEFIFilePath)
+	shimFileName := "bootx64.efi"
+	grubFileName := "grubx64.efi"
+
+	if runtime.GOARCH == "arm64" {
```
cwize1, line 230: Use a switch statement and error out if an arch isn't supported.

cwize1 (changes requested):

toolkit/tools/internal/isogenerator/isogenerator.go
gmileka: Fixed.

PR #216, opened 2025-04-29
Enable GPG checking when installing/updating packages.

cwize1:
Currently, GPG signature checking is disabled when installing/updating packages. This is less than ideal when downloading packages from external sources such as PMC (packages.microsoft.com).

This change enables GPG checking by default for both the base image's repo files and the repo files provided by the user. However, the user can disable the GPG checks for their provided repo files by setting `gpgcheck` and `repo_gpgcheck` to `0`.

GPG checking remains disabled for local directories since there is no way for the user to provide a GPG public key. Also, local directories are likely to contain a user's custom built packages anyway, which are unlikely to be signed.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/docs/imagecustomizer/api/cli.md
+++ b/docs/imagecustomizer/api/cli.md
@@ -86,13 +86,27 @@ Can be one of:
 
   The RPMs may either be in the directory itself or any subdirectories.
 
+  GPG signature checking is disabled for local directories.
+  If you wish to enable GPG signature checking, then use a repo file instead and set the
+  `gpgkey` field within the repo file.
+
 - `*.repo` file path: A path to a RPM repo definition file.
 
   The file name extension must be `.repo`.
 
-  Note: This file is not installed in the image during customization.
-  If that is also needed, then use `AdditionalFiles` to place the repo file within
+  If the repo file's `baseurl` or `gpgkey` fields contain a `file://` URL, then the
+  host's directories pointed to by the URL will be bind mounted into the chroot
+  environment and the URL will replaced with the chroot equivalent URL.
+
+  GPG signature checking is enabled by default.
+  If you wish to disable GPG checking, then set both `gpgcheck` and `repo_gpgcheck` to
+  `0` in the repo file.
+
+  The repo file will only be used during image customization and will not be added to
   the image.
+  If you want to add the repo file to the image, then use use
+  [additionalFiles](../api/configuration/os.md#additionalfiles-additionalfile) to place
+  the repo file under the `/etc/yum.repos.d` directory.
 
 This option can be specified multiple times.
 
--- a/docs/imagecustomizer/api/cli.md
+++ b/docs/imagecustomizer/api/cli.md
@@ -96,7 +96,7 @@ Can be one of:
 
   If the repo file's `baseurl` or `gpgkey` fields contain a `file://` URL, then the
   host's directories pointed to by the URL will be bind mounted into the chroot
-  environment and the URL will replaced with the chroot equivalent URL.
+  environment and the URL will be replaced with the chroot equivalent URL.
 
   GPG signature checking is enabled by default.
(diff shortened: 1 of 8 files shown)


docs/imagecustomizer/api/cli.md
```diff
@@ -86,13 +86,27 @@ Can be one of:
...
+
 - `*.repo` file path: A path to a RPM repo definition file.
 
   The file name extension must be `.repo`.
 
-  Note: This file is not installed in the image during customization.
-  If that is also needed, then use `AdditionalFiles` to place the repo file within
+  If the repo file's `baseurl` or `gpgkey` fields contain a `file://` URL, then the
+  host's directories pointed to by the URL will be bind mounted into the chroot
+  environment and the URL will replaced with the chroot equivalent URL.
```
gmileka, line 99: nit: will be replaced

toolkit/tools/internal/safemount/safemount.go
```diff
@@ -16,9 +17,9 @@ import (
 )
 
 type Mount struct {
-	target     string
-	isMounted  bool
-	dirCreated bool
+	target      string
+	isMounted   bool
+	fileCreated bool
```
gmileka, line 22: nit: sounds like this is `fileOrDirCreated`?

toolkit/tools/internal/safemount/safemount.go
```diff
@@ -47,13 +48,33 @@ func (m *Mount) newMountHelper(source, target, fstype string, flags uintptr, dat
...
+			}
+		} else {
+			// Create parent directory.
+			err = os.MkdirAll(filepath.Dir(target), os.ModePerm)
+			if err != nil {
+				return fmt.Errorf("failed to create mount target parent directory (%s):\n%w", target, err)
+			}
+
+			// Create placeholder file.
+			err = os.WriteFile(target, []byte(nil), os.ModePerm)
```
liulanze, line 71: `[]byte{}` simplify to create the empty file?

docs/imagecustomizer/api/cli.md
```diff
@@ -86,13 +86,27 @@ Can be one of:
...
+  If you wish to enable GPG signature checking, then use a repo file instead and set the
+  `gpgkey` field within the repo file.
+
 - `*.repo` file path: A path to a RPM repo definition file.
 
   The file name extension must be `.repo`.
 
-  Note: This file is not installed in the image during customization.
-  If that is also needed, then use `AdditionalFiles` to place the repo file within
+  If the repo file's `baseurl` or `gpgkey` fields contain a `file://` URL, then the
```
gmileka, line 97: nit: should we include a sample repo file?

gmileka (dismissed):

toolkit/tools/internal/safemount/safemount.go
cwize1: In Go, `nil` is considered to be an empty list. So, `[]byte(nil)` and `[]byte{}` are functionally the same thing.

toolkit/tools/internal/safemount/safemount.go
cwize1: Technically directories are a type of file. But `fileOrDirCreated` is probably clearer.

docs/imagecustomizer/api/cli.md
cwize1: Fixed

toolkit/tools/internal/safemount/safemount.go
cwize1: Fixed

docs/imagecustomizer/api/cli.md
cwize1: The bullet point is getting pretty long as it is. If I wanted to add example repo files, I'd probably need to reorganize the docs. Maybe create a concepts page for RPM repos or something.

PR #217, opened 2025-04-29
Move to trivy for license scanning

romoh:
`go-license` is not well maintained anymore: doesn't support golang 1.24 and pulls in dependencies with CVEs.
This PR moves to trivy scans instead.

---

### **Checklist**
- [ ] Tests added/updated


toolkit/scripts/tools.mk
```diff
@@ -10,10 +10,24 @@ $(call create_folder,$(BUILD_DIR)/tools)
...
 ######## GO TOOLS ########
 
-GOPATH = $(shell go env GOPATH)
-GOBINPATH = $(GOPATH)/bin
-
-GO_LICENSES_TOOL = $(GOBINPATH)/go-licenses
+# Scans go.mod for critical & high severity license violations which catches
+# Forbidden & Restricted licenses based on
+# https://trivy.dev/v0.61/docs/scanner/license
+LICENSE_SCAN_OUTPUT := ./out/LICENSES/imagecustomizer.json
```
cwize1, line 16: The metadata isn't sufficient. We need the full license files.

cwize1 (changes requested):

romoh:
Here's an example of the output. This also now includes LICENSE from this repo
image

cwize1:
> Here's an example of the output. This also now includes LICENSE from this repo image

I think the project name should be a directory and all the license files for that project placed under that directory.

toolkit/scripts/check-and-collect-licenses.sh
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env bash
```
cwize1, line 1: This script is complex enough that it should probably be written in Python.

docs/imagecustomizer/developer-guide.md
```diff
@@ -36,14 +39,10 @@ sudo go test -C ./toolkit/tools ./...
...
 
-   - [Azure Linux 2.0
-     core-efi](https://github.com/microsoft/CBL-Mariner/blob/2.0/toolkit/imageconfigs/core-efi.json)
-   - [Azure Linux 3.0
-     core-efi](https://github.com/microsoft/CBL-Mariner/blob/3.0/toolkit/imageconfigs/core-efi.json)
-   - [Azure Linux 2.0
-     core-legacy](https://github.com/microsoft/CBL-Mariner/blob/2.0/toolkit/imageconfigs/core-legacy.json)
-   - [Azure Linux 3.0
-     core-legacy](https://github.com/microsoft/CBL-Mariner/blob/3.0/toolkit/imageconfigs/core-legacy.json)
+   - [Azure Linux 2.0 core-efi](https://github.com/microsoft/CBL-Mariner/blob/2.0/toolkit/imageconfigs/core-efi.json)
```
cwize1, line 42: Change to `azurelinux` while you are at it.

toolkit/scripts/tools.mk
```diff
@@ -10,10 +10,9 @@ $(call create_folder,$(BUILD_DIR)/tools)
 
 ######## GO TOOLS ########
 
-GOPATH = $(shell go env GOPATH)
-GOBINPATH = $(GOPATH)/bin
-
-GO_LICENSES_TOOL = $(GOBINPATH)/go-licenses
+# Scans and include licenses for all the go tools
+license-scan:
```
cwize1, line 14: nit: Add `.PHONY` line

.github/workflows/build-and-test.yml
```diff
@@ -69,6 +69,12 @@ jobs:
           exit 1
         fi
 
+    - name: Install Trivy
+      run: |
+        wget https://github.com/aquasecurity/trivy/releases/download/v0.61.1/trivy_0.61.1_Linux-64bit.tar.gz
```
cwize1, line 74: Can we sha256 this file and ensure it matches the expected value? This just avoids the file from changing unexpectantly.

toolkit/scripts/tools.mk
```diff
@@ -80,19 +76,14 @@ go_ldflags := 	-X github.com/microsoft/azurelinux/toolkit/tools/internal/exe.Too
 
 # Matching rules for the above targets
 # Tool specific pre-requisites are tracked via $(go-util): $(shell find...) dynamic variable defined above
-$(TOOL_BINS_DIR)/%: $(go_common_files) $(GO_LICENSES_TOOL)
+$(TOOL_BINS_DIR)/%: $(go_common_files)
+	$(MAKE) license-scan
```
cwize1, line 80: Why not set `license-scan` as a dependency instead of recursively calling `make`?

cwize1 (changes requested):

PR #219, opened 2025-04-30
Minor refactor of verity logic.

cwize1:
This is a setup change for the upcoming 'reinitialize verity' change.

Firstly, break up the verity format helper function (`customizeVerityImageHelper`) into smaller functions. This makes the code a little cleaner. Also, the smaller function will be used in the 'reinitialize verity' change.

Secondly, change the `verityMetadata` variable to be a list everywhere, instead of a map in some places and a list in other places. Also, identify the verity devices by name instead of mount path.

---

### **Checklist**

- [x] Tests added/updated


liulanze (approved):
Thanks for this change. LGTM.

PR #225, opened 2025-05-05
MIC: Factor code out into `IsoArtifactsStore`

gmileka:
This change factors out the part that builds up a 'store' of files and properties from the input image - the `IsoArtifactsStore`.

The artifacts store can be built from either a full image (qcow, vhd(x)), or a LiveOS ISO.

Then, the store is a unified input from then on to subsequent stages of creating the final iso.

This way, should we support a new input format, we just need to write a new converter from that format to the artifact store.

Also this changes moves a couple of function to other parts of the code where they better match the functionality.

---

### **Checklist**
- [n/a] Tests added/updated
- [ok] Code conforms to style guidelines


toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -1020,30 +542,17 @@ func createLiveOSIsoImage(buildDir, baseConfigPath string, inputIsoArtifacts *Li
...
 	isoBuildDir := filepath.Join(buildDir, "liveosbuild")
-	isoArtifactsDir := filepath.Join(isoBuildDir, "artifacts")
-	isomakerBuildDir := filepath.Join(isoBuildDir, "isomaker-tmp")
-
-	// Ensure the artifacts staging directory is empty to begin with.
-	err = os.RemoveAll(isomakerBuildDir)
-	if err != nil {
-		return fmt.Errorf("failed to delete artifact staging directory (%s):\n%w", isomakerBuildDir, err)
-	}
+	// isoArtifactsDir := filepath.Join(isoBuildDir, "artifacts")
```
cwize1, line 545: Delete commented-out line?

cwize1 (dismissed):

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
gmileka: it's becoming a habit I forget these...

gmileka: fixed.

cwize1 (approved):

PR #230, opened 2025-05-06
MIC: Factor out iso grub.cfg processing and eliminate `LiveOSIsoBuilder` type

gmileka:
The last iteration of cleaning-up the LiveOS ISO generation code:
- factor out iso grub.cfg processing code.
- eliminate the `LiveOSIsoBuilder` abstraction to reduce state (state is now isolated in the `IsoArtifactsStore` type).

---

### **Checklist**
- [n/a] Tests added/updated


toolkit/tools/pkg/imagecustomizerlib/liveosisoartifactstore.go
```diff
@@ -57,6 +57,22 @@ type IsoArtifactsStore struct {
 	files *IsoFilesStore
 }
 
+func (b *IsoArtifactsStore) cleanUp() error {
+	var err error
+	if b.files != nil {
+		cleanupErr := os.RemoveAll(b.files.artifactsDir)
+		if cleanupErr != nil {
+			if err != nil {
```
cwize1, line 65: `err` is always `nil` here.

toolkit/tools/pkg/imagecustomizerlib/liveosisoconstants.go
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package imagecustomizerlib
+
+const (
```
cwize1, line 6: Why not place these in `liveosisoutils.go` to be alongside `bootloaderFilesConfig`?

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/liveosisoconstants.go
gmileka: fair - moved.

cwize1 (approved):

PR #231, opened 2025-05-07
Add support for reinitializing verity.

cwize1:
Add support for reading the existing verity settings from the base image and then reapplying them after OS customization.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"path/filepath"
 	"slices"
+	"strings"
 
 	"github.com/microsoft/azurelinux/toolkit/tools/imagecustomizerapi"
 	"github.com/microsoft/azurelinux/toolkit/tools/imagegen/diskutils"
@@ -295,6 +296,32 @@ func SystemdFormatCorruptionOption(corruptionOption imagecustomizerapi.Corruptio
 	default:
 		return "", fmt.Errorf("invalid corruptionOption provided (%s)", string(corruptionOption))
 	}
+}
+
+func ParseSystemdVerityOptions(options string) (imagecustomizerapi.CorruptionOption, error) {
+	corruptionOption := imagecustomizerapi.CorruptionOptionIoError
+
+	optionValues := strings.Split(options, ",")
+	for _, option := range optionValues {
+		switch option {
+		case "":
+			// Ignore empty string.
+
+		case "ignore-corruption":
+			corruptionOption = imagecustomizerapi.CorruptionOptionIgnore
+
+		case "panic-on-corruption":
+			corruptionOption = imagecustomizerapi.CorruptionOptionPanic
+
+		case "restart-on-corruption":
+			corruptionOption = imagecustomizerapi.CorruptionOptionRestart
+
+		default:
+			return "", fmt.Errorf("unknown verity option (%s)", option)
+		}
+	}
+
+	return corruptionOption, nil
 }
 
 func validateVerityDependencies(imageChroot *safechroot.Chroot) error {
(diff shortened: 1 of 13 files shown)


toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -297,6 +298,32 @@ func SystemdFormatCorruptionOption(corruptionOption imagecustomizerapi.Corruptio
 	}
 }
 
+func ParseSystemdVerityOptions(options string) (imagecustomizerapi.CorruptionOption, error) {
+	corruptionOption := imagecustomizerapi.CorruptionOptionIoError
+
+	optionValues := strings.Split(options, ",")
```
liulanze, line 304: The `systemd.verity_root_options=` or `systemd.verity_usr_options=` can not only take the corruption options, but there are a lot more other options, how will you plan to handle them? For example, in the LG dev branch, we have `systemd.verity_root_options=panic-on-corruption,root-hash-signature=<PATH>`, which will cause error once rebased this commit.

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -297,6 +298,32 @@ func SystemdFormatCorruptionOption(corruptionOption imagecustomizerapi.Corruptio
 	}
 }
 
+func ParseSystemdVerityOptions(options string) (imagecustomizerapi.CorruptionOption, error) {
```
liulanze, line 301: nit - `parseSystemdVerityOptions(...)`, also, should we add new functions at the end of file?

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -426,15 +428,6 @@ func customizeOSContents(ic *ImageCustomizerParameters) error {
 		ic.config.OS = &imagecustomizerapi.OS{}
 	}
 
-	// Check if dm-verity is enabled on the base image.
-	verityEnabled, err := isDmVerityEnabled(ic.rawImageFile)
-	if err != nil {
-		return err
-	}
-	if verityEnabled && !ic.config.CustomizePartitions() {
-		return fmt.Errorf("dm-verity is enabled on the base image so partitions must be specified")
```
liulanze, line 435: This logic should have been removed? I think your last commit about leveraging pre-configured partition for verity can support to remove this limitation already?

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -900,6 +902,32 @@ func customizeVerityImageHelper(buildDir string, config *imagecustomizerapi.Conf
 		return nil, fmt.Errorf("failed to get disk's (%s) sector size:\n%w", loopback.DevicePath(), err)
 	}
 
+	for _, metadata := range baseImageVerity {
```
liulanze, line 905: This root hash "replace only" logic is impressive.

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
cwize1: I'm placing it next to `ParseSystemdVerityOptions`, since they are they inverse of each other.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
cwize1: Thanks!

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
cwize1: No, in the other feature, the verity partitions are pre-created but verity is not actually initialized. Hence, the image isn't "verity enabled".

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
cwize1: We can expand this function to handle other options in the future. In particular, in your change, you'll want to expand this function to also output the value of `root-hash-signature` so that you reapply that setting when reinitialize verity is used.

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
cwize1: Fixed name

liulanze (approved):

PR #235, opened 2025-05-14
GitHub actions improvements.

cwize1:
1. Add ARM64 build.

2. Set the correct version tag based on the build type.

3. Create a binary tarball that includes the LICENSE files.

3. Publish the binary and container tarballs as GitHub action artifacts.

4. Minor bug fix for the `build-container.sh` script.

5. Ensure all GitHub actions and jobs are explicit with what permissions are required.

---

### **Checklist**

- [x] Tests added/updated


.github/workflows/binary-build.yml
```diff
@@ -0,0 +1,162 @@
...
+    inputs:
+      publishType:
+        required: true
+        type: string
+      arch:
+        required: true
+        type: string
+
+env:
+  EXPECTED_GO_VERSION: "1.24.1"
```
gmileka, line 20: where does the go version come from? Is it pre-installed?

.github/workflows/binary-build.yml
```diff
@@ -0,0 +1,162 @@
...
+    - name: Build container
+      run: |
+        set -x
+
+        CONTAINER_TAG="imagecustomizer:build"
+        ./repo/toolkit/tools/imagecustomizer/container/build-container.sh -t "$CONTAINER_TAG" -a "${{ inputs.arch }}"
+
+        docker image save "$CONTAINER_TAG" | gzip > "imagecustomizer.tar.gz"
+
+    - name: Upload version artifact
```
gmileka, line 145: curious, where are these being uploaded to?

gmileka (approved):

.github/workflows/binary-build.yml
cwize1: They are available as artifacts on the workflow run.

For example, https://github.com/microsoft/azure-linux-image-tools/actions/runs/15030930793?pr=235 in the `Artifacts` section.

.github/workflows/binary-build.yml
cwize1: This matches the version in the `go.mod` file. This helps prevent the Go version from being updated accidentally, since you have to update it here as well.

PR #237, opened 2025-05-16
Don't copy file ACLs.

cwize1:
Linux has the capability to set extended file permissions (ACLs) to handle cases where the simple user:group:world permissions model isn't sufficient. When the `cp` command is set to preserve permissions, it will also try to preserve the ACLs. However, if the destination filesystem doesn't support ACLs, it will error out.

This change fixes the problem so that only basic permissions are preserved, instead of basic and ACLs. Unfortunately, `cp` doesn't have an option to only preserve basic permissions, it's either all or nothing. So, this change replaces the use of `cp` with direct I/O calls within Go.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/internal/file/filecopybuilder.go
+++ b/toolkit/tools/internal/file/filecopybuilder.go
@@ -102,7 +102,7 @@ func (b FileCopyBuilder) Run() (err error) {
 	dstFileMode := b.FileMode
 	if !b.ChangeFileMode {
 		// Copy the source file's permissions.
-		dstFileMode = srcFileInfo.Mode().Perm()
+		dstFileMode = srcFileInfo.Mode()
 	}
 
 	err = CreateDestinationDir(b.Dst, b.DirFileMode)
--- a/toolkit/tools/internal/file/filecopybuilder.go
+++ b/toolkit/tools/internal/file/filecopybuilder.go
@@ -5,11 +5,10 @@ package file
 
 import (
 	"fmt"
+	"io"
 	"os"
 
 	"github.com/microsoft/azurelinux/toolkit/tools/internal/logger"
-	"github.com/microsoft/azurelinux/toolkit/tools/internal/shell"
-	"github.com/sirupsen/logrus"
 )
 
 type FileCopyBuilder struct {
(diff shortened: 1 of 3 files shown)


toolkit/tools/internal/file/filecopybuilder.go
```diff
@@ -5,11 +5,10 @@ package file
 
 import (
```
gmileka, line 6: Chris, does this break selinux? since selinux attributes are stored in xattr?

Also, as I was working on creating an initrd from a full OS, I noticed that we need to retain things like owners and setuid/setgid/sticky bits or else booting will not complete successfully (was fun tracking what got lost on the way while copying in/out of initrd cpio).

Would it makes sense to make this behavior change parameterized? I think we want to keep as many attributes as possible unless specifically requested to ignore them...

cwize1: I don’t believe this breaks SELinux.

The existing file copy behavior only copies the permission bits and the ACLs. It doesn’t preserve other xattrs, since we don’t pass that flag. All this change does is to stop copying the ACLs.

Also, when SELinux is enabled, `setfiles` is called which will overwrite the SELinux labels for all the files in the OS. So, it doesn’t matter what the existing SELinux labels are.

For additionalFiles, we don’t really need to preserve anything other than the file contents and probably the execute bit.

For other use cases, there might be value in adding to option to preserve other properties.

gmileka: hmm... I see, I'm bit worried that when the default is to drop something, somebody will come along in the future, use the API and be surprised that something didn't copy over (I got freshly bitten by a different piece of code).
For this change, it looks like it won't affect the liveos iso dependency on `copyPartitionFiles()`.
I'll leave this to your discretion. I'll take a look at the rest of the code.

toolkit/tools/internal/file/filecopybuilder.go
```diff
@@ -56,49 +55,98 @@ func (b FileCopyBuilder) Run() (err error) {
...
+	defer func() {
+		if dstFile != nil {
+			dstFile.Close()
+		}
+	}()
 
-	args = append(args, "--preserve=mode", b.Src, b.Dst)
+	// The permissions given to OpenFile is subject to umask.
+	// So, apply the permissions to ensure they match exactly.
+	err = dstFile.Chmod(dstFileMode)
```
gmileka, line 126: Chris, go's Chmod seems to drop the higher order bits from file mode.
See: https://github.com/microsoft/azure-linux-image-tools/blob/8e1841d911a662996c253a8b1171fb295125c1e6/toolkit/tools/internal/initrdutils/initrdutils.go#L131

cwize1: The implementation of Go's chmod function seems to have logic to handle setuid, setgid, and setsticky: https://github.com/golang/go/blob/master/src/os/file_posix.go#L60

And I tried using `additionalFiles` with files that had the setuid, setgid, and setsticky bits set and it worked.

So, I am not sure why that comment exists in `initrdutils.go`.

gmileka: thx Chris - I'll look more into why it didn't work for me.

gmileka (approved):

PR #238, opened 2025-05-16
Create release workflow.

cwize1:
Create a GitHub action workflow for publishing a new minor version.

This workflow automatically handles:

- Building the binary and container.
- Publishing the container.
- Forking the release branch.
- Tagging the release commit.
- Creating a GitHub release, which contains the binaries.
- Creating the bump version commit for main. Unfortunately, policy prevents workflows from creating a PR. But just creating the commit is still useful.

In the future, this pipeline will run all the tests before publishing the release.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/.github/workflows/open-bump-version-pr.yml
+++ b/.github/workflows/open-bump-version-pr.yml
@@ -1,0 +1,57 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+name: Fork release branch
+
+permissions:
+  # Create release branch and publish release.
+  contents: write
+  # Publish PR.
+  #pull-requests: write
+
+on:
+  workflow_call: {}
+
+jobs:
+  fork-branch:
+    name: Open bump version PR
+    runs-on: ubuntu-latest
+    permissions:
+      # Create release branch and publish release.
+      contents: write
+      # Publish PR.
+      #pull-requests: write
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+      with:
+        path: repo
+
+    - name: Open bump version PR
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: |
+        set -x
+
+        git config --global user.name 'github-actions[bot]'
+        git config --global user.email 'github-actions[bot]@users.noreply.github.com'
+
+        pushd ./repo
+
+        # Switch to main branch.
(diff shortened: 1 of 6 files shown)


.github/workflows/publish-container.yml
```diff
@@ -0,0 +1,52 @@
...
+
+    - name: Login to GHCR
+      run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u "${{ github.actor }}" --password-stdin
+
+    - name: Publish container
+      run: |
+        set -x
+        VERSION=$(<"out/version/version.txt")
+
+        DOCKER_FULL_PATH="ghcr.io/${{ github.repository_owner }}/imagecustomizer:$VERSION"
```
romoh, line 33: side note - we should probably update the github docs to reference this acr instead (once we have one release with it out)

.github/workflows/open-bump-version-pr.yml
```diff
@@ -0,0 +1,57 @@
...
+
+        # Bump version.
+        NEXT_MINOR_VERSION="$(python3 .github/workflows/scripts/bump_minor_version.py)"
+
+        # Commit and push.
+        BUMP_VERSION_BRANCH="workflows/bumpVersion${NEXT_MINOR_VERSION}"
+        git checkout -b "${BUMP_VERSION_BRANCH}"
+
+        git add -A
+        git commit -m "Bump version to v${NEXT_MINOR_VERSION}"
```
romoh, line 53: so what happens when we have major releases? should we have a flag in this pipeline that allow both?

romoh (changes requested):

.github/workflows/open-bump-version-pr.yml
cwize1: For the most part, the exact same logic applies when releasing new major versions. The one exception to this is creating the change to bump the major version. But bumping the major version is easy enough and rare enough, that I don't think it is worth it to create dedicated workflow logic for it.

.github/workflows/publish-container.yml
cwize1: Yes, will do.

.github/workflows/open-bump-version-pr.yml
romoh: Talked offline

romoh (approved):

PR #239, opened 2025-05-20
Image Customizer: Reinit root verity + hard-reset bootloader.

cwize1:
Fix a bug where Image Customizer fails if the user recustomizes a root verity enabled image, without customizing partitions, but still hard-resetting the bootloader. The bug was simply that the bootloader hard-reset logic didn't know how to handle `/dev/mapper` paths in `/etc/fstab` file.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/testdata/verity-reinit-bootloader-reset.yaml
+++ b/toolkit/tools/pkg/imagecustomizerlib/testdata/verity-reinit-bootloader-reset.yaml
@@ -1,0 +1,21 @@
+previewFeatures:
+- reinitialize-verity
+
+os:
+  bootloader:
+    resetType: hard-reset
+
+  kernelCommandLine:
+    extraCommandLine:
+    - "rd.info"
+
+  additionalFiles:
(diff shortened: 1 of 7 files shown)


jiria (dismissed):

toolkit/tools/pkg/imagecustomizerlib/testdata/verity-reinit-bootloader-reset.yaml
```diff
@@ -0,0 +1,21 @@
...
+    - "rd.info"
+
+  additionalFiles:
+  - content: |
+      cat, dog, elephant, squirrel
+    destination: /usr/local/share/animals.txt
+
+  users:
+  - name: root
+    password:
```
jiria, line 19: Should we remove this even from the test code?

cwize1: Yes, these were left over from manual testing.

romoh (approved):

PR #241, opened 2025-05-21
Add initial OTEL instrumentation logic in imagecustomizer

aditjha-msft:
This PR introduces support for using the OTEL Go sdk to instrument the tool using OTEL standards and practices. It provides initialization logic needed for a global tracer to be used across the tool's execution. For starters, the PR adds a customization time span, with more telemetry to be added in follow up PRs that help with data collection tied to the tool's KPIs. 

The PR also adds logic to disable telemetry, and uses a no-op tracer when `--disable-telemetry` is passed in. This follows OTEL standardized method of disabling telemetry collection in a clean manner.

As part of telemetry initialization, the `telemetry.go` file has been added to set up the OTEL tracer and the OTLP grpc exporter. The exporter will send data to a Python service listening at the local port, which will then forward the data to Azure Monitor. 

In the effort of splitting up PRs in chunks, the PR for the complement Python telemetry service is here: https://github.com/microsoft/azure-linux-image-tools/pull/249.

---

### **Checklist**
- [x] Tests added/updated


toolkit/tools/pkg/imagecustomizerlib/telemetry.go
```diff
@@ -0,0 +1,78 @@
...
+	"go.opentelemetry.io/otel/sdk/resource"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/trace"
+	"go.opentelemetry.io/otel/trace/noop"
+)
+
+var (
+	Tracer trace.Tracer
+
+	TracerName = "imagecustomizer"
```
himaja-kesari, line 21: define them as constants ? const (
    TracerName             = "imagecustomizer"
    TelemetryServerAddress = "localhost:4317"
)

toolkit/tools/imagecustomizer/main.go
```diff
@@ -64,6 +67,17 @@ func main() {
 
 	logger.InitBestEffort(ptrutils.PtrTo(cli.LogFlags.AsLoggerFlags()))
 
+	// initialize OpenTelemetry tracer
+	err := imagecustomizerlib.InitTelemetry(cli.DisableTelemetry)
+	if err != nil {
+		logger.Log.Tracef("Failed to initialize telemetry setup: %v", err)
```
himaja-kesari, line 73: should we warn if we failed to initialize telemetry

toolkit/tools/pkg/imagecustomizerlib/telemetry.go
```diff
@@ -0,0 +1,78 @@
...
+		sdktrace.WithBatcher(exporter),
+		sdktrace.WithResource(resource.NewSchemaless(
+			attribute.String("host.architecture", runtime.GOARCH),
+			attribute.String("host.os", distro),
+			attribute.String("host.os.version", version),
+			attribute.String("imagecustomizer.version", ToolVersion),
+		)),
+	)
+	otel.SetTracerProvider(tp)
+	Tracer = tp.Tracer(TracerName)
```
cwize1, line 58: `otel.GetTracerProvider()` will return a no-op provider if `otel.SetTracerProvider` hasn't been called. In addition, `tp.Tracer` will return the same tracer if passed the same name.

So, in library code, you can just call `otel.GetTracerProvider().Tracer(name)`. There is no need to store the tracer in a global variable.

Also, looking at [the docs](https://pkg.go.dev/go.opentelemetry.io/otel/trace#TracerProvider), I think the tracer's name should match the name of the Go package, not the name of the executable.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -192,6 +195,15 @@ func CustomizeImageWithConfigFile(buildDir string, configFile string, inputImage
 	var err error
 
 	var config imagecustomizerapi.Config
+
+	ctx := context.Background()
+	_, span := Tracer.Start(ctx, "customizeImage")
```
cwize1, line 200: Just pass `context.Background()` to `Start`. No need for the `ctx` variable.

cwize1, line 200: `CustomizeImage` is probably the more interesting function to trace. (Though you can trace both if you really want to include parsing the YAML file.)

toolkit/tools/pkg/imagecustomizerlib/telemetry.go
```diff
@@ -0,0 +1,78 @@
+package imagecustomizerlib
```
cwize1, line 1: Place this code in its package under the `toolkit/tools/internal` directory.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/telemetry.go
```diff
@@ -0,0 +1,78 @@
...
+	"go.opentelemetry.io/otel/trace"
+	"go.opentelemetry.io/otel/trace/noop"
+)
+
+var (
+	Tracer trace.Tracer
+
+	TracerName = "imagecustomizer"
+
+	TelemetryServerAddress = "localhost:4317"
```
cwize1, line 23: While we generally recommend that people run inside a container, there are some who don't. Also, dev testing is usually done using `go run` and `go test`. So, picking a fixed port number (`4317`) is probably not a good idea in case it is already being used for something else. This should probably be a CLI parameter.

toolkit/tools/pkg/imagecustomizerlib/telemetry.go
```diff
@@ -0,0 +1,78 @@
...
+func InitTelemetry(disableTelemetry bool) error {
+	if disableTelemetry || !checkServerListening() {
+		// Use noop tracer when telemetry is disabled
+		logger.Log.Trace("Disabled telemetry collection")
+		Tracer = noop.NewTracerProvider().Tracer(TracerName)
+		shutdownFn = nil
+		return nil
+	}
+
+	exporter, err := otlptracegrpc.New(context.Background(),
```
cwize1, line 37: There is the [autoexport](https://pkg.go.dev/go.opentelemetry.io/contrib/exporters/autoexport) package that allows the OTEL connection config to be controlled via environment variables. That is probably a better solution than hardcoding it.

If `OTEL_TRACES_EXPORTER` is set, then we can initialize the tracer.

toolkit/tools/go.mod
```diff
@@ -14,26 +14,44 @@ require (
...
+	go.opentelemetry.io/otel v1.36.0
+	go.opentelemetry.io/otel/sdk v1.36.0
+	golang.org/x/sys v0.33.0
 	gonum.org/v1/gonum v0.15.0
 	gopkg.in/alecthomas/kingpin.v2 v2.2.6
 	gopkg.in/ini.v1 v1.67.0
 	gopkg.in/yaml.v3 v3.0.1
 )
 
+require (
```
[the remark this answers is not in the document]
aditjha-msft: For some reason even after I ran it, there were no auto changes made to the mod file. I'm not sure why?

toolkit/tools/pkg/imagecustomizerlib/telemetry.go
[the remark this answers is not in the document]
aditjha-msft: placed as const in `toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go`

toolkit/tools/pkg/imagecustomizerlib/versionsOfToolDependencies.go
```diff
@@ -32,7 +32,7 @@ func logVersionsOfToolDeps() {
 	}
 
 	// Get distro and version
-	distro, version := getDistroAndVersion()
+	distro, version := GetDistroAndVersion()
```
aditjha-msft, line 35: I had to make this publicly accessible to use for setting resource attributes

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -305,6 +319,9 @@ func CustomizeImage(buildDir string, baseConfigPath string, config *imagecustomi
 
 	logger.Log.Infof("Success!")
 
+	elapsedTime := time.Since(startTime)
+	span.SetAttributes(attribute.Float64("customizationTime", elapsedTime.Seconds()))
```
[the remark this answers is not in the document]
aditjha-msft: good point, and app insights already calculates the duration so not needed

toolkit/tools/internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,73 @@
...
+
+	exporter, err := otlptracegrpc.New(context.Background(),
+		otlptracegrpc.WithInsecure(),
+		otlptracegrpc.WithEndpoint(TelemetryServerAddress),
+	)
+	if err != nil {
+		return fmt.Errorf("failed to create OTLP exporter: %w", err)
+	}
+
+	distro, version := imagecustomizerlib.GetDistroAndVersion()
```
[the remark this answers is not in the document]
aditjha-msft: moved to its own package

toolkit/tools/pkg/imagecustomizerlib/telemetry.go
aditjha-msft: Combining with your other comment about the `OTEL_TRACES_EXPORTER` env var, maybe it'll be better if that env var is meant to hold the server address, and instead of adding a CLI parameter we just use the env var value? What do you think about that?

cwize1: Yes, I added this comment before I discovered the env variables as an option.

toolkit/tools/imagecustomizer/main.go
```diff
@@ -38,14 +40,16 @@ type InjectFilesCmd struct {
...
 type RootCmd struct {
-	Customize     CustomizeCmd     `name:"customize" cmd:"" default:"withargs" help:"Customizes a pre-built Azure Linux image."`
-	InjectFiles   InjectFilesCmd   `name:"inject-files" cmd:"" help:"Injects files into a partition based on an inject-files.yaml file."`
-	Version       kong.VersionFlag `name:"version" help:"Print version information and quit"`
-	TimeStampFile string           `name:"timestamp-file" help:"File that stores timestamps for this program."`
+	Customize        CustomizeCmd     `name:"customize" cmd:"" default:"withargs" help:"Customizes a pre-built Azure Linux image."`
+	InjectFiles      InjectFilesCmd   `name:"inject-files" cmd:"" help:"Injects files into a partition based on an inject-files.yaml file."`
+	Version          kong.VersionFlag `name:"version" help:"Print version information and quit"`
+	TimeStampFile    string           `name:"timestamp-file" help:"File that stores timestamps for this program."`
+	DisableTelemetry bool             `name:"disable-telemetry" help:"Disable telemetry collection of the tool."`
```
cwize1, line 47: Considering that `OTEL_EXPORTER_OTLP_ENDPOINT` has to be specified, we probably don't need `--disable-telemetry` in the exe now.

toolkit/tools/internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,56 @@
...
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/trace/noop"
+)
+
+var shutdownFn func(ctx context.Context) error
+
+func InitTelemetry(disableTelemetry bool) error {
+	if disableTelemetry || os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT") == "" {
+		logger.Log.Trace("Disabled telemetry collection")
+		otel.SetTracerProvider(noop.NewTracerProvider())
```
cwize1, line 25: `GetTracerProvider()` initializes to the no-op provider. So, you don't need to do anything here.

toolkit/tools/go.mod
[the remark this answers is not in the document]
cwize1: I think once there are two lists, `go mod tidy` won't merge them. However, if you delete both lists and re-run `go mod tidy`, then you'll get only 1 list.

cwize1 (changes requested):

cwize1 (approved):

PR #250, opened 2025-05-28
GitHub Actions: Automatically detect when release is the latest.

cwize1:
Add logic that automatically detects when a release is the latest release. This is used both for marking the GitHub release as the latest release and for knowing when to update the `stable` branch.

This change also refactors the GitHub Python helper scripts so that they share common functionality.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/.github/workflows/scripts/is_latest_release.py
+++ b/.github/workflows/scripts/is_latest_release.py
@@ -1,0 +1,27 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+import argparse
+
+from version_utils import getVersionTags, tryParseVersion
+
+def main():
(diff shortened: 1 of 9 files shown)


.github/workflows/scripts/is_latest_release.py
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+import argparse
+
+from version_utils import getVersionTags, tryParseVersion
```
gmileka, line 6: What is a 'latest release'?

When we release a build, isn't that by definition the latest released since we only move versions forward?

gmileka (approved):

PR #257, opened 2025-06-03
Add osPackages field in metadata for COSI 1.1

elainezhao1:
---
This is P1 for COSI 1.1 support (bootloader related will be P2)

changes:

- updated doc (referenced from trident cosi 1.1 revision doc)
- updated metadata to include the new field
- updated couple field names (verityConfig,  FileSystem) to match the updated cosi 1.1 schema

tested locally
image

### **Checklist**
- [ ] Tests added/updated


toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -303,3 +323,39 @@ func getArchitectureForCosi() string {
...
+		)
+		if err != nil {
+			return fmt.Errorf("failed to list installed RPMs:\n%w", err)
+		}
+
+		lines := strings.Split(strings.TrimSpace(out), "\n")
+		for _, line := range lines {
+			parts := strings.Fields(line)
+			if len(parts) != 4 {
+				logger.Log.Infof("Skipping malformed RPM line while parsing installed RPMs for COSI: %q", line)
```
frhuelsz, line 342: not an expert on customizer's logging standards, but intuition tells me a warning or even error could be better candidates for this case

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -146,11 +165,12 @@ func buildCosiFile(sourceDir string, outputFile string, partitions []outputParti
 	}
 
 	metadata := MetadataJson{
-		Version:   "1.0",
-		OsArch:    getArchitectureForCosi(),
-		Id:        imageUuidStr,
-		Images:    make([]Image, len(imageData)),
-		OsRelease: osRelease,
+		Version:    "1.1",
```
frhuelsz, line 168: let's keep this at 1.1 until the entire spec is implemented, otherwise we'll have some compatibility issues

toolkit/tools/pkg/imagecustomizerlib/cosimetadata.go
```diff
@@ -1,23 +1,24 @@
...
-	OsArch    string  `json:"osArch"`
-	Images    []Image `json:"images"`
-	OsRelease string  `json:"osRelease"`
-	Id        string  `json:"id"`
+	Version    string       `json:"version"`
+	OsArch     string       `json:"osArch"`
+	Images     []FileSystem `json:"images"`
+	OsRelease  string       `json:"osRelease"`
+	Id         string       `json:"id"`
+	OsPackages []OsPackage  `json:"osPackages,omitempty"`
```
frhuelsz, line 9: nit: perhaps remove the `omitempty` ?

toolkit/tools/pkg/imagecustomizerlib/cosimetadata.go
```diff
@@ -28,3 +29,10 @@ type ImageFile struct {
 	UncompressedSize uint64 `json:"uncompressedSize"`
 	Sha384           string `json:"sha384"`
 }
+
+type OsPackage struct {
+	Name    string `json:"name"`
+	Version string `json:"version"`
+	Release string `json:"release,omitempty"`
```
frhuelsz, line 36: nit: remove `omitempty` from these two as well?

frhuelsz:
Minor details, but otherwise the json looks good to me, thanks!

docs/imagecustomizer/api/cosi.md
```diff
@@ -14,101 +21,110 @@ The COSI file MUST be an uncompressed tarball. The file extension SHOULD be `.co
 The tarball MUST contain the following files:
 
 - `metadata.json`: A JSON file that contains the metadata of the COSI file.
-- Partition image files in the folder `images`: The actual partition images
-  that will be used to install the OS.
+- Filesystem image files in the folder `images/`: The actual filesystem images
+  that Trident will use to install the OS.
+
+To allow for future extensions, the tarball MAY contain other files, but Trident
```
cwize1, line 27: When adding this file to this repo, some of the contents were changed to make the file more generic. For example, using the terms "writer" and "reader" instead of specific product names.

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -303,3 +323,39 @@ func getArchitectureForCosi() string {
...
+
+	err := imageConnection.Chroot().UnsafeRun(func() error {
+		out, _, err := shell.Execute(
+			"rpm", "-qa", "--queryformat", "%{NAME} %{VERSION} %{RELEASE} %{ARCH}\n",
+		)
+		if err != nil {
+			return fmt.Errorf("failed to list installed RPMs:\n%w", err)
+		}
+
+		lines := strings.Split(strings.TrimSpace(out), "\n")
```
cwize1, line 338: Pull the parsing logic outside the `UnsafeRun` call.

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -303,3 +323,39 @@ func getArchitectureForCosi() string {
...
+		for _, line := range lines {
+			parts := strings.Fields(line)
+			if len(parts) != 4 {
+				logger.Log.Warnf("Skipping malformed RPM line while parsing installed RPMs for COSI: %q", line)
```
cwize1, line 342: I would error out here instead. For Image Customizer, it is better to error out and have customers complain about problems so that we can fix them, than silently ignoring a problem.

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -47,8 +49,20 @@ func convertToCosi(ic *ImageCustomizerParameters) error {
 		defer os.Remove(path.Join(outputDir, partition.PartitionFilename))
 	}
 
+	buildDir := filepath.Join(outputDir, "build")
+	existingImageConnection, _, _, err := connectToExistingImage(ic.rawImageFile, buildDir, "imageroot", true)
```
cwize1, line 53: Calling `connectToExistingImage` here could cause problems, since it might invalidate the verity partitions. It would probably be better to run the query in the `customizeImageHelper` function and then plumb the information here.

docs/imagecustomizer/api/cosi.md
```diff
@@ -14,101 +21,110 @@ The COSI file MUST be an uncompressed tarball. The file extension SHOULD be `.co
...
 
-If the tarball contains other files, readers MUST ignore them. A writer SHOULD NOT
-add any other files to the tarball.
+The tarball MUST NOT have a common root directory. The metadata file MUST be at
+the root of the tarball. If it were extracted with a standard `tar` invocation,
+the metadata file would be placed in the current directory.
 
-The tarball MUST NOT have a common root directory. The `metadata.json` file and the
-`images` directory MUST be in the root directory of the tarball.
+The metadata file SHOULD, be placed at the beginning of the tarball to allow for
```
cwize1, line 37: You should keep this as:

```
The metadata file SHOULD be placed at the beginning of the tarball to allow for
quick access to the metadata without needing to traverse the entire tarball.
```

It might also be good to backport the fixes.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -303,3 +306,38 @@ func getArchitectureForCosi() string {
...
+	var out string
+	err := imageConnection.Chroot().UnsafeRun(func() error {
+		var err error
+		out, _, err = shell.Execute(
+			"rpm", "-qa", "--queryformat", "%{NAME} %{VERSION} %{RELEASE} %{ARCH}\n",
+		)
+		if err != nil {
+			return fmt.Errorf("failed to list installed RPMs:\n%w", err)
+		}
+		return nil
```
cwize1, line 320: You already have a `fmt.Errorf` outside the `UnsafeRun` call. So, you don't need one inside the `UnsafeRun` call.

cwize1 (changes requested):

cwize1 (approved):

PR #262, opened 2025-06-05
Refactor legacy boot functional tests.

cwize1:
Stop taking legacy boot images as input. Instead, convert an UEFI image to legacy boot and use that to test legacy boot customization.

This change is being made in preparation of a future change that will allow the functional tests to be runnable against more image types.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/main_test.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/main_test.go
@@ -19,8 +19,7 @@ import (
 type baseImageType string
 
 const (
-	baseImageTypeCoreEfi    baseImageType = "core-efi"
-	baseImageTypeCoreLegacy baseImageType = "core-legacy"
+	baseImageTypeCoreEfi baseImageType = "core-efi"
 )
 
 type baseImageVersion string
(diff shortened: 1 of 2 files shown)


himaja-kesari (approved):

toolkit/tools/pkg/imagecustomizerlib/main_test.go
```diff
@@ -35,10 +34,8 @@ const (
 )
 
 var (
-	baseImageCoreEfiAzl2    = flag.String("base-image-core-efi-azl2", "", "A core-efi 2.0 image to use as a base image.")
-	baseImageCoreEfiAzl3    = flag.String("base-image-core-efi-azl3", "", "A core-efi 3.0 image to use as a base image.")
-	baseImageCoreLegacyAzl2 = flag.String("base-image-core-legacy-azl2", "", "A core-legacy 2.0 image to use as a base image.")
```
gmileka, line 40: Are we okay not testing the legacy images we produce from our build pipeline?
Theoretically, uefi->legacy->legacy->uefi should cover it. But I'd vote for testing whatever that is dropped in our artifacts as well (just in case - who knows what differences might be there).

Britel: I'd step back and ask what should come out of the seeder / imagerv2. Should that always build uefi, and then any one who wants to build legacy goes through MIC?

That way - we'd have one code.

cwize1: Yeah, there is the possibility for differences, at least until image customizer takes over the building of the variant images. But legacy boot is kind of a niche scenario which ideally no-one should be using anymore. So, I am not sure it is worth keeping the complexity of accepting legacy images as input.

PR #263, opened 2025-06-05
Don't apply noexec mount flag during customization.

cwize1:
If the user applies the `noexec` to a partition's mount options, then add `noexec` to the `/etc/fstab` file but don't apply `noexec` during customization. This allows scripts like `grub2-mkconfig` to run during customization.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/testdata/verity-usr-uki.yaml
+++ b/toolkit/tools/pkg/imagecustomizerlib/testdata/verity-usr-uki.yaml
@@ -20,7 +20,10 @@ storage:
       size: 100M
 
     - id: root
-      size: 2G
+      size: 500M
+
+    - id: var
+      size: 1G
 
   verity:
   - id: verityusr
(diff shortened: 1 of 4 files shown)


romoh (approved):

toolkit/tools/pkg/imagecustomizerlib/testdata/verity-usr-uki.yaml
```diff
@@ -42,14 +45,20 @@ storage:
 
   - deviceId: root
     type: ext4
-    mountPoint: /
+    mountPoint:
+      path: /
+      options: noexec
```
himaja-kesari, line 50: Do we have a test that verifies the prevention of execution of the binaries on the mounted filesystem.

Britel: +1

himaja-kesari (approved):

PR #285, opened 2025-06-24
Add bootloader metadata for COSI 1.1

elainezhao1:
Add bootloader metadata extraction logic support grub-based boot and multiple types of UKI-based boot:

- First checks for systemd-boot config entries under /boot/loader/entries.
- If none are found, checks for standalone UKI files in the ESP.
- For each detected entry, extracts kernel version and command-line arguments, and constructs the appropriate metadata.
- Fallback to GRUB-based detection remains unchanged.

tested locally with 3 cases:

- uki standalone :
Validated by passing a uki-enabled image and confirm correct parsing of embedded .cmdline.
image

- grub: 
image

- config
tested with a sample config 

```
  additionalFiles:
    - content: |
        title      Azure Linux 3.0
        version    6.6.78.1-3.azl3
          # Comment with leading spaces

        # Test quoted values
        options    "root=UUID=12345678-90ab-cdef-1234-567890abcdef"
        options    ro  quiet   splash
        options    console=tty0  console=ttyS0,115200n8
        options    loglevel=3   systemd.show_status=1

        # Empty line above with spaces/tabs

        # edge case: quotes inside value
        options    crashkernel="512M-32G:256M,32G-:512M"
      destination: /boot/loader/entries/sample.conf
```
and obtained
```
  "bootloader": {
    "type": "systemd-boot",
    "systemdBoot": {
      "entries": [
        {
          "type": "config",
          "path": "/boot/loader/entries/sample.conf",
          "cmdline": "root=UUID=12345678-90ab-cdef-1234-567890abcdef ro  quiet   splash console=tty0  console=ttyS0,115200n8 loglevel=3   systemd.show_status=1 crashkernel=\"512M-32G:256M,32G-:512M\"",
          "kernel": "6.6.78.1-3.azl3"
        }
      ]
    }
  },
```
(uki with config is not supported yet, will have a thorough test once it's supported)

---

### **Checklist**
- [ ] Tests added/updated


toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -345,3 +350,135 @@ func getAllPackagesFromChroot(imageConnection *ImageConnection) ([]OsPackage, er
...
+	// Check for systemd-boot with .conf entries (UKI with config)
+	loaderEntryDir := filepath.Join(chrootDir, "boot", "loader", "entries")
+	if exists, err := file.DirExists(loaderEntryDir); err != nil {
+		return nil, fmt.Errorf("failed while checking if systemd-boot entry dir '%s' exists:\n%w", loaderEntryDir, err)
+	} else if exists {
+		entries, err := extractSystemdBootEntries(loaderEntryDir)
+		if err != nil {
+			return nil, fmt.Errorf("failed to extract systemd-boot entries:\n%w", err)
+		}
+		return &CosiBootloader{
```
fintelia, line 366: Shouldn't this fallthrough if the directory exists but is empty?

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -345,3 +350,135 @@ func getAllPackagesFromChroot(imageConnection *ImageConnection) ([]OsPackage, er
...
+		entry := SystemDBootEntry{
+			Path: filepath.Join("boot", "loader", "entries", file.Name()),
+		}
+
+		lines := strings.Split(string(content), "\n")
+		for _, line := range lines {
+			line = strings.TrimSpace(line)
+			switch {
+			case strings.HasPrefix(line, "options"):
+				entry.Cmdline = strings.TrimSpace(strings.TrimPrefix(line, "options"))
```
fintelia, line 440: This should append to `entry.Cmdline` rather than replace it. The spec says:

> options shall contain kernel parameters to pass to the Linux kernel to spawn. This key is optional and may appear more than once in which case all specified parameters are combined in the order they are listed.

Also don't forget to include a space between concatenated items.

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -345,3 +350,135 @@ func getAllPackagesFromChroot(imageConnection *ImageConnection) ([]OsPackage, er
...
+			case strings.HasPrefix(line, "linux"):
+				linuxPath := strings.TrimSpace(strings.TrimPrefix(line, "linux"))
+				if kernelVer, err := getKernelVersion(filepath.Base(linuxPath)); err == nil {
+					entry.Kernel = kernelVer
+				}
+			case strings.HasPrefix(line, "uki"):
```
fintelia, line 446: The version of systemd-boot in AZL3 doesn't actually support this option. I think it is planned to land in systemd 258

frhuelsz:
> I haven’t tested two entry types yet: uki-config (a config file with a UKI) and config (a config file with kernel, initrd, and cmdline). Are there any existing images we can use to validate these cases?
> 
> @fintelia @frhuelsz

Apparently `uki-config` is not possible yet (as @fintelia noted in another comment, will be supported in systemd 258).

Regarding `config`, we are not sure it is supported by customizer yet, so we are not aware of any image that would serve as an example. Perhaps @cwize1 might know?

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/cosimetadata.go
```diff
@@ -1,12 +1,48 @@
...
-	Images     []FileSystem `json:"images"`
-	OsRelease  string       `json:"osRelease"`
-	Id         string       `json:"id"`
-	OsPackages []OsPackage  `json:"osPackages"`
+	Version    string          `json:"version"`
+	OsArch     string          `json:"osArch"`
+	Images     []FileSystem    `json:"images"`
+	OsRelease  string          `json:"osRelease"`
+	Id         string          `json:"id,omitempty"`
+	Bootloader *CosiBootloader `json:"bootloader"`
```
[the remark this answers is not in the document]
elainezhao1: I think this is required for cosi 1.1 based on their doc

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -345,3 +349,172 @@ func getAllPackagesFromChroot(imageConnection *ImageConnection) ([]OsPackage, er
...
+
+		lines := strings.Split(string(content), "\n")
+		for _, line := range lines {
+			line = strings.TrimSpace(line)
+			if line == "" || strings.HasPrefix(line, "#") {
+				continue
+			}
+
+			// Split the line into whitespace-separated fields.
+			fields := strings.Fields(line)
```
fintelia, line 470: This parsing doesn't seem to match the spec. Only space characters between the key and the value are stripped. Any other whitespace characters and whitespace occurring elsewhere in the line should be kept:

> These boot loader menu snippets shall be UNIX-style text files (i.e. lines separated by a single newline character), in the UTF-8 encoding. The boot loader menu snippets are loosely inspired by Grub1’s configuration syntax. Lines beginning with “#” are used for comments and shall be ignored. The first word of a line is used as key and is separated by one or more spaces from the value. The rest of the line contains the value, a literal string.

elainezhao1: Thanks for pointing out! I updated the code to only trims spaces between key and value, tested with a sample config and internal spacing in values are preserved (e.g "root=UUID=c3174f85-0000-4f7b-9230-d1f763ee4a9f ro quiet loglevel=5   audit=1    selinux=0 )

toolkit/tools/pkg/osmodifierlib/modifierutils.go
```diff
@@ -42,13 +39,17 @@ func doModifications(ctx context.Context, baseConfigPath string, osConfig *osmod
 	}
 
 	// Add a check to make sure BootCustomizer can be initialized
-	grubExists := grubConfigSupportExists(dummyChroot)
+	bootloaderType, err := imagecustomizerlib.DetectBootloaderType(dummyChroot)
```
cwize1, line 42: Previously this code modified the grub entries even if systemd-boot was enabled. This code change modifies this behavior so that either systemd-boot or grub is updated. Is this a desired change in behavior?

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -345,3 +350,135 @@ func getAllPackagesFromChroot(imageConnection *ImageConnection) ([]OsPackage, er
...
+		content, err := os.ReadFile(absPath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read %s: %w", absPath, err)
+		}
+
+		entry := SystemDBootEntry{
+			Path: filepath.Join("boot", "loader", "entries", file.Name()),
+		}
+
+		lines := strings.Split(string(content), "\n")
```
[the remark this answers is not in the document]
elainezhao1: leave for later

toolkit/tools/pkg/osmodifierlib/modifierutils.go
elainezhao1: Previously, my goal was to: 
On GRUB-based systems: Only GRUB configs are updated
On systemd-boot/UKI-based systems: grub config shoud not be updated  (e.g. selinux is only updated in /etc/selinux/config) 

but at that time I didn't know a robust way to detect bootloader type and old logic was relying on the presence of grub.cfg, which can be more permissive, so I switched package-based detection.  In an upcoming task, I will create tests for osmodifier to test multiple test images

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -153,12 +157,13 @@ func buildCosiFile(sourceDir string, outputFile string, partitions []outputParti
...
 
 	metadata := MetadataJson{
-		Version:    "1.0",
+		Version:    "1.1",
 		OsArch:     getArchitectureForCosi(),
 		Id:         imageUuidStr,
 		Images:     make([]FileSystem, len(imageData)),
 		OsRelease:  osRelease,
 		OsPackages: osPackages,
+		Bootloader: *cosiBootMetadata,
```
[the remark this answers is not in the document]
elainezhao1: I added helper function to convert nil to empty struct at this step, this way we can pass nil instead of passing CosiBootloader{} through many layers. e.g. customizeImageHelper just return nil and don't need to create empty struct. Does it make sense?

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -346,3 +351,199 @@ func getAllPackagesFromChroot(imageConnection *imageconnection.ImageConnection)
...
+			SystemdBoot: nil,
+		}, nil
+	}
+
+	return nil, fmt.Errorf("no supported bootloader configuration found")
+}
+
+func extractUkiEntriesIfPresent(chrootDir, buildDir string) ([]SystemDBootEntry, error) {
+	espDir := filepath.Join(chrootDir, "boot/efi")
+	cmdlines, err := extractKernelCmdlineFromUkiEfis(espDir, buildDir)
```
liulanze, line 401: This may break LG usage? The `objcopy dump` will tamper the UKI sha256 hash & signature. @fintelia Thoughts?

toolkit/tools/pkg/imagecustomizerlib/cosimetadata.go
```diff
@@ -1,12 +1,44 @@
 package imagecustomizerlib
 
+type BootloaderType string
```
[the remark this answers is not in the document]
cwize1: This seems like feature creep. We can move Go type definition if and when we add support for `type` under `bootloader` in the API.

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
cwize1: We really need to fix that. Would it make sense to copy the UKIs to temporary files before calling `objcopy` so that the modifications only affect the temporary copy?

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -397,8 +397,37 @@ func extractCosiBootMetadata(buildDirAbs string, imageConnection *imageconnectio
 }
 
 func extractUkiEntriesIfPresent(chrootDir, buildDir string) ([]SystemDBootEntry, error) {
-	espDir := filepath.Join(chrootDir, "boot/efi")
-	cmdlines, err := extractKernelCmdlineFromUkiEfis(espDir, buildDir)
+	espDir := filepath.Join(chrootDir, EspDir)
+	linuxDir := filepath.Join(espDir, UkiOutputDir)
+
+	tempDir, err := os.MkdirTemp(buildDir, "uki-temp-*")
```
cwize1, line 403: Add a comment saying why these files are being copied.

cwize1, line 403: Should the temp file copy be handled in `extractKernelCmdlineFromUkiEfis` so that it fixes the issue for all users of `extractKernelCmdlineFromUkiEfis`?

Actually, it might be good handle the temp file copy and the `objcopy` call in a single helper function. That way, if we find an alternative to `objcopy`, it is easy to replace the code.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
elainezhao1: @cwize1 A separate question, to your knowledge, do you think coping efi files will affect shrinking filesystems? I'm doing some local testing and see  

```
INFO[0005][main] Shrinking partition (/dev/loop17p2)
DEBU[0005][main] Executing: [e2fsck -fy /dev/loop17p2]
DEBU[0005][main] e2fsck 1.46.5 (30-Dec-2021)
DEBU[0005][main] /dev/loop17p2 has unsupported feature(s): FEATURE_C12
DEBU[0005][main] e2fsck: Get a newer version of e2fsck!
DEBU[0005][main]
DEBU[0005][main] /dev/loop17p2: ********** WARNING: Filesystem still has errors **********
DEBU[0005][main]
DEBU[0005][main] Detaching Loopback Device Path: /dev/loop17
DEBU[0005][main] Executing: [losetup -d /dev/loop17]
2025/07/11 00:08:05 image customization failed:
failed to customize raw image:
failed to shrink filesystems:
failed to shrink ext4 filesystem (/dev/loop17p2):
failed to check (/dev/loop17p2) with e2fsck:
exit status 12
```

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
elainezhao1: Will mark this comment as resolved, talked with Lanze, image tools dev build is broken now so LG pipeline cannot be run for this change, will circle back with Lanze

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
elainezhao1: updated, ready for another review

cwize1 (approved):

liulanze (approved):

PR #286, opened 2025-06-24
Image Customizer: Fix SSH authorized_keys permissions.

cwize1:
1. Don't copy the provided SSH public keys into the `.ssh` directory. Instead, only copy the values inside the files into the `authorized_keys` files.

2. Write directly to the `authorized_keys` file instead of using temporary file.

3. Only open the `authorized_keys` file once when writing to it.

4. Don't change the file ownership/permissions of any existing files within the `.ssh` directory.

5. Use the inbuilt Go functions for setting file ownership/permissions, instead of Shell commands.

6. Ensure the file permissions are correct. (600 for files. 700 for directories.)

---

### **Checklist**

- [x] Tests added/updated


toolkit/tools/internal/userutils/userutils.go
```diff
@@ -206,3 +211,168 @@ func PasswordExpiresDaysIsValid(passwordExpiresDays int64) error {
...
+		return fmt.Errorf("failed to get user's (%s) home directory:\n%w", username, err)
+	}
+
+	userSSHKeyDirFullPath := filepath.Join(installChroot.RootDir(), userSSHKeyDir)
+	authorizedKeysFullPath := filepath.Join(userSSHKeyDirFullPath, SSHAuthorizedKeysFileName)
+
+	allSSHKeys := []string(nil)
+
+	// Add existing keys in the authorized_keys file, if requested.
+	if includeExistingKeys {
```
aditjha-msft, line 279: just curious, when would `includeExistingKeys` be set to true? Is it set to true anywhere currently?

aditjha-msft (approved):

gmileka (approved):

PR #287, opened 2025-06-24
Add unittests for previewFeature

### LINKED ISSUE #279: Add unit tests for PreviewFeature
Not tests for PreviewFeature under toolkit/tools/imagecustomizerapi/previewfeaturetype.go. Need to add unit tests to improve code coverage

romoh:
Improve code coverage

fixes #279 

---

### **Checklist**
- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/imagecustomizerapi/previewfeaturetype_test.go
+++ b/toolkit/tools/imagecustomizerapi/previewfeaturetype_test.go
@@ -1,0 +1,35 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package imagecustomizerapi
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPreviewFeature_IsValid(t *testing.T) {
+	validFeatures := []PreviewFeature{
+		PreviewFeatureUki,
+		PreviewFeatureOutputArtifacts,
+		PreviewFeatureInjectFiles,
+		PreviewFeaturePackageSnapshotTime,
+	}
+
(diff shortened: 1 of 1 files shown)


toolkit/tools/imagecustomizerapi/previewfeaturetype_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+	for _, feature := range validFeatures {
+		err := feature.IsValid()
+		assert.NoError(t, err, "expected no error for valid feature: %s", feature)
+	}
+
+	invalidFeature := PreviewFeature("invalid-feature")
+	err := invalidFeature.IsValid()
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "invalid-feature")
```
cwize1, line 28: Use `assert.ErrorContains` instead of `assert.Error` and `assert.Contains`.

toolkit/tools/imagecustomizerapi/previewfeaturetype_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+	// Test PreviewFeatureReinitializeVerity is not valid (since it's not typed as PreviewFeature)
```
cwize1, line 30: The reason why this fails is because `PreviewFeatureReinitializeVerity` is missing in the `PreviewFeature.IsValid()` function. Go is happily doing the type-cast without issue.

Also, it looks like the `Config.PreviewFeatures` field isn't being checked for validity. Hence, why this wasn't causing any problems.

cwize1 (changes requested):

PR #293, opened 2025-06-26
Instrument Operation Flow for Image Customizer

aditjha-msft:
This PR enables detailed instrumentation of the customization flow using the otel sdk to trace operations involved in an image customization run. It requires context propagation in order to cleanly build parent-child relationships between spans, and having `ctx` passed around will enable us to capture errors in a follow up PR. 

There are also improvements to the telemetry setup logic in `telemetry.go` to follow standard otel conventions, and using lowercase snake case for span names and attributes to follow standard naming guidelines by otel. Read here if interested: [OTel Naming](https://opentelemetry.io/docs/specs/semconv/general/naming/)

Many files get touched in this PR solely because of the requirement to plumb the `ctx` throughout the code and related test files. However, having context propagation setup will be beneficial for the code in terms of future ctrl-c support as well as error capturing for telemetry.

The instrumentation logic in this PR allows us to trace the execution which can then be viewed in App Insights as such:
image

Passing Test Pipeline: https://dev.azure.com/mariner-org/ECF/_build/results?buildId=850701&view=results

---

### **Checklist**
- [x] Tests added/updated


toolkit/tools/pkg/imagecustomizerlib/imagecreator.go
```diff
@@ -85,38 +86,41 @@ func doOsCustomizationsImageCreator(
 		return err
 	}
 
+	ctx := context.Background()
+
 	if err = addRemoveAndUpdatePackages(
+		context.Background(),
```
cwize1, line 92: `ctx`?

toolkit/tools/pkg/imagecustomizerlib/imagecreator.go
```diff
@@ -85,38 +86,41 @@ func doOsCustomizationsImageCreator(
 		return err
 	}
 
+	ctx := context.Background()
```
cwize1, line 89: Any chance you can plumb the `ctx` variable all the way from the top-level `main` functions?

This is more morally correct. Also, it will make it easier to add Ctrl+C support in the future.

toolkit/tools/pkg/imagecustomizerlib/customizebootloader.go
```diff
@@ -34,12 +36,15 @@ func handleBootLoader(baseConfigPath string, config *imagecustomizerapi.Config,
...
 }
 
-func hardResetBootLoader(baseConfigPath string, config *imagecustomizerapi.Config, imageConnection *ImageConnection,
+func hardResetBootLoader(ctx context.Context, baseConfigPath string, config *imagecustomizerapi.Config, imageConnection *ImageConnection,
 	partUuidToFstabEntry map[string]diskutils.FstabEntry, newImage bool,
 ) error {
 	var err error
 	logger.Log.Infof("Hard reset bootloader config")
 
+	_, span := otel.GetTracerProvider().Tracer(OtelTracerName).Start(ctx, "reset_bootloader")
```
cwize1, line 45: In the future we will likely add support for a soft reset. So, it would be good to call this `hard_reset_bootloader`.

gmileka (dismissed):

toolkit/tools/pkg/imagecustomizerlib/customizeoverlays.go
```diff
@@ -29,6 +32,13 @@ func enableOverlays(overlays *[]imagecustomizerapi.Overlay, selinuxMode imagecus
 
 	logger.Log.Infof("Enable filesystem overlays")
 
+	_, span := otel.GetTracerProvider().Tracer(OtelTracerName).Start(ctx, "enable_overlays")
+	span.SetAttributes(
+		attribute.Int("overlays_count", len(*overlays)),
+		attribute.String("selinux_mode", string(selinuxMode)),
```
cwize1, line 38: The SELinux mode is likely logged elsewhere. So, you probably don’t need to log it here as well.

toolkit/tools/pkg/imagecustomizerlib/customizeselinux.go
```diff
@@ -104,6 +114,12 @@ func selinuxSetFiles(selinuxMode imagecustomizerapi.SELinuxMode, imageChroot *sa
 
 	logger.Log.Infof("Setting file SELinux labels")
 
+	_, span := otel.GetTracerProvider().Tracer(OtelTracerName).Start(ctx, "set_selinux_files")
+	span.SetAttributes(
+		attribute.String("selinux_mode", string(selinuxMode)),
```
cwize1, line 119: We probably don’t need to log the same information multiple times.

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -41,6 +44,12 @@ func enableVerityPartition(verity []imagecustomizerapi.Verity, imageChroot *safe
 
 	logger.Log.Infof("Enable verity")
 
+	_, span := otel.GetTracerProvider().Tracer(OtelTracerName).Start(ctx, "enable_verity_partition")
+	span.SetAttributes(
+		attribute.Int("verity_count", len(verity)),
```
cwize1, line 49: It might be good to log the name of the verity device, as it will tell us what type of verity is being used.

toolkit/tools/pkg/imagecustomizerlib/imagecreator.go
```diff
@@ -28,14 +29,14 @@ func CustomizeImageHelperImageCreator(buildDir string, baseConfigPath string, co
 	}
 	defer toolsChroot.Close(false)
 
-	imageConnection, partUuidToFstabEntry, _, err := connectToExistingImage(rawImageFile, toolsChrootDir, toolsRootImageDir, true, false)
+	imageConnection, partUuidToFstabEntry, _, err := connectToExistingImage(context.Background(), rawImageFile, toolsChrootDir, toolsRootImageDir, true, false)
```
cwize1, line 32: Use `ctx`?

toolkit/tools/pkg/imagecustomizerlib/rpmsourcesmounts.go
```diff
@@ -33,11 +36,17 @@ type rpmSourcesMounts struct {
...
 
-func mountRpmSources(buildDir string, imageChroot *safechroot.Chroot, rpmsSources []string,
+func mountRpmSources(ctx context.Context, buildDir string, imageChroot *safechroot.Chroot, rpmsSources []string,
 	useBaseImageRpmRepos bool,
 ) (*rpmSourcesMounts, error) {
 	var err error
 
+	_, span := otel.GetTracerProvider().Tracer(OtelTracerName).Start(ctx, "mount_rpm_sources")
+	span.SetAttributes(
+		attribute.Int("rpm_sources_count", len(rpmsSources)),
```
cwize1, line 46: It would be good to log `useBaseImageRpmRepos` as well.

toolkit/tools/pkg/imagecustomizerlib/runscripts.go
```diff
@@ -29,6 +32,13 @@ func runUserScripts(baseConfigPath string, scripts []imagecustomizerapi.Script,
 
 	logger.Log.Infof("Running %s scripts", listName)
 
+	_, span := otel.GetTracerProvider().Tracer(OtelTracerName).Start(ctx, "run_user_scripts")
+	span.SetAttributes(
+		attribute.Int("scripts_count", len(scripts)),
+		attribute.String("list_name", listName),
```
cwize1, line 38: I would change the name of the event to: `run_postcustomization_scripts` and `run_finalizeimage_scripts` instead of logging `listName`.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/rpmsourcesmounts.go
aditjha-msft: fixed

toolkit/tools/pkg/imagecustomizerlib/imagecreator.go
aditjha-msft: oops missed this one, thanks

toolkit/tools/pkg/imagecustomizerlib/runscripts.go
aditjha-msft: updated, thanks

toolkit/tools/pkg/imagecustomizerlib/customizeselinux.go
aditjha-msft: removed

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
cwize1: I'd log the list of `imagecustomizerapi.Verity.Name` fields. You could also add list of the `MountPath` fields.

aditjha-msft: updated. I actually figured its better to log this info in `customizeVerityImageHelper` as that is where we provision verity and the time changes based on root/usr. Let me know if you think thats fine.

toolkit/tools/pkg/imagecustomizerlib/verityutils.go
```diff
@@ -135,3 +137,12 @@ func roundUpToPowerOf2(n uint32) uint32 {
 	}
 	return res
 }
+
+func getVerityNames(verity []verityDeviceMetadata) []string {
+	names := make([]string, len(verity))
+	for i, v := range verity {
+		names[i] = v.name
+	}
+	verityNames := sliceutils.RemoveDuplicatesFromSlice(names)
```
cwize1, line 146: The names should be unique. So, `RemoveDuplicatesFromSlice` is unnecessary.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/verityutils.go
aditjha-msft: Oh I see, I noticed in the api docs that the name value can only ever be `root` or `usr`. So I thought if there are multiple usr verity entries in the config, then the result of this list would be `[usr, usr, usr]`. If there is more than one verity in the config, will it just be one `usr` and one `root` max?

aditjha-msft: anyways, updated to remove that function call

cwize1 (approved):

PR #297, opened 2025-07-01
Add arbitrary variant image support to imagecustomizer container

vinceaperri:
The purpose of this pull request is to enhance the imagecustomizer container's `run.sh` script to support the linuxguard
OCI artifact. The script now recognizes a new environment variable, `BASE_IMAGE_NAME`, allowing users to specify a base
image other than the default `minimal-os`. For backwards compatibility, `minimal-os` remains the default value.

Additional updates include support for a new variable, `DEVELOPER_MODE`, which switches the OCI artifact registry to the
Azure development container registry. By default, the script uses `mcr.microsoft.com`, ensuring backwards compatibility.
The script also now accepts any latest tag matching the pattern `latest-*` (such as `latest-preview`, `latest-amd64`, or
`latest-arm64`), not just `latest`, which improves flexibility for future linuxguard tags.

The script detects the host architecture via `uname -m` and pulls the corresponding image. If the architecture is
unrecognized, it defaults to `amd64`. While this may result in a failure if imagecustomizer does not support the host's
architecture, it is preferable to defer this failure rather than prematurely restrict the script.

A comprehensive usage/help string has been added to accommodate the expanded input options, which can be accessed with
`-h` or `--help`.

**Review notes:**

- The `--platform` option must be passed to `oras` when pulling multi-architecture images so that only the required
  architecture is downloaded.
- The manifest is fetched after pulling the tag to automatically detect the image file name. Since different base images
  use different file extensions (e.g., `image.vhdx` for minimal-os, `image.vhd` for linuxguard), the script determines
  the correct file name by process of elimination.
- Added `jq` to the list of installed packages in the container to ensure correct parsing of the manifest.
- Upgraded `oras` to version 1.2.2 to enable progress bars during interactive TTY docker runs, which enhances user
  experience with large image files.

### DIFF
--- a/toolkit/tools/imagecustomizer/container/run.sh
+++ b/toolkit/tools/imagecustomizer/container/run.sh
@@ -73,9 +73,9 @@ BASE_IMAGE_NAME="${BASE_IMAGE_NAME:-minimal-os}"
 VERSION_MAJOR_MINOR="$(echo "$ARG_VERSION" | cut -d'.' -f1-2)"
 OCI_ARTIFACT_REPOSITORY="azurelinux/$VERSION_MAJOR_MINOR/image/$BASE_IMAGE_NAME"
 
-VERSION_FINAL_PARTS="$(echo "$ARG_VERSION" | cut -d'.' -f3-)"
-if [[ "$VERSION_FINAL_PARTS" == latest || "$VERSION_FINAL_PARTS" == latest-* ]]; then
-    OCI_ARTIFACT_TAG="$VERSION_FINAL_PARTS"
+VERSION_SUFFIX="$(echo "$ARG_VERSION" | cut -d'.' -f3-)"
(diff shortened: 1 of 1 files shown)


toolkit/tools/imagecustomizer/container/run.sh
```diff
@@ -1,45 +1,116 @@
...
+if [[ -n "${DEVELOPER_MODE:-}" ]]; then
+    OCI_ARTIFACT_REGISTRY="acrafoimages.azurecr.io"
+fi
 
-# Pull the OCI artifact, and check if the pull was successful.
-oras pull $OCI_ARTIFACT_PATH -o $ARTIFACT_DIR
-if [ $? -ne 0 ]; then
-    echo "Error: Failed to fetch image with the provided version $version"
-    exit 1
+BASE_IMAGE_NAME="${BASE_IMAGE_NAME:-seed}"
```
gmileka, line 76: should the default be `minimal-os` and not `seed`?

gmileka (approved):

PR #298, opened 2025-07-03
add support for PackageSnapshotTime feature and validation checks for image creator config

himaja-kesari:
this PR adds support for the packagesnapshottime feature in the image creator.
adds validation checks for image creator configuration
adds more validation tests to validate the image creator configs
---

### **Checklist**
- [x] Tests added/updated


toolkit/tools/pkg/imagecustomizerlib/imageConnection.go
```diff
@@ -82,3 +93,43 @@ func (c *ImageConnection) CleanClose() error {
 
 	return nil
 }
+
+func ConnectToImage(buildDir string, imageFilePath string, includeDefaultMounts bool, mounts []MountPoint,
```
cwize1, line 97: This code shouldn't ever really be used in production code, since all the mount points are hardcoded. Would it be possible to move this into something like `testutils`?

toolkit/tools/imagecreator/main.go
```diff
@@ -28,6 +28,7 @@ type ImageCreatorCmd struct {
 	OutputImageFile   string   `name:"output-image-file" help:"Path to write the customized image to."`
 	OutputImageFormat string   `name:"output-image-format" placeholder:"(vhd|vhd-fixed|vhdx|qcow2|raw)" help:"Format of output image." enum:"${imageformat}" default:""`
 	exekong.LogFlags
+	PackageSnapshotTime string `name:"package-snapshot-time" help:"Only packages published before this snapshot time will be available during customization. Supports 'YYYY-MM-DD' or full RFC3339 timestamp (e.g., 2024-05-20T23:59:59Z)."`
```
cwize1, line 31: Adding a new feature should not be included in a PR titled "add validation tests".

toolkit/tools/pkg/imagecreatorlib/validation.go
```diff
@@ -0,0 +1,127 @@
...
+		return fmt.Errorf("input field is not supported by the image creator")
+	}
+	if c.Iso != nil {
+		return fmt.Errorf("iso field is not supported by the image creator")
+	}
+	if c.Pxe != nil {
+		return fmt.Errorf("pxe field is not supported by the image creator")
+	}
+
+	if len(c.PreviewFeatures) > 0 {
```
cwize1, line 25: I wouldn't block `PreviewFeatures`.

toolkit/tools/pkg/imagecreatorlib/validation.go
```diff
@@ -0,0 +1,127 @@
...
+	"strings"
+
+	"github.com/microsoft/azurelinux/toolkit/tools/imagecustomizerapi"
+	"github.com/microsoft/azurelinux/toolkit/tools/pkg/imagecustomizerlib"
+)
+
+func validateSupportedFields(c *imagecustomizerapi.Config) error {
+	// Verify that the config file does not contain any fields that are not supported by the image creator.
+	if c.Input != (imagecustomizerapi.Input{}) {
+		return fmt.Errorf("input field is not supported by the image creator")
```
cwize1, line 16: Either say `the image creator tool` or `image creator` not `the image creator`.

toolkit/tools/pkg/imagecreatorlib/validation.go
```diff
@@ -0,0 +1,127 @@
...
+	"context"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/microsoft/azurelinux/toolkit/tools/imagecustomizerapi"
+	"github.com/microsoft/azurelinux/toolkit/tools/pkg/imagecustomizerlib"
+)
+
+func validateSupportedFields(c *imagecustomizerapi.Config) error {
```
cwize1, line 13: Long term, it might not be a good idea to split the validation into `validateSupportedFields` and `validateMandatoryFields`. In the future, the validation logic may become more complex than a binary feature is or is not supported.

toolkit/tools/pkg/imagecreatorlib/validation.go
```diff
@@ -0,0 +1,127 @@
...
+
+	if c.Storage.ResetPartitionsUuidsType != imagecustomizerapi.ResetPartitionsUuidsTypeDefault {
+		return fmt.Errorf("reset partitions uuids field is not supported by the image creator")
+	}
+
+	if c.Storage.Verity != nil {
+		return fmt.Errorf("storage verity field is not supported by the image creator")
+	}
+
+	if c.OS != nil && len(c.OS.AdditionalFiles) > 0 {
```
cwize1, line 37: It might be good to use a helper function for all the OS fields to avoid duplicating the `c.OS != nil` check.

toolkit/tools/pkg/imagecreatorlib/validation.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	// Check if the tools tar file is a valid tar file
+	if !isValidTarFile(toolsTar) {
+		return fmt.Errorf("tools tar file %s is not a valid tar file", toolsTar)
+	}
+	return nil
+}
+
+func isValidTarFile(toolsTar string) bool {
+	return strings.HasSuffix(toolsTar, ".tar.gz")
```
cwize1, line 126: We don't use the filename to determine the file type. We look at its contents, typically looking for a magic number at the start of the file.

cwize1 (changes requested):

toolkit/tools/imagecreator/main.go
himaja-kesari: updated the title

toolkit/tools/pkg/imagecreatorlib/validation.go
himaja-kesari: updated

toolkit/tools/pkg/imagecreatorlib/validation.go
himaja-kesari: updated to use "the image creator tool"

toolkit/tools/pkg/imagecreatorlib/validation.go
himaja-kesari: updated the code to check whether a file is a valid .tar.gz archive by verifying both the gzip and tar structure:

toolkit/tools/pkg/imagecreatorlib/validation.go
himaja-kesari: I agree, we need a better way to validate going forward

toolkit/tools/pkg/imagecustomizerlib/imageConnection.go
himaja-kesari: moved it into testutils and resolved the circular dependency

toolkit/tools/pkg/imageconnection/imageConnection.go
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
```
cwize1, line 1: This package should probably be in `internal` not `pkg`. `pkg` is for public APIs.

toolkit/tools/pkg/imagecreatorlib/validation.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+	return nil
+}
+
+func validateConfig(ctx context.Context, baseConfigPath string, config *imagecustomizerapi.Config, rpmsSources []string, toolsTar string,
+	outputImageFile, outputImageFormat string, packageSnapshotTime string,
+) error {
+	err := validateSupportedFields(config)
+	if err != nil {
+		return fmt.Errorf("invalid config file %s:\n%w", baseConfigPath, err)
```
cwize1, line 74: nit: `(%s)`. This makes it clearer when the string is empty.

cwize1 (changes requested):

toolkit/tools/pkg/imageconnection/imageConnection.go
himaja-kesari: moved it to internal,

PR #299, opened 2025-07-03
Enable Image Customizer Error Telemetry 

aditjha-msft:
Enable error telemetry for imagecustomizer as well as flushing all telemetry before shutdown. 

Needed to add event logic to telemetry hopper because error spans are structured and processed a bit differently.

For errors, we are keeping a generic error message in the span for now, to prevent any accidental collection of user data through returned error strings. In a follow-up effort, we'll have a structured Error Type that would be included in the span's error field.

---

### **Checklist**
- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/imagecustomizer/main.go
+++ b/toolkit/tools/imagecustomizer/main.go
@@ -86,13 +86,13 @@ func main() {
 
 	switch parseContext.Command() {
 	case "customize":
-		err := customizeImage(ctx, cli.Customize)
+		err = customizeImage(ctx, cli.Customize)
 		if err != nil {
 			logger.Log.Errorf("image customization failed:\n%v", err)
 		}
 
 	case "inject-files":
-		err := injectFiles(ctx, cli.InjectFiles)
+		err = injectFiles(ctx, cli.InjectFiles)
 		if err != nil {
 			logger.Log.Errorf("inject-files failed:\n%v", err)
 		}
@@ -102,6 +102,9 @@ func main() {
 	}
 
 	if err != nil {
+		if err := telemetry.ShutdownTelemetry(ctx); err != nil {
+			logger.Log.Warnf("Failed to shutdown telemetry: %v", err)
+		}
 		os.Exit(1)
 	}
 }
--- a/toolkit/tools/imagecustomizer/main.go
+++ b/toolkit/tools/imagecustomizer/main.go
@@ -5,8 +5,9 @@ package main
 
 import (
 	"context"
+	"fmt"
+	"log"
 	"maps"
-	"os"
 	"strings"
 
 	"github.com/alecthomas/kong"
@@ -68,6 +69,13 @@ func main() {
 
 	logger.InitBestEffort(ptrutils.PtrTo(cli.LogFlags.AsLoggerFlags()))
 
+	err := runCommand(ctx, parseContext.Command(), cli)
+	if err != nil {
+		log.Fatalf("%v", err)
+	}
+}
+
+func runCommand(ctx context.Context, command string, cli *RootCmd) error {
 	// initialize OpenTelemetry tracer
 	err := telemetry.InitTelemetry(cli.DisableTelemetry, imagecustomizerlib.ToolVersion)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -237,14 +238,23 @@ func cleanUp(ic *ImageCustomizerParameters) error {
...
-) error {
+) (err error) {
 	ctx, span := otel.GetTracerProvider().Tracer(OtelTracerName).Start(ctx, "customize_image")
 	span.SetAttributes(
 		attribute.String("output_image_format", string(outputImageFormat)),
 	)
-	defer span.End()
+	defer func() {
+		if err != nil {
+			span.RecordError(err)
```
cwize1, line 248: The error message could theoretically contain user data. For example, if a postCustomization script failed. So, I am not sure it is wise to record the error message in telemetry.

cwize1, line 248: Why are you passing the error message to the telemetry in 3 different ways?

toolkit/tools/imagecustomizer/main.go
```diff
@@ -86,20 +86,27 @@ func main() {
...
 		if err != nil {
-			log.Fatalf("inject-files failed:\n%v", err)
+			logger.Log.Errorf("inject-files failed:\n%v", err)
 		}
 
 	default:
 		panic(parseContext.Command())
 	}
+
+	if err != nil {
```
cwize1, line 104: This code is a little confusing. I would create a helper function so that you can ensure the `telemetry.ShutdownTelemetry` defer is run before `os.Exit` is called. That way, you can avoid the duplicate call to `telemetry.ShutdownTelemetry`.

cwize1 (changes requested):

toolkit/tools/imagecustomizer/main.go
```diff
@@ -88,18 +88,22 @@
 	case "customize":
 		err := customizeImage(ctx, cli.Customize)
 		if err != nil {
-			log.Fatalf("image customization failed:\n%v", err)
+			logger.Log.Errorf("image customization failed:\n%v", err)
```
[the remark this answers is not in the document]
aditjha-msft: Discussed with @cwize1 that the fields in this alert are just metadata and not actually sensitive data. Ignoring.

toolkit/tools/imagecustomizer/main.go
```diff
@@ -86,20 +86,27 @@
 
 	switch parseContext.Command() {
 	case "customize":
-		err := customizeImage(ctx, cli.Customize)
+		err = customizeImage(ctx, cli.Customize)
 		if err != nil {
-			log.Fatalf("image customization failed:\n%v", err)
+			logger.Log.Errorf("image customization failed:\n%v", err)
```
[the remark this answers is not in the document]
aditjha-msft: Discussed with @cwize1 that the fields in this alert are just metadata and not actually sensitive data. Ignoring.

toolkit/tools/imagecustomizer/main.go
cwize1: Like so:

```golang
	err := runCommand(ctx, parseContext.Command(), cli)
	if err != nil {
		log.Fatalf("%v", err)
	}
}

func runCommand(ctx context.Context, command string, cli *RootCmd) error {
	// initialize OpenTelemetry tracer
	err := telemetry.InitTelemetry(cli.DisableTelemetry, imagecustomizerlib.ToolVersion)
	if err != nil {
		logger.Log.Warnf("Failed to initialize telemetry setup: %v", err)
	}
	defer func() {
		if err := telemetry.ShutdownTelemetry(ctx); err != nil {
			logger.Log.Warnf("Failed to shutdown telemetry: %v", err)
		}
	}()

	if cli.TimeStampFile != "" {
		timestamp.BeginTiming("imagecustomizer", cli.TimeStampFile)
		defer timestamp.CompleteTiming()
	}

	switch command {
	case "customize":
		err = customizeImage(ctx, cli.Customize)
		if err != nil {
			return fmt.Errorf("image customization failed:\n%w", err)
		}

	case "inject-files":
		err = injectFiles(ctx, cli.InjectFiles)
		if err != nil {
			return fmt.Errorf("inject-files failed:\n%w", err)
		}

	default:
		panic(command)
	}

	return nil
}
```

aditjha-msft: Ah got it, that makes sense. Updated with this approach.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
aditjha-msft: Not passing error message anymore, but those two functions do require a string so just keeping it generic for now. These would be replaced with the error type once we have structured errors.

PR #312, opened 2025-07-11
Image Customizer: fixing a crash in pxe build flow

gmileka:
The PXE flow got broken with the recent changes to accommodate kdump files. The problem comes from a typo where we are referencing the iso configuration in the pxe context where it may be nil and hence causes a crash.

- [-] Tests added/updated


cwize1 (approved):
Is there a test for this scenario?

PR #328, opened 2025-07-17
Use OS images cache for functional tests workflow.

cwize1:
Instead of downloading the images from MCR (Microsoft Container Registry), cache the OS images used by the functional tests in an Azure Storage Account. This should improve the reliability of the tests since they will no longer be reliant on resources managed by other teams.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/.github/workflows/scripts/download-image.sh
+++ b/.github/workflows/scripts/download-image.sh
@@ -1,14 +1,17 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-# Downloads an OS image from an Azure Stroage Account Blob Storage.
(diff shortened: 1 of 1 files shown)


.github/workflows/scripts/download-image.sh
```diff
@@ -0,0 +1,47 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+# Downloads an OS image from an Azure Stroage Account Blob Storage.
```
himaja-kesari, line 4: *storage

himaja-kesari (approved):

PR #331, opened 2025-07-18
VMTests: Fix when run as 'root' user.

cwize1:
The VMTests suite uses the name of the current user as the name of the test user it deploys to the OS image. This makes it easier for the user to manually SSH into the VM when debugging. However, this doesn't work when the test is being run as "root" (e.g. sudo) since the SSH login is typically disabled for the "root" user. This change adds code to detect this and use a different username.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/test/vmtests/vmtests/test_min_change.py
+++ b/test/vmtests/vmtests/test_min_change.py
@@ -4,7 +4,6 @@
 import logging
 import os
 import platform
-from getpass import getuser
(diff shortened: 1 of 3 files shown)


aditjha-msft (approved):
lgtm, branch needs update

edit: never mind, didn't realize you can merge anyways

PR #338, opened 2025-07-22
Add OSModifier tests

elainezhao1:
Added vm based tests for testing osmodifier api: 
- enable/disable systemd services
- User configuration
- kernel module
- GRUB bootloader configuration (kernel command line, overlays, verity, SELinux)
- UKI/systemd-boot SELinux config

tested with minimal-os.vhdx, 3.0 core efi image, uki-enabled image
image

---

### **Checklist**
- [ ] Tests added/updated


test/osmodifiertest/pyproject.toml
```diff
@@ -0,0 +1,23 @@
...
+[tool.isort]
+line_length = 120
+profile = "black"
+skip = ["build", "venv", ".venv", "out", "__pycache__", ".pytest_cache", ".mypy_cache"]
+extend_skip = ["build", "venv"]
+skip_glob = ["build/*", "venv/*", ".venv/*"]
+
+[tool.black]
+line-length = 120
+extend-exclude = '''
```
cwize1, line 13: In `vmtests`, there are two directories named `vmtests`. The second `vmtests` directory contains the actual Python code. Please follow that same pattern here to avoid needing to add all these excludes and skips,

test/osmodifiertest/test_osmodifier.py
```diff
@@ -0,0 +1,344 @@
...
+import tempfile
+from getpass import getuser
+from pathlib import Path
+from typing import Generator, List, Tuple
+
+import libvirt  # type: ignore
+import pytest
+import yaml
+from docker import DockerClient
+from vmtests.vmtests.utils import local_client
```
cwize1, line 16: We should probably move the `utils` to a common location if they are going to be used by two different test suites.

test/osmodifiertest/conftest.py
```diff
@@ -0,0 +1,98 @@
...
+
+import random
+import shutil
+import string
+from pathlib import Path
+from typing import Generator, List
+
+import pytest
+from vmtests.vmtests.conftest import (
+    close_list,
```
cwize1, line 12: These should be moved into a common location.

test/osmodifiertest/conftest.py
```diff
@@ -0,0 +1,98 @@
...
+@pytest.fixture(scope="session")
+def input_image(request: pytest.FixtureRequest) -> Generator[Path, None, None]:
+    image = request.config.getoption("--input-image")
+    if not image:
+        raise Exception("--input-image is required for test")
+    yield Path(image)
+
+
+@pytest.fixture(scope="session")
+def test_instance_name(request: pytest.FixtureRequest) -> Generator[str, None, None]:
```
cwize1, line 76: This isn't a name of a test instance. It is a name for a session.

test/osmodifiertest/conftest.py
```diff
@@ -0,0 +1,98 @@
...
+    instance_suffix = "".join(random.choice(string.ascii_uppercase) for _ in range(5))
+    yield request.node.name + "-" + instance_suffix
+
+
+@pytest.fixture(scope="session")
+def close_list(keep_environment: bool) -> Generator[List[Closeable], None, None]:
```
cwize1, line 82: You should probably give this session-scoped `close_list` a different name from the function-scoped `close_list`.

cwize1 (changes requested):

test/osmodifiertest/test_osmodifier.py
elainezhao1: moved utils dir to test/utils. Now there are individual Makefiles in each test directory (vmtests/, osmodifier/)
Each handles formatting/linting for their own code
But shared utilities in test/utils/ don't get formatted when running make format from subdirectories
I wonder if I should add a root test/Makefile that coordinates formatting across all directories

elainezhao1: another question is - I've added `sys.path.insert(0, str(Path(__file__).parent.parent))` in several files to make shared utilities importable because conftest.py needs to import from test/utils/
otherwise I'm getting error `Relative imports (from ..utils import X) fail with "attempted relative import beyond top-level package"`
let me know if this is ok

cwize1: Perhaps it might be easier to have all the tests in a single parent directory? That way, they can easily share the utils, Makefile, etc.

Something along the lines of:

```
/test/vmtests/Makefile
/test/vmtests/vmtests/conftest.py
/test/vmtests/vmtests/imagecustomizer/conftest.py
/test/vmtests/vmtests/imagecustomizer/test_min_change.py
/test/vmtests/vmtests/osmodifier/conftest.py
/test/vmtests/vmtests/osmodifier/test_osmodifier.py
/test/vmtests/vmtests/utils
etc.
```

cwize1:
FYI: Himaja is also adding to VMTests in #352

test/vmtests/vmtests/conftest.py
```diff
@@ -22,39 +22,28 @@
 TEST_CONFIGS_DIR = SCRIPT_PATH.joinpath("../../../toolkit/tools/pkg/imagecustomizerlib/testdata")
 
 
-def pytest_addoption(parser: pytest.Parser) -> None:
```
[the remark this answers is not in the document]
elainezhao1: good catch! I've kept the common flags here

test/vmtests/vmtests/conftest.py
```diff
@@ -42,19 +38,20 @@ def keep_environment(request: pytest.FixtureRequest) -> Generator[bool, None, No
...
-def test_instance_name(request: pytest.FixtureRequest) -> Generator[str, None, None]:
+
+def _generate_test_name(base_name: str) -> str:
+    """Helper function to generate unique test instance names."""
     instance_suffix = "".join(random.choice(string.ascii_uppercase) for _ in range(5))
-    yield request.node.name + "-" + instance_suffix
+    return f"{base_name}-{instance_suffix}"
 
 
 # pytest has an in-built fixture called tmp_path. But that uses /tmp, which sits in memory.
```
cwize1, line 57: `test_temp_dir` relies on `session_temp_dir`. But `session_temp_dir` isn't defined at this level. This is a little confusing. Any reason `session_temp_dir` can't be defined in this `conftest.py`?

cwize1 (changes requested):

cwize1 (approved):

test/vmtests/vmtests/conftest.py
elainezhao1: fixed it.  Previously I passed different temp folder names. now they are all under vm tests, so it makes sense to have vmtests- prefix

PR #340, opened 2025-07-23
Add tar.gz to gitignore

aditjha-msft:
---

Noticed the file `toolkit/tools/pkg/imagecreatorlib/testdata/tools.tar.gz` being created during local builds, adding to gitignore.

### **Checklist**
- [x] Tests added/updated

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ build/
 out/
 
 *.tmp
+*.tar.gz
 
 # Python
 __pycache__/


cwize1:
It would probably be better to ensure the `tools.tar.gz` is placed in a more appropriate directory, than ignoring all tar.gz files.

aditjha-msft:
guess I don't really have the context here, but I see it shows up in the `imagecreatorlib/testdata` folder. So do we want it checked in then? I see there is already a `testchroot.tar.gz` checked in under `toolkit/tools/internal/safechroot/testdata`. Should we be excluding these `testdata` folders instead maybe? @cwize1 @himaja-kesari

himaja-kesari:
> > guess I don't really have the context here, but I see it shows up in the `imagecreatorlib/testdata` folder. So do we want it checked in then? I see there is already a `testchroot.tar.gz` checked in under `toolkit/tools/internal/safechroot/testdata`. Should we be excluding these `testdata` folders instead maybe? @cwize1 @himaja-kesari
> 
> I think this might be coming from tests: `TestValidateOutput_AcceptsValidPaths` and `TestValidateConfig_EmptyPackagestoInstall`. Those tests are putting the temporary file `tools.tar.gz` in the testdata directory (which contains test data checked into the repo) instead of a temporary test directory.

I will make some changes to the functional tests so that it's created in a tmp directory and deleted after the test runs.

aditjha-msft:
@himaja-kesari will make changes as part of separate PR, closing this one.

PR #349, opened 2025-07-30
Add API for mounting partition as readonly during customization.

cwize1:
Add an API that allows the user to specify that a partition should be set as readonly during OS customization.

The primary scenario being targeted is recustomizing an image with a usr verity partition + UKIs without changing/invalidating the /usr partition or the UKIs. A functional test has been added for this scenario.

---

### **Checklist**

- [x] Tests added/updated


cwize1:
Will use a different API.

PR #358, opened 2025-08-05
Fix unintended password reset for existing users

elainezhao1:
updates the addOrUpdateUser logic to avoid unintentionally clearing a pre-existing user's password when no new password is explicitly provided in the config.

---

### **Checklist**
- [ ] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/customizeusers.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/customizeusers.go
@@ -120,7 +120,6 @@ func addOrUpdateUser(user imagecustomizerapi.User, baseConfigPath string, imageC
 		}
 	}
 
-
 	// Set user's password expiry.
 	if user.PasswordExpiresDays != nil {
 		err = installutils.Chage(imageChroot, *user.PasswordExpiresDays, user.Name)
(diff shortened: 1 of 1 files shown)


toolkit/tools/pkg/imagecustomizerlib/customizeusers.go
```diff
@@ -72,12 +75,16 @@ func addOrUpdateUser(user imagecustomizerapi.User, baseConfigPath string, imageC
...
 		}
 
-		hashedPassword = password
-		if !passwordIsHashed {
-			// Hash the password.
-			hashedPassword, err = userutils.HashPassword(password)
-			if err != nil {
-				return err
+		// Only proceed if password is not empty
+		if strings.TrimSpace(password) != "" {
```
cwize1, line 79: Passwords are allowed to have spaces at the start and at the end.

cwize1, line 79: If the user specifies the password field but provides an empty password value, then we should remove the password from the user. If the user doesn't provide the password field, then the user's password should be unchanged.

cwize1 (changes requested):

PR #368, opened 2025-08-13
Set container ENTRYPOINT.

cwize1:
Set the `imagecustomizer` executable as the default ENTRYPOINT for the container.

Technically, the telemetry hopper is already the entrypoint. So, this change is actually just having the telemetry hopper call `imagecustomizer`, instead of requiring the user to pass in the command name. But when the telemetry hopper is removed, `imagecustomizer` will become the entrypoint.

Also, update the `run.sh` script to include a copy of the telemetry hopper logic. Now, to use the `run.sh` script, the user must pass `--entrypoint /usr/local/bin/run.sh` to docker to override the default command.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/docs/imagecustomizer/how-to/azure-vm.md
+++ b/docs/imagecustomizer/how-to/azure-vm.md
@@ -152,14 +152,13 @@ in front of any HTTP endpoints.
 5. Run Image Customizer to create the new image:
 
    ```bash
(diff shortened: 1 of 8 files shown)


aditjha-msft (approved):

liulanze:
LGTM, based on your change, I will update golden container testing logic.

liulanze (approved):

PR #369, opened 2025-08-13
Imagecreator: fedora base image support

himaja-kesari:
adding support for creating fedora base images using imagecreator 
---

### **Checklist**
- [x] Tests added/updated


toolkit/tools/pkg/imagecustomizerlib/distrohandler.go
```diff
@@ -0,0 +1,62 @@
...
+
+const (
+	distroNameAzureLinux DistroName = "azurelinux"
+	distroNameFedora     DistroName = "fedora"
+)
+
+// distroHandler represents the interface for distribution-specific configuration
+type distroHandler interface {
+	// Package manager configuration
+	getPackageManagerBinary() string
```
cwize1, line 32: Everything except for `managePackages` is an implementation detail of package installation with tdnf/dnf. If you want an interface to handle tdnf/dnf differences, then create a new interface. Don't pollute the distro one.

toolkit/tools/pkg/imagecustomizerlib/customizepackages_rpm.go
```diff
@@ -0,0 +1,414 @@
...
+const (
+	dnfDownloadPattern = `^\s*([a-zA-Z0-9\-._+]+(?:\.[a-zA-Z0-9_]+)*\.rpm)\s+.*\d+.*[kMG]B/s.*\|\s*\d+`
+)
+
+// azureLinuxDistroConfig implements distroHandler for Azure Linux
+type azureLinuxDistroConfig struct {
+	version string
+}
+
+func (d *azureLinuxDistroConfig) getPackageManagerBinary() string { return "tdnf" }
```
cwize1, line 55: Place `azureLinuxDistroConfig` and `fedoraDistroConfig` in their own files.

Actually, the difference here is not actually `azureLinux` and `fedora` but `tdnf` and `dnf`. So, that is how the files should be split. (See my other comment about not polluting `distroHandler` with stuff specific to dnf/tdnf.)

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
```diff
@@ -37,167 +29,94 @@ var (
...
-	ctx, span := otel.GetTracerProvider().Tracer(OtelTracerName).Start(ctx, "configure_packages")
-	defer span.End()
-
-	if snapshotTime == "" {
-		snapshotTime = string(config.Packages.SnapshotTime)
-	}
-
-	tdnfChroot := imageChroot
+// executePackageManagerCommand runs a package manager command with proper chroot handling
+func executePackageManagerCommand(args []string, imageChroot *safechroot.Chroot, toolsChroot *safechroot.Chroot, distroConfig distroHandler) error {
```
cwize1, line 35: `executePackageManagerCommand` is specific to tdnf/dnf. So, it shouldn't live in the `customizepackages.go` file. Also, you should give it a more specific name. For example, `executeDnf`.

Similar comment for other functions in this file.

toolkit/tools/internal/testutils/testutils.go
```diff
@@ -104,12 +104,20 @@ func GetDownloadedRpmsDir(t *testing.T, testutilsDir string, azureLinuxVersion s
 	return downloadedRpmsDir
 }
 
-func GetDownloadedToolsFile(t *testing.T, testutilsDir string, azureLinuxVersion string, imagecreator bool) string {
-	GetDownloadedToolsFile := filepath.Join(testutilsDir, "testrpms/tools.tar.gz")
+func GetDownloadedToolsFile(t *testing.T, testutilsDir string, linuxVersion string, imagecreator bool) string {
+	var toolsFileName string
+	// Use different tools file for Fedora
+	if linuxVersion == "42" {
```
cwize1, line 110: The code needs to be more generic than this. We don't want to maintain `if` checks like this for every distro+version combo we support.

toolkit/tools/internal/testutils/testrpms/download-test-rpms.sh
```diff
@@ -22,11 +22,58 @@ REPO_NO_KEY_FILE="$DOWNLOADER_RPMS_DIRS/rpms-$IMAGE_VERSION-nokey.repo"
...
+DOCKERFILE_NAME="Dockerfile"
+PACKAGE_LIST=""
+
+if [[ "$CONTAINER_IMAGE" == *"fedora"* ]]; then
+  DOCKERFILE_NAME="Dockerfile.fedora"
+  echo "Using Fedora-specific Dockerfile"
+  # Extract packages from fedora.yaml
+  if command -v python3 >/dev/null 2>&1; then
+    PACKAGE_LIST=$(python3 -c "
+import yaml
```
cwize1, line 36: This Python code is almost identical, except for the name of the file. Move this code to a generic Python script and use it for both distros.

toolkit/tools/internal/testutils/testrpms/download-test-rpms.sh
```diff
@@ -22,11 +22,58 @@ REPO_NO_KEY_FILE="$DOWNLOADER_RPMS_DIRS/rpms-$IMAGE_VERSION-nokey.repo"
 
 mkdir -p "$OUT_DIR"
 
+# Determine which Dockerfile to use and extract package list based on the base image
+TESTDATA_DIR="$SCRIPT_DIR/../../../../tools/pkg/imagecreatorlib/testdata"
+DOCKERFILE_NAME="Dockerfile"
+PACKAGE_LIST=""
+
+if [[ "$CONTAINER_IMAGE" == *"fedora"* ]]; then
```
cwize1, line 30: We need to avoid distro if-checks as much as possible. Otherwise, adding more distros+versions is going to be painful. Try to make things as declarative as possible.

For example, have a:

- Map from distro to minimal-os config
- Map from distro to downloader Dockerfile

It might be worth considering changing these scripts to use Python instead of Bash.

toolkit/tools/internal/testutils/testrpms/download-test-utils.sh
```diff
@@ -22,23 +22,39 @@ do
...
+# Determine the tools file name based on the image version
+BUILD_DIR="$SCRIPT_DIR/build"
+mkdir -p "$BUILD_DIR"
+
+case "${IMAGE_VERSION}" in
+  42)
+    TOOLS_FILE="$BUILD_DIR/tools-fedora.tar.gz"
+    ;;
+  *)
+    TOOLS_FILE="$BUILD_DIR/tools.tar.gz"
```
cwize1, line 41: Be consistent with naming. Add distro name (and probably version as well) to all versions of the tools tarball.

toolkit/tools/internal/testutils/testrpms/download-test-utils.sh
```diff
@@ -22,23 +22,39 @@ do
...
             echo "  -h Show help"
             exit 1;;
     esac
 done
 
+# Determine the tools file name based on the image version
+BUILD_DIR="$SCRIPT_DIR/build"
+mkdir -p "$BUILD_DIR"
+
+case "${IMAGE_VERSION}" in
```
cwize1, line 36: Doing a switch on just the distro version is not ok. It should be either on distro or distro+version.

toolkit/tools/pkg/imagecreatorlib/imagecreator.go
```diff
@@ -208,3 +218,22 @@ func createImageCreatorParameters(buildDir string,
 
 	return ic, nil
 }
+
+// getTargetOs returns the appropriate TargetOs based on distro and version
+func getTargetOs(distro string, distroVersion string) (targetos.TargetOs, error) {
```
cwize1, line 223: Would it make sense to use a nested maps instead of nested switch statements? That way, it will be more declarative.

toolkit/tools/internal/testutils/testrpms/downloader/Dockerfile.azurelinux
```diff
@@ -0,0 +1,18 @@
+ARG baseimage
+FROM ${baseimage}
+ARG imagecreator
```
cwize1, line 3: Remove the `imagecreator` ARG and handle `jq golang` in `download-test-rpms.sh`. That way, we simplify this file.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -517,7 +518,7 @@ func customizeOSContents(ctx context.Context, ic *ImageCustomizerParameters) err
 	// Customize the raw image file.
 	partUuidToFstabEntry, baseImageVerityMetadata, readonlyPartUuids, osRelease, err := customizeImageHelper(ctx,
 		ic.buildDirAbs, ic.configPath, ic.config, ic.rawImageFile, ic.rpmsSources, ic.useBaseImageRpmRepos,
-		partitionsCustomized, ic.imageUuidStr, ic.packageSnapshotTime, ic.outputImageFormat)
+		partitionsCustomized, ic.imageUuidStr, ic.packageSnapshotTime, ic.outputImageFormat, "")
```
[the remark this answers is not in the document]
himaja-kesari: updated passing the interface,

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
```diff
@@ -322,65 +279,133 @@ func callTdnf(tdnfArgs []string, imageChroot *safechroot.Chroot, toolsChroot *sa
...
+
+		allPackages = append(allPackages, packageList.Packages...)
 	}
 
-	return tdnfArgs
+	allPackages = append(allPackages, packages...)
+	return allPackages, nil
+}
+
+// TODO: update this function when adding support for fedora for image customizer
```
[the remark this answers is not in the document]
himaja-kesari: will be updated as a follow up PR, this function is being used in the imagecustomizer workflow.

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
```diff
@@ -23,295 +23,252 @@ import (
...
+	getBinaryName() string
+
+	// getDownloadRegex returns the regex pattern for parsing download progress lines
+	getDownloadRegex() *regexp.Regexp
+
+	// getTransactionErrorRegex returns the regex pattern for detecting transaction errors
+	getTransactionErrorRegex() *regexp.Regexp
+
+	// getSummaryLines returns the lines that indicate the start of operation summaries
+	getSummaryLines() []string
```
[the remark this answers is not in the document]
himaja-kesari: with the new abstraction layer I am having separate output parsers for tdnf and dnf.

toolkit/tools/pkg/imagecustomizerlib/packagemanager_tdnf.go
```diff
@@ -0,0 +1,95 @@
...
+		version = "3.0" // default version for Azure Linux
+	}
+	return &tdnfPackageManager{version: version}
+}
+
+func (pm *tdnfPackageManager) getPackageManagerBinary() string { return "tdnf" }
+func (pm *tdnfPackageManager) getPackageType() PackageType     { return packageTypeRPM }
+func (pm *tdnfPackageManager) getReleaseVersion() string       { return pm.version }
+func (pm *tdnfPackageManager) getConfigFile() string           { return customTdnfConfRelPath }
+func (pm *tdnfPackageManager) getPackageSourceDir() string     { return rpmsMountParentDirInChroot }
```
cwize1, line 30: Does `getPackageSourceDir` need to be an abstraction point at the moment? It seems like it is always the same value.

toolkit/tools/pkg/imagecustomizerlib/customizepackages_rpm.go
```diff
@@ -0,0 +1,169 @@
...
+	"context"
+
+	"github.com/microsoft/azurelinux/toolkit/tools/imagecustomizerapi"
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/safechroot"
+)
+
+// Distribution Configurations
+
+// azureLinuxDistroConfig implements distroHandler for Azure Linux
+type azureLinuxDistroConfig struct {
```
cwize1, line 13: Each class (i.e. a struct with methods) should have its own file. Structs that only store data are not required to live in their own file.

toolkit/tools/pkg/imagecustomizerlib/distrohandler.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// NewDistroHandler creates the appropriate distro handler with version support
+func NewDistroHandler(distroName string, version string) distroHandler {
+	switch distroName {
+	case string(distroNameFedora):
+		return newFedoraDistroConfig(version, packageManagerDNF)
+	case string(distroNameAzureLinux):
+		fallthrough
+	default:
```
cwize1, line 79: Don't have defaults like this. Instead, `panic` if an unexpected value is seen.

toolkit/tools/pkg/imagecustomizerlib/distrohandler.go
```diff
@@ -0,0 +1,94 @@
...
+	getPackageManager() packageManagerHandler
+
+	// Package management operations
+	managePackages(ctx context.Context, buildDir string, baseConfigPath string, config *imagecustomizerapi.OS,
+		imageChroot *safechroot.Chroot, toolsChroot *safechroot.Chroot,
+		rpmsSources []string, useBaseImageRpmRepos bool, snapshotTime string) error
+}
+
+// NewDistroHandler creates the appropriate distro handler with version support
+func NewDistroHandler(distroName string, version string) distroHandler {
```
cwize1, line 73: Please key everything off `targetos.TargetOs`. Don't go passing around raw distro names and distro version strings.

By the way, feel free to extend `TargetOs` as needed. For example, you could add a `GetDistro` and `GetDistroVersion` functions to `TargetOs`.

toolkit/tools/imagegen/diskutils/filesystem.go
```diff
@@ -69,6 +71,23 @@ var (
...
+		},
+	}
+
+	// The default xfs options used by Fedora 42 (kernel v6.11+)
+	// Based on modern XFS features supported in recent kernels
+	fedora42XfsOptions = xfsOptions{
+		Features: []string{"bigtime", "crc", "finobt", "inobtcount", "reflink", "rmapbt", "sparse", "nrext64"},
+	}
+
 	targetOsFileSystemsOptions = map[targetos.TargetOs]fileSystemsOptions{
```
cwize1, line 91: It would probably make sense to replace `targetOsFileSystemsOptions` with a `getFileSystemsOptions` function in `distroHandler`.

toolkit/tools/pkg/imagecustomizerlib/packagemanager_rpm.go
```diff
@@ -0,0 +1,204 @@
...
+
+	err := executePackageManagerCommand(args, imageChroot, toolsChroot, pmHandler)
+	if err != nil {
+		return fmt.Errorf("%w:\n%w", ErrPackageCacheClean, err)
+	}
+	return nil
+}
+
+// prefillDnfCache downloads packages to DNF cache for offline installation
+func prefillDnfCache(ctx context.Context, installPackages []string, updatePackages []string, imageChroot *safechroot.Chroot, toolsChroot *safechroot.Chroot, pmHandler rpmPackageManagerHandler) error {
```
cwize1, line 175: The behavior for tdnf is to download the RPM repo metadata once and cache it. Whereas, this is downloading all packages, even if they might not be required (e.g. that version of the package is already installed).

toolkit/tools/pkg/imagecustomizerlib/packagemanager_rpm.go
```diff
@@ -0,0 +1,204 @@
...
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/logger"
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/safechroot"
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/shell"
+	"github.com/sirupsen/logrus"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+)
+
+// executePackageManagerCommand runs a package manager command with proper chroot handling
+func executePackageManagerCommand(args []string, imageChroot *safechroot.Chroot, toolsChroot *safechroot.Chroot, pmHandler rpmPackageManagerHandler) error {
```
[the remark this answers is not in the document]
himaja-kesari: this executes both tdnf and dnf do you mean executeRpmPackagemanager ?

toolkit/tools/pkg/imagecustomizerlib/packagemanager_rpm.go
himaja-kesari: as per my understanding DNF will only install packages that are already downloaded as .rpm files in the cache. It will not attempt to fetch any RPMs from the network—even if the metadata (like repomd.xml and primary.xml.zck) clearly indicates which RPMs are needed.

himaja-kesari: this should work for image creator as there are no packages installed as we are building the images from scratch

toolkit/tools/pkg/imagecustomizerlib/customizepackages_rpm.go
himaja-kesari: added abstraction at distro and package level, resolving the comments as I have updated the codebase.

toolkit/tools/imagegen/diskutils/filesystem.go
himaja-kesari: I will update this flow as part of adding fedora support to image customizer, otherwise I have to update the image customizer code flow as well or duplicate the code.

toolkit/tools/pkg/imagecustomizerlib/packagemanager_tdnf.go
himaja-kesari: yeah! makes sense to remove this function

toolkit/tools/pkg/imagecustomizerlib/packagemanager_tdnf.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+// TDNF Package Manager Implementation
+type tdnfPackageManager struct {
+	version string
+}
+
+func newTdnfPackageManager(version string) *tdnfPackageManager {
+	if version == "" {
+		version = "3.0" // default version for Azure Linux
```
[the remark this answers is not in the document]
himaja-kesari: I agree, but just though since at the present state we only support azurelinux 3.0 and fedora 42 for image creator wouldn't it be convenient to have a default value instead of having customer to always provide the values.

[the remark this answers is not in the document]
himaja-kesari: yeah! makes sense updated.

toolkit/tools/pkg/imagecustomizerlib/distrohandler.go
```diff
@@ -0,0 +1,94 @@
...
+// DistroName represents the distribution name
+type DistroName string
+
+const (
+	distroNameAzureLinux DistroName = "azurelinux"
+	distroNameFedora     DistroName = "fedora"
+)
+
+// packageManagerHandler represents the interface for package manager implementation details
+type packageManagerHandler interface {
```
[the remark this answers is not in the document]
himaja-kesari: removed the generic packageManagerHandler  and the code now has only rpmPackageManagerHandler

toolkit/tools/internal/testutils/testutils.go
himaja-kesari: removed the changes to all the files in testutils, will incorporate all the feedback for files in testutils as part of a separate PR. thanks for the suggestion.

toolkit/tools/pkg/imagecustomizerlib/customizepackages_rpm.go
```diff
@@ -0,0 +1,98 @@
...
+	packageManagerChroot := imageChroot
+	if toolsChroot != nil {
+		packageManagerChroot = toolsChroot
+	}
+
+	// Get package manager from distribution config
+	pmHandler := distroConfig.getPackageManager()
+
+	// Setup distribution-specific configuration if needed
+	if distroConfig.getDistroName() == distroNameAzureLinux {
```
cwize1, line 26: You should handle snapshot time in a similar way to `getCacheOnlyOptions()`.

Then `getDistroName` can be removed from `distroHandler`.

toolkit/tools/pkg/imagecustomizerlib/customizepackages_rpm.go
```diff
@@ -0,0 +1,98 @@
...
+		// Setup Azure Linux specific TDNF configuration with snapshot
+		err = createTempTdnfConfigWithSnapshot(packageManagerChroot, imagecustomizerapi.PackageSnapshotTime(snapshotTime))
```
cwize1, line 28: An error should result if `snapshotTime` is provided but the distro doesn't support it.

toolkit/tools/pkg/imagecustomizerlib/distrohandler.go
```diff
@@ -0,0 +1,76 @@
...
+	GetTargetOs() targetos.TargetOs
+
+	// Package management operations
+	managePackages(ctx context.Context, buildDir string, baseConfigPath string, config *imagecustomizerapi.OS,
+		imageChroot *safechroot.Chroot, toolsChroot *safechroot.Chroot,
+		rpmsSources []string, useBaseImageRpmRepos bool, snapshotTime string) error
+}
+
+// NewDistroHandler creates the appropriate distro handler with version support
+func NewDistroHandler(distroName string, version string) distroHandler {
```
cwize1, line 55: All the internal code should use `TargetOs`. Distro name and version should only be handled at the API layer.

toolkit/tools/pkg/imagecustomizerlib/azurelinux_distro_config.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/targetos"
+)
+
+// azureLinuxDistroConfig implements distroHandler for Azure Linux
+type azureLinuxDistroConfig struct {
+	version        string
+	packageManager rpmPackageManagerHandler
+}
+
+func newAzureLinuxDistroConfig(version string, packageManagerType PackageManagerType) *azureLinuxDistroConfig {
```
cwize1, line 20: Why does `newAzureLinuxDistroConfig` take a `packageManagerType`? Doesn't Azure Linux always use `tdnf`?

toolkit/tools/pkg/imagecustomizerlib/distrohandler.go
```diff
@@ -0,0 +1,76 @@
...
+		return newFedoraDistroConfig(version, packageManagerDNF)
+	case string(distroNameAzureLinux):
+		return newAzureLinuxDistroConfig(version, packageManagerTDNF)
+	default:
+		panic("unsupported distro name: " + distroName)
+	}
+}
+
+// NewDistroHandlerWithPackageManager creates a distro handler with specific package manager
+func NewDistroHandlerWithPackageManager(distroName string, version string, pmType PackageManagerType) distroHandler {
```
cwize1, line 67: `NewDistroHandlerWithPackageManager` is not being used. Delete?

toolkit/tools/pkg/imagecustomizerlib/packagemanager_rpm.go
```diff
@@ -0,0 +1,179 @@
...
+
+	// Package manager specific output handling
+	createOutputCallback() func(string)
+
+	// Package manager specific cache options for install/update operations
+	getCacheOnlyOptions() []string
+}
+
+// executeRpmPackageManagerCommand runs a package manager command with proper chroot handling
+func executeRpmPackageManagerCommand(args []string, imageChroot *safechroot.Chroot, toolsChroot *safechroot.Chroot, pmHandler rpmPackageManagerHandler) error {
```
cwize1, line 35: The functions in `packagemanager_rpm.go` should actually live in `customizepackages_rpm.go`. `rpmPackageManagerHandler` should remain in `packagemanager_rpm.go`.

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
```diff
@@ -195,149 +63,7 @@ func collectPackagesList(baseConfigPath string, packageLists []string, packages
...
-	return tdnfChroot.UnsafeRun(func() error {
-		return shell.NewExecBuilder("tdnf", tdnfArgs...).
-			StdoutCallback(stdoutCallback).
-			LogLevel(shell.LogDisabledLevel, logrus.DebugLevel).
-			ErrorStderrLines(1).
-			Execute()
-	})
-}
-
+// TODO remove this after adding fedora support for image customizer
```
cwize1, line 66: Why would this need to be removed when Fedora support is added to Image Customizer?

toolkit/tools/internal/safechroot/safechroot.go
```diff
@@ -678,6 +678,11 @@ func defaultMountPoints() []*MountPoint {
 			fstype: "devpts",
 			data:   "gid=5,mode=620",
 		},
+		{
+			target: "/tmp",
```
cwize1, line 682: Adding `/tmp` here is a fairly non-trivial behavioral change. We might want to do this as a separate change. We might also want to do this before v1 of Image Customizer since it is a breaking change.

toolkit/tools/pkg/imagecustomizerlib/packagemanager_rpm.go
```diff
@@ -0,0 +1,179 @@
...
+	return nil
+}
+
+// removeRpmPackages removes packages using the appropriate package manager
+func removeRpmPackages(ctx context.Context, allPackagesToRemove []string, imageChroot *safechroot.Chroot, toolsChroot *safechroot.Chroot, pmHandler rpmPackageManagerHandler) error {
+	if len(allPackagesToRemove) <= 0 {
+		return nil
+	}
+
+	_, span := otel.GetTracerProvider().Tracer(OtelTracerName).Start(ctx, "remove_packages")
```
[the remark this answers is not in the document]
himaja-kesari: make sense will do this as a follow up change.

toolkit/tools/internal/targetos/targetos.go
```diff
@@ -15,6 +15,7 @@ type TargetOs string
 const (
 	TargetOsAzureLinux2 TargetOs = "azl2"
 	TargetOsAzureLinux3 TargetOs = "azl3"
+	TargetOsFedora42    TargetOs = "fedora42"
 )
 
 func GetInstalledTargetOs(rootfs string) (TargetOs, error) {
```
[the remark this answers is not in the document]
himaja-kesari: not being used for image creator will be updated as part of the imagecustomizer support PR

toolkit/tools/imagecreator/main.go
```diff
@@ -27,6 +27,8 @@ type ImageCreatorCmd struct {
 	ToolsTar          string   `name:"tools-file" help:"Path to tdnf worker tarball" required:""`
 	OutputImageFile   string   `name:"output-image-file" help:"Path to write the customized image to."`
 	OutputImageFormat string   `name:"output-image-format" placeholder:"(vhd|vhd-fixed|vhdx|qcow2|raw)" help:"Format of output image." enum:"${imageformat}" default:""`
+	Distro            string   `name:"distro" help:"Target distribution for the image." enum:"azurelinux,fedora" default:"azurelinux"`
```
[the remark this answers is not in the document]
himaja-kesari: for different distro versions  of the same distro we may be able to reuse the config files but not for different distros

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
himaja-kesari: this can be made more generic such that it can support both tdnf and dnf

toolkit/tools/imagecreator/main.go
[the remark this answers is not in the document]
himaja-kesari: I think we may be able to reuse the same config files for fedora and azl 4.0

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
cwize1: There are probably lots of things that will need updating as more distros are added. We probably don't need to add comments like this.

Also, this function will likely look very different with multi-distro support since both the package manager and the package names will be different between distros.

toolkit/tools/pkg/imagecustomizerlib/azurelinux_distro_config.go
cwize1: Should the `PackageManagerType` be selected based on `version`?

toolkit/tools/pkg/imagecustomizerlib/distrohandler_fedora.go
```diff
@@ -0,0 +1,54 @@
...
+		panic("unsupported package manager type for Fedora: " + string(packageManagerType))
+	}
+
+	return &fedoraDistroHandler{
+		version:        version,
+		packageManager: pm,
+	}
+}
+
+func (d *fedoraDistroHandler) getDistroName() DistroName { return distroNameFedora }
```
cwize1, line 35: Delete unused `getDistroName` function.

toolkit/tools/internal/safechroot/safechroot.go
himaja-kesari: [created a separate PR.](https://github.com/microsoft/azure-linux-image-tools/pull/391)

toolkit/tools/pkg/imagecustomizerlib/packagemanager_rpm.go
himaja-kesari: updated

toolkit/tools/pkg/imagecustomizerlib/customizepackages_rpm.go
```diff
@@ -0,0 +1,262 @@
...
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/logger"
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/safechroot"
+	"github.com/microsoft/azurelinux/toolkit/tools/internal/shell"
+	"github.com/sirupsen/logrus"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+)
+
+// managePackagesRpm provides a shared implementation for RPM-based package management
+func managePackagesRpm(ctx context.Context, buildDir string, baseConfigPath string, config *imagecustomizerapi.OS,
```
cwize1, line 19: Make the package manager a parameter of `managePackagesRpm`. That way, `getPackageManager` can be removed from `distroHandler`, which avoids having RPM specific types in `distroHandler`.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -600,10 +601,12 @@ func customizeOSContents(ctx context.Context, ic *ImageCustomizerParameters) err
 		ic.rawImageFile = newRawImageFile
 	}
 
+	// TODO: Add support for fedora in image customizer
+	distroHandler := NewDistroHandler(string(distroNameAzureLinux), "3.0")
```
cwize1, line 605: Image Customizer supports Azure Linux 2.0.

toolkit/tools/pkg/imagecustomizerlib/azurelinux_distro_config.go
himaja-kesari: you are correct, will add the support when adding support for 4.0, for now Azure Linux always uses tdnf

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
himaja-kesari: removed the todo comment

toolkit/tools/pkg/imagecustomizerlib/distrohandler_fedora.go
himaja-kesari: updated

toolkit/tools/pkg/imagecustomizerlib/customizepackages_rpm.go
himaja-kesari: updated

toolkit/tools/pkg/imagecustomizerlib/distrohandler.go
himaja-kesari: updated the code

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
himaja-kesari: updated to use the imageconnection object to determine the installed os in the base image.

cwize1 (commented):
LGTM.

Holding off approval until v1 is finalized.

PR #387, opened 2025-08-19
Boost functional test timeout

cwize1:
Boost the test suite timeout to avoid the tests failing while they are still running.

---

### **Checklist**
- [x] Tests added/updated


himaja-kesari:
has this been happening recently because of newly added tests ?

cwize1:
> has this been happening recently because of newly added tests ?

Yes, plus some tests were accidentally disabled and were enabled.

Britel (approved):

PR #398, opened 2025-08-27
Image Customizer: Allow files larger than 4GB.

gmileka:
Image Customizer: Allow files larger than 4GB.

---

- [n/a] Tests added/updated


toolkit/tools/internal/isogenerator/isogenerator.go
```diff
@@ -88,7 +88,10 @@ func BuildIsoImage(stagingPath string, enableBiosBoot bool, isoOsFilesDirPath st
 
 	mkisofsArgs = append(mkisofsArgs,
```
cwize1, line 89: So, technically the `mkisofs` command doesn't exist anymore. Nowadays the command is a proxy for either `genisoimage` or `xorrisofs`. Though `genisoimage` is also pretty defunct as well, since its website is offline. Whereas `xorrisofs` is still maintained.

The `-udf -allow-limited-size` args are specific to `genisoimage`. So, if a user's `mkisofs` is a proxy to `xorrisofs`, then this call will fail.

One option would be to explicitly use `genisoimage`, to ensure that these args are available.

Another option would be to try using `xorrisofs` instead. I think it might handle large files without any extra params.

toolkit/tools/internal/isogenerator/isogenerator.go
```diff
@@ -88,7 +88,10 @@ func BuildIsoImage(stagingPath string, enableBiosBoot bool, isoOsFilesDirPath st
...
 		// General mkisofs parameters.
-		"-R", "-l", "-D", "-J", "-joliet-long", "-o", outputImagePath, "-V", DefaultVolumeId)
+		"-R", "-l", "-D",
+		"-iso-level", "3", "-udf", "-allow-limited-size", // allow files larger than 4GB.
```
cwize1, line 92: `GiB`

cwize1 (commented):
The PR title should be more specific. And the PR description should have more details.

toolkit/tools/internal/isogenerator/isogenerator.go
himaja-kesari, line 92: why is this change needed ? can you explain the use case for my understanding

himaja-kesari, line 92: is "-iso-level", "3" still applicable here

gmileka: fixed.

gmileka: Yep, level 3 means a file can be split into more than one section. Given that a section can hold up to 4GiB - we'd need more than one section to hold files larger than 4GiB - so, this flag enables it.

toolkit/tools/internal/isogenerator/isogenerator.go
cwize1: The documentation for `genisoimage` has this:

> -udf Include UDF filesystem support in the generated filesystem image.  UDF support is currently in alpha status and for this reason, it is not possible to create UDF-only images. UDF data structures are currently coupled to the Joliet structures, so there are many pitfalls with the current implementation. There is no UID/GID support, there is no POSIX permission support, there is no support for symlinks.  Note that UDF wastes the space from sector ~20 to sector 256 at the beginning of the disc in addition to the space needed for real UDF data structures.

That doesn't exactly inspire a great deal of confidence.

Also, `genisoimage` is no longer maintained. It doesn't even have a website anymore. So, I'd rather not take a dependency on it.

cwize1 (changes requested):

gmileka:
Closing since genisoimage is an older tool that is not well maintained anymore and also its support for udf is alpha quality. We'd rather find better way to support udf. We will revisit.

PR #399, opened 2025-08-27
Use xorriso

gmileka:
---

### **Checklist**
- [ ] Tests added/updated


fintelia:
I think also passing `--efi-boot-part --efi-boot-image` to xorriso may enable the resulting ISO to directly be used as a bootable USB

gmileka:
> I think also passing `--efi-boot-part --efi-boot-image` to xorriso may enable the resulting ISO to directly be used as a bootable USB

I've created a separate bug to track this work (15053).

PR #405, opened 2025-08-28
Workflows: Add tdnf retry loop.

cwize1:
In the workflows, add a retry loop to the `tdnf` install command. This will hopefully prevent the `error: can't create transaction lock on /var/lib/rpm/.rpm.lock (Resource temporarily unavailable)` error in the future.

---

### **Checklist**

- [x] Tests added/updated

Workflow run: https://github.com/microsoft/azure-linux-image-tools/actions/runs/17310281055

### DIFF
--- a/.github/workflows/tests-functional.yml
+++ b/.github/workflows/tests-functional.yml
@@ -58,7 +58,7 @@ jobs:
             PACKAGES+=" grub2-pc systemd-ukify"
         fi
 
-        ./repo/.github/workflows/scripts/retry.sh 5 5 \
+        sudo ./repo/.github/workflows/scripts/retry.sh 5 5 \
             tdnf install -y $PACKAGES
 
         sudo tdnf list installed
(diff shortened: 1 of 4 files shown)


.github/workflows/tests-functional.yml
```diff
@@ -42,16 +42,28 @@ jobs:
...
+            sed createrepo_c squashfs-tools cdrkit e2fsprogs dosfstools \
+            xfsprogs zstd veritysetup grub2 binutils lsof \
+            git azure-cli"
         # grub2-pc is only available on x86.
         if [[ "$HOST_ARCH" == "amd64" ]]; then
-          sudo tdnf install -y grub2-pc systemd-ukify
+          PACKAGES+=" grub2-pc systemd-ukify"
         fi
 
+        # Run a retry loop to handle transient networking issues and background RPM installs.
```
gmileka, line 54: Should we factor this out into a script the yml files call into?

himaja-kesari (approved):

PR #406, opened 2025-08-29
Improve container file paths.

cwize1:
Move the files within the container to more appropriate directories.

Specifically, place all miscellaneous files under `/usr/lib/imagecustomizer`. Also, don't use `local` directory to keep the paths consistent with the RPM in the [azurelinux](https://github.com/microsoft/azurelinux) repo.

---

### **Checklist**

- [x] Tests added/updated


liulanze (approved):
LGTM, the downstream RPM PR has also been revised.

PR #418, opened 2025-09-04
extract packages from the config file

himaja-kesari:
---

### **Checklist**
- [ ] Tests added/updated

### DIFF
--- a/toolkit/tools/internal/testutils/testrpms/download-test-utils.sh
+++ b/toolkit/tools/internal/testutils/testrpms/download-test-utils.sh
@@ -45,16 +45,52 @@ esac
 
 set -x
 
-# call the script to create the tools file if requested
+# Declarative configuration
+TESTDATA_DIR="$SCRIPT_DIR/../../../pkg/imagecreatorlib/testdata"
+DISTRO="azurelinux"
+CONFIG_FILE="minimal-os.yaml"
+
+# Initialize package list
+PACKAGE_LIST=""
+
+# Handle tools file creation and package extraction for imagecreator testing
 if [ "$IMAGE_CREATOR" = "true" ]; then
   echo "Creating tools file: $TOOLS_FILE"
   $SCRIPT_DIR/create-tools-file.sh "$CONTAINER_IMAGE" "$TOOLS_FILE"
   echo "Tools file created successfully."
+
+  # Check for python3 availability
+  if ! command -v python3 >/dev/null 2>&1; then
+    echo "Error: python3 is required but not found in PATH"
+    echo "Please install python3 to extract package lists from config files"
+    exit 1
+  fi
+
+  # Extract package list from config file
+  CONFIG_PATH="$TESTDATA_DIR/$CONFIG_FILE"
+  if [[ -f "$CONFIG_PATH" ]]; then
+    PACKAGE_LIST=$(python3 "$SCRIPT_DIR/extract_packages.py" "$CONFIG_PATH")
+    echo "Package list from $CONFIG_FILE: $PACKAGE_LIST"
+  else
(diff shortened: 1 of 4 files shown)


toolkit/tools/internal/testutils/testrpms/extract_packages.py
```diff
@@ -0,0 +1,86 @@
...
+            data = yaml.safe_load(f)
+            
+            # Handle direct package list (like in golang.yaml)
+            if 'packages' in data and data['packages']:
+                return data['packages']
+            
+            packages = data.get('os', {}).get('packages', {}).get('install', [])
+            return packages
+            
+    except FileNotFoundError:
```
cwize1, line 24: Is there a reason why you are handling these exceptions? Why not just let the script fail?

toolkit/tools/internal/testutils/testrpms/extract_packages.py
```diff
@@ -0,0 +1,86 @@
...
+        print(f"Error reading file '{config_file_path}': {e}", file=sys.stderr)
+        return []
+
+def extract_packages(config_file_path):
+    """Extract packages from a YAML configuration file, including installLists."""
+    all_packages = []
+    
+    try:
+        with open(config_file_path, 'r') as f:
+            data = yaml.safe_load(f)
```
cwize1, line 40: After `yaml.safe_load`, you don't need to keep the file open.

toolkit/tools/internal/testutils/testrpms/download-test-utils.sh
```diff
@@ -45,16 +45,52 @@ esac
...
+  # Check for python3 availability
+  if ! command -v python3 >/dev/null 2>&1; then
+    echo "Error: python3 is required but not found in PATH"
+    echo "Please install python3 to extract package lists from config files"
+    exit 1
+  fi
+
+  # Extract package list from config file
+  CONFIG_PATH="$TESTDATA_DIR/$CONFIG_FILE"
+  if [[ -f "$CONFIG_PATH" ]]; then
```
cwize1, line 71: Reverse this `if` statement. Then, remove the `else` block because `exit 1` makes it redundant.

toolkit/tools/internal/testutils/testrpms/download-test-utils.sh
```diff
@@ -45,16 +45,52 @@ esac
...
 fi
 
-# call the script to download the rpms
-echo "Downloading test rpms for Azure Linux version: $IMAGE_VERSION"
-$SCRIPT_DIR/download-test-rpms.sh "$CONTAINER_IMAGE"  "$IMAGE_VERSION" "$IMAGE_CREATOR" 
-echo "Test rpms downloaded successfully."
+# Combine with common testing packages
+COMMON_PACKAGES="jq golang"
+if [[ -n "$PACKAGE_LIST" ]]; then
+    FINAL_PACKAGE_LIST="$COMMON_PACKAGES $PACKAGE_LIST"
```
cwize1, line 86: Having extra spaces probably wouldn't cause any issues. So, you could simplify this to `FINAL_PACKAGE_LIST="jq golang $PACKAGE_LIST"`

cwize1 (changes requested):

toolkit/tools/internal/testutils/testrpms/extract_packages.py
himaja-kesari: updated the script to fail

toolkit/tools/internal/testutils/testrpms/extract_packages.py
himaja-kesari: updated!

cwize1 (dismissed):

PR #442, opened 2025-09-23
Reorganize code slightly.

cwize1:
Move the config validation logic into its own file. Also, put all the verity logic in a single file.

---

### **Checklist**

- [x] Tests added/updated


toolkit/tools/pkg/imagecustomizerlib/validation.go
```diff
@@ -0,0 +1,324 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
```
aditjha-msft, line 3: Should we maybe call this `configvalidation.go`?

aditjha-msft (approved):
lgtm

PR #455, opened 2025-10-02
Base Configs Support - Phase 1 

elainezhao1:
This PR introduces hierarchical configuration support in Image Customizer, allowing users to compose configs on top of one another via the .baseConfigs field.

Implement circular dependency detection
Create config inheritance chain resolution
include fields below:
- .input.image
- .output.image
- .output.artifacts

---

### **Checklist**
- [x] Tests added/updated

### DIFF
--- a/docs/imagecustomizer/concepts/iso.md
+++ b/docs/imagecustomizer/concepts/iso.md
@@ -13,6 +13,10 @@ as a [Live OS](./liveos.md) iso image. A Live OS iso image is a bootable image
 that boots into a root file system included on the iso media without the need
 to have anything pre-installed on the target machine.
 
+The ISO file's format is an ISO 9660 with the following extensions: Joilet, Rock Ridge,
+and El Torito. The ISO file is given a volume ID of `CDROM`. The ISO file cannot contain
+files that have a size greater than 4 GiB.
+
 ## Creating a Live OS ISO
 
 The input image can be a full disk image (vhd/vhdx/qcow2/raw) or previously
--- a/toolkit/tools/pkg/imagecustomizerlib/main_test.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/main_test.go
@@ -91,6 +91,7 @@ var (
 	baseImageCoreEfiAzl3   = flag.String("base-image-core-efi-azl3", "", "An Azure Linux 3.0 core-efi image to use as a base image.")
 	baseImageBareMetalAzl2 = flag.String("base-image-bare-metal-azl2", "", "An Azure Linux 2.0 bare-metal image to use as a base image.")
 	baseImageBareMetalAzl3 = flag.String("base-image-bare-metal-azl3", "", "An Azure Linux 3.0 bare-metal image to use as a base image.")
+	logLevel               = flag.String("log-level", "info", "The log level (error, warning, info, debug, or trace)")
 )
 
 var (
@@ -106,6 +107,13 @@ func TestMain(m *testing.M) {
 	logger.InitStderrLog()
 
 	flag.Parse()
+
+	if logLevel != nil {
+		err := logger.SetStderrLogLevel(*logLevel)
+		if err != nil {
+			logger.Log.Panicf("Failed to set log level, error: %s", err)
+		}
+	}
 
 	workingDir, err = os.Getwd()
 	if err != nil {


toolkit/tools/pkg/imagecustomizerlib/baseconfigs_test.go
```diff
@@ -0,0 +1,80 @@
+package imagecustomizerlib_test
+
+import (
+	"testing"
+
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/imagecustomizerapi"
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/pkg/imagecustomizerlib"
+)
+
+func TestResolveOverrideFields_OutputOverrides(t *testing.T) {
```
himaja-kesari, line 10: also add a test for .input.image.path overrides

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -543,6 +543,12 @@ func customizeOSContents(ctx context.Context, ic *ImageCustomizerParameters) err
 	ctx, span := otel.GetTracerProvider().Tracer(OtelTracerName).Start(ctx, "customize_os_contents")
 	defer span.End()
 
+	resolvedConfig, err := LoadBaseConfig(ic.config, ic.configPath)
+	if err != nil {
+		return fmt.Errorf("failed to load base config:\n%w", err)
+	}
+	ic.config = resolvedConfig.Config
```
himaja-kesari, line 550: when you are reassigning the config with the resolved config it would only populate the output, input and artifacts fields and not the other fields like storage and os, etc.

toolkit/tools/pkg/imagecustomizerlib/baseconfigs_test.go
```diff
@@ -0,0 +1,80 @@
...
+			Artifacts: &imagecustomizerapi.Artifacts{
+				Path: "./current-artifacts",
+				Items: []imagecustomizerapi.OutputArtifactsItemType{
+					imagecustomizerapi.OutputArtifactsItemShim,
+				},
+			},
+		},
+	}
+
+	resolved := &imagecustomizerlib.ResolvedConfig{
```
himaja-kesari, line 39: see my comment above to create and reuse NewResolvedConfig function

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
himaja-kesari: Ic config with this assignment changes Image Customizer Config: &{Input:{Image:{Path:}} Storage:{ResetPartitionsUuidsType: BootType: Disks:[] FileSystems:[] Verity:[] ReinitializeVerity: VerityPartitionsType:0} Iso:<nil> Pxe:<nil> OS:<nil> Scripts:{PostCustomization:[] FinalizeCustomization:[]} PreviewFeatures:[] Output:{Image:{Path: Format:} Artifacts:0xc0000497d0} BaseConfigs:[]}

toolkit/tools/pkg/imagecustomizerlib/baseconfigs.go
```diff
@@ -0,0 +1,145 @@
...
+	return chain, nil
+}
+
+func (r *ResolvedConfig) Resolve() {
+	r.resolveOverrideFields()
+	r.resolveMergeFields()
+}
+
+func (r *ResolvedConfig) resolveOverrideFields() {
+	for _, cfg := range r.InheritanceChain {
```
cwize1, line 91: Thinking about it, for interlaced operations and simple merges, following the `InheritanceChain` ordering is probably fine.

Though, it's possible that for more complex merges, we may want to merge following the tree structure. So, it might be good to store the tree as well as the chain.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
elainezhao1: Thanks for pointing out! Will update the struct to resolve this

toolkit/tools/pkg/imagecustomizerlib/baseconfigs.go
```diff
@@ -0,0 +1,145 @@
...
+	resolved := &ResolvedConfig{
+		InheritanceChain: configChain,
+		Config: &imagecustomizerapi.Config{
+			Input: imagecustomizerapi.Input{
+				Image: imagecustomizerapi.InputImage{},
+			},
+			Output: imagecustomizerapi.Output{
+				Image: imagecustomizerapi.OutputImage{},
+				Artifacts: &imagecustomizerapi.Artifacts{
+					Items: []imagecustomizerapi.OutputArtifactsItemType{},
```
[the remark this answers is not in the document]
liulanze: +1, FYI the path is either relative or absolute for _all_ of the artifacts.

toolkit/tools/pkg/imagecustomizerlib/baseconfigs.go
```diff
@@ -0,0 +1,145 @@
...
+		if val := strings.TrimSpace(cfg.Output.Image.Path); val != "" {
+			r.Config.Output.Image.Path = val
+		}
+
+		// .output.image.format
+		if val := strings.TrimSpace(string(cfg.Output.Image.Format)); val != "" {
+			r.Config.Output.Image.Format = imagecustomizerapi.ImageFormatType(val)
+		}
+
+		// .output.image.artifacts.path
```
liulanze, line 107: Was the `artifacts.path` the only one to be checked here? No need for `artifacts.items`?

toolkit/tools/pkg/imagecustomizerlib/baseconfigs.go
```diff
@@ -0,0 +1,145 @@
...
+		if val := strings.TrimSpace(cfg.Output.Artifacts.Path); val != "" {
```
liulanze, line 108: `cfg.Output` or `cfg.Output.Artifacts` may be nil. It may be good to check before deref, like:

```
if cfg.Output.Artifacts != nil { ... }
```

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -543,6 +540,16 @@ func customizeOSContents(ctx context.Context, ic *ImageCustomizerParameters) err
 	ctx, span := otel.GetTracerProvider().Tracer(OtelTracerName).Start(ctx, "customize_os_contents")
 	defer span.End()
 
+	resolvedConfig, err := loadBaseConfig(ic.config, ic.configPath)
```
cwize1, line 543: The base configs should be loaded early during customization. You probably should move `loadBaseConfig` into `CustomizeImage()`.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -543,6 +540,16 @@ func customizeOSContents(ctx context.Context, ic *ImageCustomizerParameters) err
...
+	if err != nil {
+		return fmt.Errorf("failed to load base config:\n%w", err)
+	}
+	ic.config.Input.Image.Path = resolvedConfig.InputImagePath
```
cwize1, line 547: The final input image and output image are resolved in the `createImageCustomizerParameters` function.

toolkit/tools/imagecustomizerapi/baseconfigs.go
```diff
@@ -0,0 +1,31 @@
...
+	if strings.TrimSpace(b.Path) == "" {
+		return fmt.Errorf("path must not be empty or whitespace")
+	}
+	return nil
+}
+
+func (b BaseConfigs) IsValid() error {
+	for i, base := range b {
+		if err := base.IsValid(); err != nil {
+			return fmt.Errorf("invalid baseConfig item at index %d:\n%w", i, err)
```
cwize1, line 27: It would probably be better to delete the `BaseConfigs` type and just move the `for` loop into `Config.IsValid`.

toolkit/tools/pkg/imagecustomizerlib/resolvedconfig.go
```diff
@@ -0,0 +1,27 @@
...
+	InputImagePath       string
+	OutputImagePath      string
+	OutputImageFormat    imagecustomizerapi.ImageFormatType
+	OutputArtifactsPath  string
+	OutputArtifactsItems []imagecustomizerapi.OutputArtifactsItemType
+
+	InheritanceChain []*imagecustomizerapi.Config
+}
+
+func NewResolvedConfig(chain []*imagecustomizerapi.Config) *ResolvedConfig {
```
cwize1, line 18: You probably don't need the `NewResolvedConfig` anymore. Just move the code directly into `loadBaseConfig`.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/resolvedconfig.go
elainezhao1: I think it'd probably be good to keep NewResolvedConfig for functional tests. and btw I updated loadBaseConfig to resolveBaseConfigs, I think resolveBaseConfigs require a real config file which will have a path, which is difficult to be used by functional tests if test is not file-based.

toolkit/tools/pkg/imagecustomizerlib/baseconfigs_test.go
```diff
@@ -0,0 +1,80 @@
...
+			Artifacts: &imagecustomizerapi.Artifacts{
+				Path: "./base-artifacts",
+				Items: []imagecustomizerapi.OutputArtifactsItemType{
+					imagecustomizerapi.OutputArtifactsItemUkis,
+				},
+			},
+		},
+	}
+
+	current := &imagecustomizerapi.Config{
```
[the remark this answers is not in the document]
elainezhao1: Added test for the malformed case. Thanks

toolkit/tools/pkg/imagecustomizerlib/baseconfigs.go
```diff
@@ -0,0 +1,169 @@
...
+		rc.Config.Output = merged.Config.Output
+	}
+
+	return nil
+}
+
+func BuildInheritanceChain(ctx context.Context, cfg *imagecustomizerapi.Config, baseDir string, visited map[string]bool, pathStack []string) ([]*imagecustomizerapi.Config, error) {
+	if cfg.BaseConfigs != nil {
+		for i, base := range cfg.BaseConfigs {
+			err := base.IsValid()
```
cwize1: These checks are already handled by `Config.IsValid`. You don't need to repeat them again here.

toolkit/tools/pkg/imagecustomizerlib/configvalidation.go
```diff
@@ -54,6 +54,8 @@ func ValidateConfig(ctx context.Context, baseConfigPath string, config *imagecus
 		Options:        options,
 	}
 
+	ResolveBaseConfigs(ctx, rc)
```
cwize1: I would probably move this below the `config.IsValid()` and `options.IsValid()` checks, just so you know the config has basic validity before trying to merge the base configs.

toolkit/tools/pkg/imagecustomizerlib/baseconfigs.go
```diff
@@ -0,0 +1,169 @@
...
+				return nil, fmt.Errorf("invalid baseConfig item at index %d:\n%w", i, err)
+			}
+		}
+	}
+
+	var chain []*imagecustomizerapi.Config
+
+	for _, base := range cfg.BaseConfigs {
+		basePath := filepath.Join(baseDir, base.Path)
+		absPath, err := filepath.Abs(basePath)
```
cwize1: Use `file.GetAbsPathWithBase` instead of `filepath.Join` and `filepath.Abs`. This will properly handle the case where the base config path is specified as an absolute path.

toolkit/tools/pkg/imagecustomizerlib/baseconfigs.go
```diff
@@ -0,0 +1,169 @@
...
+			return nil, fmt.Errorf("cycle detected in baseConfigs: %v -> %s", pathStack, absPath)
+		}
+
+		visited[absPath] = true
+		pathStack = append(pathStack, absPath)
+
+		// Load base file into struct
+		var baseCfg imagecustomizerapi.Config
+		if err := imagecustomizerapi.UnmarshalYamlFile(absPath, &baseCfg); err != nil {
+			return nil, fmt.Errorf("failed to load base config %s: %w", absPath, err)
```
cwize1: `"failed to load base config (%s):\n%w"`

toolkit/tools/pkg/imagecustomizerlib/baseconfigs.go
```diff
@@ -0,0 +1,122 @@
...
+	// Add the current config last
+	chain = append(chain, &ConfigWithBasePath{
+		Config:         cfg,
+		BaseConfigPath: configFilePath,
+	})
+
+	return chain, nil
+}
+
+func resolveOutputArtifacts(configChain []*ConfigWithBasePath) *imagecustomizerapi.Artifacts {
```
cwize1: I would move `resolveOutputArtifacts` and `mergeOutputArtifactTypes` into `configvalidation.go` so that all the merge logic lives in a single file.

toolkit/tools/pkg/imagecustomizerlib/baseconfigs.go
```diff
@@ -0,0 +1,122 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/imagecustomizerapi"
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/file"
+)
+
+type ConfigWithBasePath struct {
+	*imagecustomizerapi.Config
```
cwize1: nit: `Config *imagecustomizerapi.Config`

Omitting the variable name does a weird form on inheritance which I am not a fan of.

toolkit/tools/pkg/imagecustomizerlib/configvalidation.go
```diff
@@ -130,30 +135,38 @@ func ValidateConfig(ctx context.Context, baseConfigPath string, config *imagecus
...
 
-	case input.Image.Path != "":
-		inputImageAbsPath := file.GetAbsPathWithBase(baseConfigPath, input.Image.Path)
-		if yes, err := file.IsFile(inputImageAbsPath); err != nil {
-			return "", fmt.Errorf("%w (path='%s'):\n%w", ErrInvalidInputImageFileConfig, input.Image.Path, err)
-		} else if !yes {
-			return "", fmt.Errorf("%w (path='%s')", ErrInputImageFileNotFile, input.Image.Path)
+	// Resolve input image path
+	var inputImageAbsPath string
+	for _, configWithBase := range configChain {
```
cwize1: If you iterate through this list in reverse (`slices.Backward`), then you could early return as soon as you find a non-empty value. This would probably make the code a little cleaner. (The `return "", ErrInputImageFileRequired` would happen at the end of the function after all possible sources have been exhausted.)

You could also do a similar pattern for `outputImageFormat` and `outputImageFile`. (You'll need separate `validateOutputImageFormat()` and `validateOutputImageFile` functions to do this.)

toolkit/tools/imagecustomizerapi/schema.json
```diff
@@ -45,6 +45,15 @@
       "additionalProperties": false,
```
cwize1: Remember to update the docs.

toolkit/tools/pkg/imagecustomizerlib/testdata/base-config-malformed.yaml
```diff
@@ -0,0 +1,7 @@
+previewFeatures:
```
cwize1: What's malformed about this config file?

toolkit/tools/pkg/imagecustomizerlib/configvalidation.go
```diff
@@ -64,6 +64,11 @@ func ValidateConfig(ctx context.Context, baseConfigPath string, config *imagecus
 		return nil, err
 	}
 
+	configChain, err := buildConfigChain(ctx, rc)
```
cwize1: Store the `ConfigChain` in `ResolvedConfig`. It'll be needed by the features that walk the chain (e.g. packages).

toolkit/tools/pkg/imagecustomizerlib/baseconfigs_test.go
```diff
@@ -0,0 +1,91 @@
...
+	// Verify merged artifact items
+	expectedItems := []imagecustomizerapi.OutputArtifactsItemType{
+		imagecustomizerapi.OutputArtifactsItemUkis,
+		imagecustomizerapi.OutputArtifactsItemShim,
+	}
+	actual := rc.Config.Output.Artifacts.Items
+	assert.Equal(t, len(expectedItems), len(actual))
+
+	for _, item := range expectedItems {
+		assert.Containsf(t, actual, item, "expected output artifact item %q not found in resolved config: %v",
```
cwize1: You can use the `assert.ElementsMatch` to compare two lists while ignoring item ordering.

toolkit/tools/pkg/imagecustomizerlib/baseconfigs_test.go
```diff
@@ -0,0 +1,91 @@
...
+		BuildDir: buildDir,
+	}
+
+	var config imagecustomizerapi.Config
+	err := imagecustomizerapi.UnmarshalYamlFile(currentConfigFile, &config)
+	assert.NoError(t, err)
+
+	_, err = ValidateConfig(t.Context(), testDir, &config, false, options)
+
+	assert.ErrorContains(t, err, ErrInvalidBaseConfigs.Error())
```
cwize1: `assert.ErrorIs(t, err, ErrInvalidBaseConfigs)`

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/testdata/current-config.yaml
```diff
@@ -0,0 +1,22 @@
+previewFeatures:
```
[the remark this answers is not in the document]
cwize1: It might be good to give all the configs in the same hierarchy the same name prefix. For example: `malformed-tree-base.yaml` and `malformed-tree.yaml`. That way, all the related configs will sit next to each other.

toolkit/tools/pkg/imagecustomizerlib/testdata/base-config-malformed.yaml
cwize1: Let's keep the `baseCfg.IsValid` in `buildConfigChainHelperValidation` for now. (We can revisit this decision in the future if it becomes a problem.)

But for specific individual fields (e.g. output path, input path), we can allow the check to happen against the merged config.

docs/imagecustomizer/api/configuration/config.md
```diff
@@ -104,6 +104,15 @@ Supported options:
...
+- `base-configs`: Enables support for hierarchical configuration inheritance.
+
+  When this option is specified, the `baseConfigs` configuration becomes available.
+  This allows configurations to inherit from one or more base configurations.
+
+  See [Base Config](./baseConfig.md) for more details.
+
+  Added in v1.1.0.
+
 ## output [[output](./output.md)]
```
cwize1: You need to add the `baseConfigs` field to this file.

toolkit/tools/pkg/imagecustomizerlib/resolvedconfig.go
```diff
@@ -29,6 +29,9 @@ type ResolvedConfig struct {
 	// Input image
 	InputImageFile string
 
+	// Output artifacts
+	OutputArtifacts *imagecustomizerapi.Artifacts
```
cwize1: Use this value in the `CustomizeImageOptions` function?

toolkit/tools/pkg/imagecustomizerlib/baseconfigs_test.go
```diff
@@ -0,0 +1,51 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/imagecustomizerapi"
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/file"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestBaseConfigsInputAndOutput(t *testing.T) {
```
cwize1: It would be good to have at least one test that actually runs through an actual image customization.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/baseconfigs_test.go
elainezhao1: to add a full run functional test, any suggestions on how I can obtain a test image from which I can extract 2 artifacts (to test merged artifact items) ? I think current empty.vhdx would not work. if it is not ok to check in a non empty vhdx, can I exclude the artifacts item in the full run?

toolkit/tools/pkg/imagecustomizerlib/baseconfigs_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+	currentConfigFile := filepath.Join(testDir, "hierarchical-config.yaml")
+
+	err := CustomizeImageWithConfigFile(t.Context(), buildDir, currentConfigFile, baseImage, nil,
+		outImageFile, "vhdx", false, "")
+	if !assert.NoError(t, err) {
+		return
+	}
+
+	assert.FileExists(t, outImageFile)
```
cwize1: It would be good to actually verify what's in the image.

For example, see `TestCustomizeImageHostname` for how to verify the hostname.

PR #457, opened 2025-10-02
Enhance Error Checking 

himaja-kesari:
Enhanced error assertion in tests to verify the specific error type.

## Changes
Modified test assertion to check 
- The specific error type using `ErrPackageRepoMetadataRefresh`

### **Checklist**
- [ ] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/customizepackages_test.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/customizepackages_test.go
@@ -288,7 +288,7 @@ func TestCustomizeImagePackagesBadRepo(t *testing.T) {
 	// Customize image.
 	err := CustomizeImageWithConfigFile(t.Context(), buildDir, configFile, baseImage, []string{repoFile}, outImageFilePath, "raw",
 		true /*useBaseImageRpmRepos*/, "" /*packageSnapshotTime*/)
-	assert.ErrorContains(t, err, "failed to refresh tdnf repo metadata")
+	assert.ErrorContains(t, err, ErrPackageRepoMetadataRefresh.Error())
 }
 
 func ensureTdnfCacheCleanup(t *testing.T, imageConnection *imageconnection.ImageConnection, dirPath string,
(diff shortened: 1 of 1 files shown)


toolkit/tools/pkg/imagecustomizerlib/customizepackages_test.go
```diff
@@ -288,7 +288,7 @@ func TestCustomizeImagePackagesBadRepo(t *testing.T) {
 	// Customize image.
 	err := CustomizeImageWithConfigFile(t.Context(), buildDir, configFile, baseImage, []string{repoFile}, outImageFilePath, "raw",
 		true /*useBaseImageRpmRepos*/, "" /*packageSnapshotTime*/)
-	assert.ErrorContains(t, err, "failed to refresh tdnf repo metadata")
+	assert.ErrorContains(t, err, ErrPackageRepoMetadataRefresh.Error())
```
cwize1, line 291: `assert.ErrorIs(t, err, ErrPackageRepoMetadataRefresh)`?

himaja-kesari: I'm leaning toward using ErrorContains because it’s more resilient to future changes. If we ever update the error message, ErrorIs might fail due to strict matching, whereas ErrorContains would still pass as long as the key part of the message remains. It gives us flexibility

cwize1: `Error` checks that the error message matches exactly.

`ErrorContains` checks the error message contains a string.

`ErrorIs` checks if the error object contains the object provided.

cwize1 (approved):

toolkit/tools/pkg/imagecustomizerlib/customizepackages_test.go
himaja-kesari: thanks for the explanation, updated to ErrorIs

PR #479, opened 2025-10-20
Improve SELinux Mode Detection with Enhanced GRUB Fallback Logic

himaja-kesari:
This PR enhances the SELinux mode detection in BootCustomizer by implementing robust fallback logic between GRUB configuration variables and providing comprehensive error reporting. The changes improve reliability when detecting SELinux configuration in different GRUB setups, particularly for Fedora distributions.

Problem:
The existing SELinux mode detection logic in getSELinuxModeFromGrub() was fragile and could fail when:

GRUB_CMDLINE_LINUX variable was not present in grub
Only GRUB_CMDLINE_LINUX_DEFAULT contained the SELinux configuration

changes made :
 Try to extract SELinux args from GRUB_CMDLINE_LINUX (higher priority) and also  GRUB_CMDLINE_LINUX_DEFAULT
Enhanced error reporting: Capture and report both errors if both attempts fail

### **Checklist**
- [ ] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/defaultgrubutils.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/defaultgrubutils.go
@@ -208,8 +208,8 @@ func GetDefaultGrubFileLinuxArgsFromMultipleVars(defaultGrubFileContent string) 
 	}
 
 	if !foundAny {
-		return nil, fmt.Errorf("failed to find any GRUB command line variables (GRUB_CMDLINE_LINUX or GRUB_CMDLINE_LINUX_DEFAULT) in %s",
-			installutils.GrubDefFile)
+		return nil, fmt.Errorf("failed to find any of the specified GRUB variables in %s: %v",
+			installutils.GrubDefFile, varNames)
 	}
 
 	return allArgs, nil


toolkit/tools/pkg/imagecustomizerlib/bootcustomizer.go
```diff
@@ -11,10 +11,8 @@ import (
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/safechroot"
 )
 
-var (
-	// Boot customization errors
-	ErrBootGrubMkconfigGeneration = NewImageCustomizerError("Boot:GrubMkconfigGeneration", "failed to generate grub.cfg via grub2-mkconfig")
-)
+// Boot customization errors
```
cwize1: Undo change

himaja-kesari: for a single variable the () are not needed

toolkit/tools/pkg/imagecustomizerlib/bootcustomizer.go
```diff
@@ -90,8 +88,14 @@ func (b *BootCustomizer) getSELinuxModeFromGrub() (imagecustomizerapi.SELinuxMod
 	if b.isGrubMkconfig {
 		_, args, _, err = GetDefaultGrubFileLinuxArgs(b.defaultGrubFileContent, defaultGrubFileVarNameCmdlineForSELinux)
 		if err != nil {
```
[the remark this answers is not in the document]
himaja-kesari: Yes, Fedora Cloud does not include the GRUB_CMDLINE_LINUX variable in its grub configuration, which causes IC to fail when trying to customize the image due to this error.

toolkit/tools/pkg/imagecustomizerlib/defaultgrubutils.go
```diff
@@ -175,6 +175,42 @@ func GetDefaultGrubFileLinuxArgs(defaultGrubFileContent string, varName defaultG
...
+// multiple GRUB variables like GRUB_CMDLINE_LINUX and GRUB_CMDLINE_LINUX_DEFAULT.
+//
+// Parameters:
+//   - defaultGrubFileContent: The content of the /etc/default/grub file
+//   - varNames: Variable names to check and merge arguments from
+//
+// Returns:
+//   - args: The merged list of kernel command-line args from all found variables
+//   - error: Error if parsing fails or no variables are found
+func GetDefaultGrubFileLinuxArgsFromMultipleVars(defaultGrubFileContent string, varNames ...defaultGrubFileVarName,
```
cwize1: There are only two variables. So, you could just embed the names within this function's implementation, instead of requiring callers to pass them in.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/bootcustomizer.go
```diff
@@ -88,9 +87,11 @@ func (b *BootCustomizer) getSELinuxModeFromGrub() (imagecustomizerapi.SELinuxMod
 
 	// Get the SELinux kernel command-line args.
 	if b.isGrubMkconfig {
-		_, args, _, err = GetDefaultGrubFileLinuxArgs(b.defaultGrubFileContent, defaultGrubFileVarNameCmdlineForSELinux)
+		// Check both GRUB_CMDLINE_LINUX and GRUB_CMDLINE_LINUX_DEFAULT variables
+		// and merge arguments from both if they exist
+		args, err = GetDefaultGrubFileLinuxArgsFromMultipleVars(b.defaultGrubFileContent)
 		if err != nil {
-			return "", err
+			return "", fmt.Errorf("failed to find SELinux args in grub file (%s): %w", installutils.GrubDefFile, err)
```
cwize1: Only adding an extra error message to this `if err != nil` check seems odd. Should there be symmetry with the `else` block? Perhaps add an extra error message in `GetSELinuxMode` against the call to `getLinuxCommandLineArgs`?

toolkit/tools/pkg/imagecustomizerlib/bootcustomizer.go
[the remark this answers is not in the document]
himaja-kesari: there are couple of places to be updated will do it in a separate PR

cwize1 (approved):

PR #485, opened 2025-10-23
Add BTRFS subvolume support for rootfs detection and fstab reading in image customizer

himaja-kesari:
---
This PR adds comprehensive BTRFS subvolume support to the image customizer library to properly handle BTRFS filesystems with subvolumes when detecting rootfs partitions and reading fstab entries. Previously, the image customizer only checked default mount locations, which would fail for BTRFS systems using subvolumes where /etc/fstab might be located in a specific subvolume rather than the filesystem root

### **Checklist**
- [ ] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
@@ -22,6 +22,7 @@ import (
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/safemount"
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/shell"
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/sliceutils"
+	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 )
 
@@ -105,12 +106,13 @@ func findBootPartitionFromEsp(efiSystemPartition *diskutils.PartitionInfo, diskP
 // Searches for the partition that contains the /etc/fstab file.
 // While technically it is possible to place /etc on a different partition, doing so is fairly difficult and requires
 // a custom initramfs module.
-func findRootfsPartition(diskPartitions []diskutils.PartitionInfo, buildDir string) (*diskutils.PartitionInfo, error) {
+func findRootfsPartition(diskPartitions []diskutils.PartitionInfo, buildDir string) (*diskutils.PartitionInfo, string, error) {
 	logger.Log.Debugf("Searching for rootfs partition")
 
 	tmpDir := filepath.Join(buildDir, tmpPartitionDirName)
 
 	var rootfsPartitions []*diskutils.PartitionInfo
+	var rootfsPaths []string
 	for i := range diskPartitions {
 		diskPartition := diskPartitions[i]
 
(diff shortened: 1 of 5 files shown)


toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -121,23 +121,34 @@ func findRootfsPartition(diskPartitions []diskutils.PartitionInfo, buildDir stri
...
+		// Handle BTRFS partitions specially
+		if diskPartition.FileSystemType == "btrfs" {
+			found, err := findRootfsInBtrfsPartition(&diskPartition, tmpDir)
+			if err != nil {
+				return nil, fmt.Errorf("failed to check BTRFS partition (%s) for rootfs:\n%w",
+					diskPartition.Path, err)
+			}
+			if found {
+				rootfsPartitions = append(rootfsPartitions, &diskPartition)
+			}
```
cwize1: Add `continue` here?

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -791,3 +808,153 @@ func getPartitionNum(partitionLoopDevice string) (int, error) {
...
+
+	// Close the current mount and mount each subvolume individually
+	err = partitionMount.CleanClose()
+	if err != nil {
+		return false, fmt.Errorf("failed to close partition mount (%s):\n%w", diskPartition.Path, err)
+	}
+
+	// Check each subvolume
+	for _, subvolume := range subvolumes {
+		subvolMount, err := mountBtrfsSubvolume(diskPartition.Path, subvolume, tmpDir)
```
cwize1: You need to have `defer subvolMount.Close()` and `subvolMount.CleanClose()`, same as the rest of the code.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -791,3 +808,153 @@ func getPartitionNum(partitionLoopDevice string) (int, error) {
...
+		}
+	}
+
+	logger.Log.Debugf("Found BTRFS subvolumes: %v", subvolumes)
+	return subvolumes, nil
+}
+
+// mountBtrfsSubvolume mounts a specific BTRFS subvolume
+func mountBtrfsSubvolume(devicePath, subvolume, mountPoint string) (*safemount.Mount, error) {
+	mountOptions := fmt.Sprintf("subvol=%s,ro", subvolume)
```
cwize1: In the [mount](https://www.man7.org/linux/man-pages/man2/mount.2.html) syscall, there are integer flags and additional string options. In the [mount](https://www.man7.org/linux/man-pages/man8/mount.8.html) command, all the options are specified as strings, but some are converted into integer flags when sent to the syscall. `ro` is one such flag. It is converted to `unix.MS_RDONLY`, which you already have.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -791,3 +808,153 @@ func getPartitionNum(partitionLoopDevice string) (int, error) {
...
+		if exists && subvolume != "" {
+			logger.Log.Debugf("Found fstab in BTRFS subvolume: %s", subvolume)
+		}
+
+		return exists, nil
+	})
+}
+
+// readFstabFromBtrfsPartition reads fstab entries from a BTRFS partition by searching through subvolumes
+func readFstabFromBtrfsPartition(rootfsPartition *diskutils.PartitionInfo, tmpDir string) ([]diskutils.FstabEntry,
```
cwize1: Return the subvolume from `findRootfsPartition` so that you don't have to find the subvolume again.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -791,3 +808,153 @@ func getPartitionNum(partitionLoopDevice string) (int, error) {
...
+		unix.MS_RDONLY, "", true)
+	if err != nil {
+		return false, fmt.Errorf("failed to mount partition (%s):\n%w", diskPartition.Path, err)
+	}
+	defer partitionMount.Close()
+
+	// List actual subvolumes using btrfs tools
+	subvolumes, err := listBtrfsSubvolumes(tmpDir)
+	if err != nil {
+		logger.Log.Debugf("Failed to list BTRFS subvolumes for partition %s:\n%v", diskPartition.Path, err)
```
cwize1: Don't ignore all errors.

Under what condition does `listBtrfsSubvolumes` fail but it is ok? Does `btrfs subvolume list` fail if there are no subvolumes? If so, then detect this case in `listBtrfsSubvolumes` and then return an empty list instead of an error.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -791,3 +808,153 @@ func getPartitionNum(partitionLoopDevice string) (int, error) {
...
+	err = partitionMount.CleanClose()
+	if err != nil {
+		return false, fmt.Errorf("failed to close partition mount (%s):\n%w", diskPartition.Path, err)
+	}
+
+	// Check each subvolume
+	for _, subvolume := range subvolumes {
+		subvolMount, err := mountBtrfsSubvolume(diskPartition.Path, subvolume, tmpDir)
+		if err != nil {
+			logger.Log.Debugf("Failed to mount BTRFS subvolume %s:\n%v", subvolume, err)
```
cwize1: Don't ignore errors. It is better to fail if there is something unexpected even if it is actually ok, than to silently continue when something is wrong.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -791,3 +808,153 @@ func getPartitionNum(partitionLoopDevice string) (int, error) {
...
+			continue
+		}
+
+		found, err := callback(subvolume, subvolMount)
+		subvolMount.Close() // Always close the subvolume mount
+
+		if err != nil {
+			logger.Log.Debugf("Error processing BTRFS subvolume %s:\n%v", subvolume, err)
```
cwize1: Ignoring the error here is silly and makes the `readFstabFromBtrfsPartition` function extremely confusing. If the error isn't a problem, then let the callback return `nil`.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -791,3 +808,153 @@ func getPartitionNum(partitionLoopDevice string) (int, error) {
...
+
+	for _, line := range lines {
+		if strings.TrimSpace(line) == "" {
+			continue
+		}
+
+		// Parse btrfs subvolume list output format:
+		// ID 256 gen 7 top level 5 path @
+		// ID 257 gen 7 top level 5 path @home
+		parts := strings.Fields(line)
```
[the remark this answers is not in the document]
himaja-kesari: fixed

cwize1 (approved):

PR #487, opened 2025-10-24
Make isPackageInstalled function generic 

himaja-kesari:
---
This PR refactors the isPackageInstalled function to be distribution-agnostic 

Todo - need to delete the old function and update few calls in liveiso.go and cosi common files
### **Checklist**
- [ ] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/customizeos.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/customizeos.go
@@ -99,7 +99,7 @@ func doOsCustomizations(ctx context.Context, rc *ResolvedConfig, imageConnection
 		return err
 	}
 
-	verityUpdated, err := enableVerityPartition(ctx, rc.Config.Storage.Verity, imageChroot)
+	verityUpdated, err := enableVerityPartition(ctx, rc.Config.Storage.Verity, imageChroot, distroHandler)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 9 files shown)


toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
```diff
@@ -62,13 +62,13 @@ func collectPackagesList(baseConfigPath string, packageLists []string, packages
...
+	// get target OS from the image
+	targetOs, err := targetos.GetInstalledTargetOs(imageChroot.RootDir())
 	if err != nil {
-		return false
+		return false, fmt.Errorf("unable to detect target OS while checking if package '%s' is installed: %w",
+			packageName, err)
 	}
-	return true
+
+	return NewDistroHandlerFromTargetOs(targetOs).isPackageInstalled(imageChroot, packageName), nil
```
cwize1: We eventually want to get to a point where the creation of the `distroHandler` only happens once at the start of the customization process. While plumbing this through the stack will be a bit of a pain, it will make future changes easier.

himaja-kesari: I’ve updated the calls to use distroHandler. Next, I’ll follow up to remove the old function and update any remaining references.

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
```diff
@@ -72,3 +72,10 @@ func isPackageInstalled(imageChroot safechroot.ChrootInterface, packageName stri
 	}
 	return true
 }
+
+// isPackageInstalledWithDistroHandler checks if a package is installed using the provided distro handler
+func isPackageInstalledWithDistroHandler(imageChroot safechroot.ChrootInterface, packageName string,
+	distroHandler distroHandler,
+) bool {
+	return distroHandler.isPackageInstalled(imageChroot, packageName)
```
cwize1: Why not just use `distroHandler.isPackageInstalled` directly?

himaja-kesari: updated!

cwize1 (approved):

PR #492, opened 2025-10-28
Add Hierarchical Config Support for .os.services, .os.packages, .os.groups. os.additionalFiles, os.modules

elainezhao1:
---

### **Checklist**
- [x] Tests added/updated

### DIFF
--- a/docs/imagecustomizer/api/cli/cli.md
+++ b/docs/imagecustomizer/api/cli/cli.md
@@ -124,6 +124,29 @@ When the output image format is set to iso, the generated image is a LiveOS
 iso image. For more details on this format, see:
 [Image Customizer ISO Support](../../concepts/iso.md).
 
+## --output-selinux-policy-path=DIRECTORY-PATH
+
+Optional.
+
+This is a preview feature.
+Its API and behavior is subject to change.
+You must enable this feature by specifying `output-selinux-policy` in the
+[previewFeatures](../configuration/config.md#previewfeatures-string) API.
+
+If both
+[output.selinuxPolicyPath](../configuration/output.md#selinuxpolicypath-string) and
+`--output-selinux-policy-path` are provided, then the `--output-selinux-policy-path`
+value is used.
+
+The directory path to output the SELinux policy files extracted from the customized
+image. The SELinux policy type is determined by reading the `SELINUXTYPE` value from
+`/etc/selinux/config` in the image (e.g., `targeted`), and the corresponding directory
+(e.g., `/etc/selinux/targeted`) will be extracted and copied to this location.
+
+If not specified, SELinux policy extraction is disabled.
+
+Added in v1.1.
+
 ## --config-file=FILE-PATH
 
 Required.
--- a/docs/imagecustomizer/api/configuration/config.md
+++ b/docs/imagecustomizer/api/configuration/config.md
@@ -117,6 +117,16 @@ Supported options:
 
   Added in v1.1.
 
+- `output-selinux-policy`: Enables extraction of SELinux policy files from the
+  customized image.
+
+  When this option is specified, the `output.selinuxPolicyPath` configuration
+  becomes available.
+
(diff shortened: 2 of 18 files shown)


toolkit/tools/pkg/imagecustomizerlib/customizeos.go
```diff
@@ -30,47 +30,49 @@ func doOsCustomizations(ctx context.Context, rc *ResolvedConfig, imageConnection
 		return err
 	}
 
-	err = addRemoveAndUpdatePackages(ctx, rc.BuildDirAbs, rc.BaseConfigPath, rc.Config.OS, imageChroot, nil,
```
cwize1: The ordering of these operations is considered part of the API and should not be changed.

Also, there should be a `for` loop for each API.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/customizeos.go
```diff
@@ -30,20 +30,25 @@ func doOsCustomizations(ctx context.Context, rc *ResolvedConfig, imageConnection
...
-		rc.Options.RpmsSources, rc.Options.UseBaseImageRpmRepos, distroHandler, rc.PackageSnapshotTime)
-	if err != nil {
-		return err
+	for _, configWithBase := range rc.ConfigChain {
+		err = addRemoveAndUpdatePackages(ctx, rc.BuildDirAbs, rc.BaseConfigPath, configWithBase.Config.OS,
+			imageChroot, nil, rc.Options.RpmsSources, rc.Options.UseBaseImageRpmRepos, distroHandler,
+			configWithBase.Config.OS.Packages.SnapshotTime)
```
cwize1: If you aren't going to use `rc.PackageSnapshotTime` here, then you probably need to refactor/delete the `validatePackageSnapshotTime` function. You also need to use `rc.Options.PackageSnapshotTime` somewhere.

This is likely the cause of your test failure.

toolkit/tools/pkg/imagecustomizerlib/baseconfigs.go
```diff
@@ -52,7 +54,8 @@ func buildConfigChainHelper(ctx context.Context, cfg *imagecustomizerapi.Config,
 		}
 
 		// Recurse into base config
-		subChain, err := buildConfigChainHelper(ctx, &baseCfg, absPath, visited, pathStack)
+		baseConfigDir := filepath.Dir(absPath)
```
cwize1: Good catch

toolkit/tools/pkg/imagecustomizerlib/customizeos.go
cwize1: `AddOrUpdateUsers` can also create groups. Specifically, when a user is created, a group of the same name is typically also automatically created.

I wonder if it makes sense to process groups and users of the same config together.

toolkit/tools/pkg/imagecustomizerlib/customizeos.go
```diff
@@ -30,47 +32,70 @@ func doOsCustomizations(ctx context.Context, rc *ResolvedConfig, imageConnection
...
+		snapshotTime := configWithBase.Config.OS.Packages.SnapshotTime
+		if rc.Options.PackageSnapshotTime != "" {
+			snapshotTime = rc.Options.PackageSnapshotTime
+		}
 
-	err = UpdateHostname(ctx, rc.Config.OS.Hostname, imageChroot)
-	if err != nil {
-		return err
+		if snapshotTime != "" {
+			if !slices.Contains(configWithBase.Config.PreviewFeatures, imagecustomizerapi.PreviewFeaturePackageSnapshotTime) {
```
cwize1: This is already checked in `ImageCustomizerOptions.verifyPreviewFeatures` and `Config.IsValid`. So, I don't think you need to duplicate the check here.

cwize1 (changes requested):

liulanze (approved):

PR #497, opened 2025-11-05
Image Customizer: Improve input/output (signed) artifacts API

liulanze:
When outputting artifacts and generating the inject-files configuration, make the source and destination file names identical, there should no longer be separate signed and unsigned files. Instead, users are expected to replace the unsigned file with its signed version. The unsignedSource field can be removed since it is no longer needed under this approach. Additionally, place both unsigned and signed files in subdirectories named after their corresponding OutputArtifactsItemType, making it easier for users to identify how each file should be signed. Finally, introduce a new type field that specifies the OutputArtifactsItemType for each artifact, serving the same purpose as the directory naming convention.

---

### **Checklist**
- [x] Tests added/updated


docs/imagecustomizer/api/configuration/injectArtifactMetadata.md
```diff
@@ -58,16 +58,30 @@ Added in v0.14.
 
 Required.
 
-Path to the signed artifact file to be injected. This path may be relative to the
+Path to the artifact file to be injected. This path may be relative to the
 `inject-files.yaml` config file or an absolute path.
 
+The file at this path should be initially the unsigned artifact extracted during
```
cwize1: This sounds like instructions for the user, instead of the behavior of the tool.

I would rephrase it to:

`When the config file is generated by the output artifacts API, this will be the path of the unsigned artifact. Users should replace this file with the signed version before running the inject-files command.`

docs/imagecustomizer/api/configuration/injectArtifactMetadata.md
```diff
@@ -58,16 +58,30 @@ Added in v0.14.
...
+
+Required.
 
-Optional.
+The type of artifact being injected. This field helps users identify which signing
+process to apply to each artifact.
 
-Path to the original unsigned artifact (if available). This field is for informational
-or auditing purposes only — it is not used during injection.
+Supported values:
```
cwize1: It might be good to provide a link to the `.output.artifacts.items` API, since the supported values are equivalent.

toolkit/tools/pkg/imagecustomizerlib/artifactsinputoutput.go
```diff
@@ -120,88 +120,115 @@ func outputArtifacts(ctx context.Context, items []imagecustomizerapi.OutputArtif
...
+		}
+
 		for _, verity := range verityMetadata {
 			if verity.hashSignaturePath == "" {
 				continue
 			}
 
-			unsignedHashFile := verity.name + ".hash"
-			destPath := filepath.Join(outputDir, unsignedHashFile)
+			// Use the exact destination filename (e.g., "usr.hash" if hashSignaturePath is "/boot/usr.hash").
```
cwize1: Is there a test that covers `HashSignaturePath`?

PR #502, opened 2025-11-13
Support UKI re-customization with enabling /boot cleanup.

liulanze:
This PR enables support for re-customizing UKI (Unified Kernel Image) images by implementing /boot directory cleanup and improving kernel cmdline extraction fallback logic. This also consider the case where kernel gets tdnf install/update/remove through image customizer.

---

### **Checklist**
- [x] Tests added/updated

### DIFF
--- a/docs/imagecustomizer/api/configuration/uki.md
+++ b/docs/imagecustomizer/api/configuration/uki.md
@@ -69,3 +69,91 @@ os:
 ```
 
 Added in v0.8.
+
+## reinitialize [string]
+
+Controls the behavior when customizing an image that already contains UKI files.
+
+optional.
+
+This field is relevant when customizing an existing image that already has UKI
+files in the `/boot/efi/EFI/Linux` directory. It determines whether to preserve
+the existing UKI files or regenerate them during the customization process.
+
+Supported options:
+
+- (unspecified): When `reinitialize` is not specified, the tool generates new
+  UKI files based on the kernel and initramfs files present in the image. The
+  `kernels` field must be specified to indicate which kernels to build UKIs for.
+  This is the default behavior for first-time UKI creation. Will show error when
+  detecting UKI from the base image.
+
+- `passthrough`: Preserve existing UKI files without modification. The kernel,
+  initramfs, and command-line arguments embedded in the existing UKIs are left
+  unchanged. When this option is specified, the `kernels` field must not be
+  specified.
+
+- `refresh`: Extract the kernel and initramfs from existing UKI files, then
+  regenerate new UKI files with updated configurations. The `kernels` field
+  must be specified to indicate which kernels to build UKIs for.
+
+Example (passthrough mode):
+
+```yaml
+# Customize an existing UKI image without regenerating UKI files.
+# This preserves the existing kernel, initramfs, and cmdline in the UKI.
+os:
+  uki:
+    reinitialize: passthrough
+  
+  # You can still perform OS customizations:
+  packages:
+    install:
+      - nginx
+      - vim
+  
+  additionalFiles:
+    - path: /etc/app-config.txt
+      content: |
+        Application configuration
+
+previewFeatures:
+  - uki
+```
+
+Example (refresh mode with verity):
+
+```yaml
+# Recustomize an existing UKI+verity image with updated verity hashes.
+# Extracts kernel/initramfs from existing UKIs and rebuilds them with
+# updated kernel command-line arguments (including new verity hashes).
+storage:
+  reinitializeVerity: all
+
+os:
+  bootloader:
+    resetType: hard-reset
+  
+  uki:
+    kernels: auto
+    reinitialize: refresh
+  
+  kernelCommandLine:
+    extraCommandLine:
+      - rd.info
+  
+  packages:
+    install:
+      - openssh-server
+  
+  additionalFiles:
+    - path: /etc/uki-recustomization.txt
+      content: |
+        UKI recustomization with verity refresh
+
+previewFeatures:
+  - uki
+  - reinitialize-verity
+```
+
+Added in v1.0.1
--- a/toolkit/tools/pkg/imagecustomizerlib/customizeselinux_test.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/customizeselinux_test.go
@@ -8,10 +8,12 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"strings"
 	"testing"
 
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/file"
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/imageconnection"
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/shell"
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/testutils"
 	"github.com/stretchr/testify/assert"
 )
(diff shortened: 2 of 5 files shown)


toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -747,6 +747,10 @@ func extractKernelCmdlineFromGrub(bootPartition diskutils.PartitionInfo, bootDir
 	grubCfgPath := filepath.Join(tmpDirBoot, bootDirPath, DefaultGrubCfgPath)
 	kernelToArgs, err := extractKernelCmdlineFromGrubFile(grubCfgPath)
 	if err != nil {
+		// Preserve os.IsNotExist check by not wrapping if file doesn't exist.
+		if os.IsNotExist(err) {
```
cwize1: Use `errors.Is(err, fs.ErrNotExist)` instead of `os.IsNotExist(err)`. Then you won't need to do this.

toolkit/tools/pkg/imagecustomizerlib/customizeos.go
```diff
@@ -30,6 +30,20 @@ func doOsCustomizations(ctx context.Context, rc *ResolvedConfig, imageConnection
 		return err
 	}
 
+	// If UKI is enabled, extract kernel and initramfs from existing UKIs for re-customization.
+	if rc.Config.OS.Uki != nil {
```
cwize1: What happens when the base image has UKIs but the user doesn't specify UKI in the config file?

cwize1:
Please add tests

toolkit/tools/pkg/imagecustomizerlib/customizeos.go
liulanze: 1. During validation phase, the `connectToExistingImage()` is trying to read kernel cmdline args from `grub.cfg` but failed, then falls back to extract from UKI, which is fine.
2. `ensureGrubCfgForUki()` won't be called, no grub.cfg available.
3. At package phase, if new Kernel installed / updated, won't impact the UKI / bootloader, in addition, the kernel remove will cause error.
4. Bootloader handling phase, it might regenerate grub.cfg anyway, but the grub.cfg and UKI will be in inconsistent state I think.

Maybe I need to detect base image has UKI but config doesn't specify it then fail with error. Thoughts?

cwize1: Take a look at the `TestCustomizeImageVerityUsrUki` test. The `verity-reinit-usr-nop.yaml` config currently relies on the config not specifying the UKI field as a way of ensuring the UKI is not updated/refreshed, which would invalidate any signature on the UKI. We need to ensure this scenario continues to work. But we can change the API if necessary.

I think it might make sense to require the user to explicitly tell us how to handle the base image's UKI.

Something like:

```yaml
uki:
  reinitialize: ["" | "passthrough" | "refresh"]
```

Where:

- `""`: Error if the base image has UKIs
- `"passthrough"`: Ensure the existing UKIs are used. Don't allow new UKIs.
- `"refresh"`: Delete the existing UKIs and regenerate them from scratch.

toolkit/tools/imagecustomizerapi/uki.go
```diff
@@ -8,13 +8,23 @@ import (
...
+	err := u.Reinitialize.IsValid()
 	if err != nil {
-		return fmt.Errorf("invalid uki kernels:\n%w", err)
+		return fmt.Errorf("invalid uki reinitialize:\n%w", err)
+	}
+
+	// When reinitialize is passthrough, kernels is optional (we preserve existing UKIs).
+	// For refresh mode or when reinitialize is not set, kernels must be specified.
+	if u.Reinitialize != UkiReinitializePassthrough {
+		err = u.Kernels.IsValid()
```
cwize1: `Kernels.IsValid()` has more checks than just whether or not it is specified.

toolkit/tools/pkg/imagecustomizerlib/bootcustomizer.go
```diff
@@ -88,13 +94,23 @@ func (b *BootCustomizer) getSELinuxModeFromGrub() (imagecustomizerapi.SELinuxMod
 
 	// Get the SELinux kernel command-line args.
 	if b.isGrubMkconfig {
+		// If grub config file is missing, indicate that with fs.ErrNotExist.
+		if b.defaultGrubFileContent == "" {
+			return imagecustomizerapi.SELinuxModeDefault, fs.ErrNotExist
```
cwize1: Let's stop using `fs.ErrNotExist` to represent when something is missing. It just causes problems. Add a `bool` to the return list instead.

toolkit/tools/pkg/imagecustomizerlib/bootcustomizer.go
```diff
@@ -114,16 +130,73 @@ func (b *BootCustomizer) getSELinuxModeFromGrub() (imagecustomizerapi.SELinuxMod
...
+	if err != nil {
+		return imagecustomizerapi.SELinuxModeDefault, err
+	}
+
+	if len(kernelToArgs) == 0 {
+		return imagecustomizerapi.SELinuxModeDefault, fs.ErrNotExist
+	}
+
+	// Use the first kernel's cmdline since they should all have the same SELinux settings.
+	var firstCmdline string
```
cwize1: De-duplicate code between `getSELinuxModeFromUki` and `extractKernelCmdlineFromUki`.

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
```diff
@@ -59,18 +65,67 @@ type UkiKernelInfo struct {
...
+	}
+
+	if !hasUkis {
+		// Base image doesn't have UKIs, no validation needed
+		return nil
+	}
+
+	// Base image has UKIs
+	if config.OS == nil || config.OS.Uki == nil {
+		return fmt.Errorf("base image contains UKI files but config does not specify os.uki field: " +
```
cwize1: We don't need two separate error messages for this. Just use the one below for both cases.

liulanze: Fixed.

toolkit/tools/pkg/imagecustomizerlib/bootcustomizer.go
```diff
@@ -61,14 +101,14 @@ func (b *BootCustomizer) AddKernelCommandLine(extraCommandLine []string) error {
 
 	combinedArgs := GrubArgsToString(extraCommandLine)
 
-	if b.isGrubMkconfig {
+	if b.bootConfigType == BootConfigTypeGrubMkconfig {
```
cwize1: Use `switch` statement.

Similar comment elsewhere.

toolkit/tools/pkg/imagecustomizerlib/bootcustomizer.go
```diff
@@ -16,41 +20,77 @@ var (
...
 }
 
+func determineBootConfigType(grubCfgContent string, imageChroot safechroot.ChrootInterface) BootConfigType {
+	// If grub.cfg doesn't exist, check for UKI
+	if grubCfgContent == "" {
+		hasUkis, err := baseImageHasUkis(imageChroot.(*safechroot.Chroot))
+		if err == nil && hasUkis {
+			return BootConfigTypeUki
+		}
+		// No grub.cfg and no UKIs - default to grub-legacy (edge case)
```
cwize1: Previously we would error out for this edge case. So, I would report an error here.

toolkit/tools/pkg/imagecustomizerlib/bootcustomizer.go
cwize1: Add a comment for the `BootConfigTypeUki` saying why nothing needs to be done.

Similar comment elsewhere.

toolkit/tools/pkg/imagecustomizerlib/bootcustomizer.go
```diff
@@ -81,54 +121,117 @@ func (b *BootCustomizer) AddKernelCommandLine(extraCommandLine []string) error {
...
 }
 
-// Gets the image's configured SELinux mode.
-func (b *BootCustomizer) getSELinuxModeFromGrub() (imagecustomizerapi.SELinuxMode, error) {
+// getSELinuxModeFromGrub gets the image's configured SELinux mode from grub configuration.
+// Returns (mode, found, error) where:
+//   - mode: The detected SELinux mode (or SELinuxModeDefault if not found)
+//   - found: true if grub config exists and was successfully parsed
+//   - error: Any error encountered during parsing
+func (b *BootCustomizer) getSELinuxModeFromGrub() (imagecustomizerapi.SELinuxMode, bool, error) {
```
cwize1: Rename `getSELinuxModeFromGrub` to getSELinuxModeFromCmdline` and then handle the UKI case in this function. This will simplify and de-duplicate the code (e.g. the call to `getSELinuxModeFromConfigFile`).

toolkit/tools/pkg/imagecustomizerlib/customizebootloader.go
```diff
@@ -30,12 +30,12 @@ var (
 	ErrBootloaderUnsupportedRootMountId = NewImageCustomizerError("Bootloader:UnsupportedRootMountId", "unsupported root mount identifier")
 )
 
-func handleBootLoader(ctx context.Context, rc *ResolvedConfig, imageConnection *imageconnection.ImageConnection,
+func handleBootLoader(ctx context.Context, buildDir string, rc *ResolvedConfig, imageConnection *imageconnection.ImageConnection,
```
cwize1: `buildDir` is available in `ResolvedConfig`.

toolkit/tools/pkg/imagecustomizerlib/defaultgrubutils.go
```diff
@@ -354,6 +356,10 @@ func UpdateDefaultGrubFileVariable(defaultGrubFileContent string, varName string
 func isGrubMkconfigEnabled(imageChroot *safechroot.Chroot) (bool, error) {
 	grub2ConfigFile, err := ReadGrub2ConfigFile(imageChroot)
 	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			// If grub.cfg doesn't exist (e.g., UKI-only image), grub-mkconfig is not enabled.
+			return false, nil
```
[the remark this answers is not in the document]
cwize1: `isGrubMkconfigEnabled` should probably be changed to return `BootConfigType`. And then `configureDiskBootLoader` should no-op on UKI.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -779,6 +791,7 @@ func extractKernelCmdlineFromGrub(bootPartition diskutils.PartitionInfo, bootDir
 	grubCfgPath := filepath.Join(tmpDirBoot, bootDirPath, DefaultGrubCfgPath)
 	kernelToArgs, err := extractKernelCmdlineFromGrubFile(grubCfgPath)
 	if err != nil {
+		_ = bootPartitionMount.CleanClose() // Synchronous cleanup on error
```
cwize1: `CleanClose` can fail. And when it does, the state of the disk is unknown.

Please change the code to return a `bool` when the grub config file is missing.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/customizeselinux_test.go
```diff
@@ -210,20 +212,77 @@ func verifyKernelCommandLine(t *testing.T, imageConnection *imageconnection.Imag
 	notExistsArgs []string,
 ) {
 	grubCfgFilePath := filepath.Join(imageConnection.Chroot().RootDir(), "/boot/grub2/grub.cfg")
+
+	// Check if grub.cfg exists (non-UKI image)
 	grubCfgContents, err := file.Read(grubCfgFilePath)
-	assert.NoError(t, err, "read grub.cfg file")
+	if err != nil {
```
cwize1: Don’t dynamically check things in test verification logic. The test knows if the image is UKI or not.

docs/imagecustomizer/api/configuration/uki.md
```diff
@@ -69,3 +69,91 @@ os:
...
+This field is relevant when customizing an existing image that already has UKI
+files in the `/boot/efi/EFI/Linux` directory. It determines whether to preserve
+the existing UKI files or regenerate them during the customization process.
+
+Supported options:
+
+- (unspecified): When `reinitialize` is not specified, the tool generates new
+  UKI files based on the kernel and initramfs files present in the image. The
+  `kernels` field must be specified to indicate which kernels to build UKIs for.
+  This is the default behavior for first-time UKI creation. Will show error when
```
cwize1: `An error will occur if the base image has UKIs`

toolkit/tools/pkg/imagecustomizerlib/customizebootloader.go
liulanze: Oh, nice, thanks.

docs/imagecustomizer/api/configuration/uki.md
```diff
@@ -69,3 +69,91 @@ os:
...
+optional.
+
+This field is relevant when customizing an existing image that already has UKI
+files in the `/boot/efi/EFI/Linux` directory. It determines whether to preserve
+the existing UKI files or regenerate them during the customization process.
+
+Supported options:
+
+- (unspecified): When `reinitialize` is not specified, the tool generates new
+  UKI files based on the kernel and initramfs files present in the image. The
```
vinceaperri: Could we have a specific default value for reinitialize? It would be easier to understand what the default behavior is rather than calling this (unspecified). See pxe.initramfsType config documentation for a good example of this format (string enum value).

vinceaperri (changes requested):

docs/imagecustomizer/api/configuration/uki.md
vinceaperri: Will this fail is the list of kernels is empty?

PR #504, opened 2025-11-18
Add hierarchical support for storage fields

elainezhao1:
---

### **Checklist**
- [ ] Tests added/updated


toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -524,7 +524,7 @@ func customizeOSContents(ctx context.Context, rc *ResolvedConfig) (imageMetadata
 		}
 	}
 
-	if len(rc.Config.Storage.Verity) > 0 || len(im.baseImageVerityMetadata) > 0 {
+	if len(rc.Storage.Verity) > 0 || len(im.baseImageVerityMetadata) > 0 {
```
cwize1: There are still a lot of references to `Config.Storage` in the codebase. These need to be redirected to `ResolvedConfig.Storage` instead. In VS Code, you can right click a field and select `Find All References`.

Also, you'll probably want to delete the `Config.CustomizePartitions()` function, since that contains a hidden reference to `Config.Storage`.

Also, the fact that you missed so much means that you need to expand your test coverage.

toolkit/tools/pkg/imagecustomizerlib/configvalidation.go
```diff
@@ -625,3 +662,32 @@ func resolveKernelCommandLine(configChain []*ConfigWithBasePath) imagecustomizer
 		ExtraCommandLine: mergedArgs,
 	}
 }
+
+func resolveStorage(configChain []*ConfigWithBasePath) imagecustomizerapi.Storage {
+	var resolvedStorage imagecustomizerapi.Storage
```
cwize1: You probably also need to think about `Storage.VerityPartitionsType`

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/baseconfigs_test.go
```diff
@@ -244,4 +243,79 @@ func TestBaseConfigsFullRun(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	options := ImageCustomizerOptions{
+		BuildDir: buildDir,
+	}
+
+	_, err = ValidateConfig(t.Context(), absBaseConfigPath, &config, false, options)
+
+	assert.Error(t, err)
+	assert.ErrorContains(t, err, "cannot specify 'resetPartitionsUuidsType'")
```
cwize1: It would be good to include the entire error message so that it is clearer why this failure is expected.

toolkit/tools/pkg/imagecustomizerlib/configvalidation.go
```diff
@@ -503,14 +517,44 @@ func validateIsoPxeCustomization(rc *ResolvedConfig) error {
...
+		storage := configWithBase.Config.Storage
+
+		hasDisks := len(storage.Disks) > 0
+		hasResetUUID := storage.ResetPartitionsUuidsType != imagecustomizerapi.ResetPartitionsUuidsTypeDefault
+		hasReinitVerity := storage.ReinitializeVerity != imagecustomizerapi.ReinitializeVerityTypeDefault
+
+		if baseHasDisks {
+			if hasResetUUID {
+				return fmt.Errorf(
+					"cannot specify 'resetPartitionsUuidsType' in config at layer %d when a base config specifies '.storage.disks'", i)
```
cwize1: I don't think `at layer %d` is particularly meaningful to the user, since they see the config tree and not the list. If possible, it might be nice to give the config file names. If not, I would just remove `at layer %d`.

toolkit/tools/pkg/imagecustomizerlib/configvalidation.go
```diff
@@ -625,3 +669,56 @@ func resolveKernelCommandLine(configChain []*ConfigWithBasePath) imagecustomizer
...
+			resolvedStorage.Disks = storage.Disks
+			resolvedStorage.BootType = storage.BootType
+			resolvedStorage.FileSystems = storage.FileSystems
+			resolvedStorage.Verity = storage.Verity
+			return resolvedStorage
+		}
+
+		// .storage.resetPartitionsUuidsType - override
+		if storage.ResetPartitionsUuidsType != imagecustomizerapi.ResetPartitionsUuidsTypeDefault {
+			resolvedStorage.ResetPartitionsUuidsType = storage.ResetPartitionsUuidsType
```
cwize1: If you are iterating backwards through the chain, you need to use early return. Otherwise, you will pick the first value in the chain instead of the last.

That being said, I think the storage merging logic is complex enough that it would probably be better to merge storage using recursion on the config tree. (The other fields have simple merging logic. So, going backwards through the linear chain works fine.)

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/configvalidation.go
```diff
@@ -503,15 +524,75 @@ func validateIsoPxeCustomization(rc *ResolvedConfig) error {
...
 
-func resolveOutputArtifacts(configChain []*ConfigWithBasePath) *imagecustomizerapi.Artifacts {
+func validateStorage(configChain []*ConfigWithPath) error {
+	var baseHasDisks bool
+
+	for _, configWithBase := range configChain {
+		storage := configWithBase.Config.Storage
+
+		hasDisks := storage.CustomizePartitions()
+		hasResetUUID := storage.ResetPartitionsUuidsType != imagecustomizerapi.ResetPartitionsUuidsTypeDefault
```
liulanze: nit: these local variables are redundant to have, can change it to something like:

```
if storage.ResetPartitionsUuidsType != imagecustomizerapi.ResetPartitionsUuidsTypeDefault {
```

cwize1:
PR is fairly out of date now due to #670 and #809

PR #516, opened 2025-12-11
Add support for baremetal images

fintelia:
This PR adds a new "baremetal-image" image type and implements the logic for creating COSI 1.2 files.

A baremetal image consists of a COSI 1.2 file with a VHD footer attached to it. This construction enables several benefits:

* Unlike a normal VHD, the file is compressed so it is considerably faster to transfer.
* The updated metadata in COSI 1.2 carries all the information need to reconstruct a partition table for the image
* Tooling and services which today operate on VHDs without inspecting their contents will seamlessly be able to operate on baremetal-images as well.

---

### **Checklist**
- [ ] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
@@ -601,9 +601,9 @@ func padToMegabyte(imageFile string) error {
 
 	padSize := int64(1024*1024) - (currentSize % int64(1024*1024))
 	if padSize != int64(1024*1024) {
-		_, err = cosiFile.Write(make([]byte, padSize))
+		err = cosiFile.Truncate(currentSize + padSize)
 		if err != nil {
-			return fmt.Errorf("failed to pad image file to 1 MB:\n%w", err)
+			return fmt.Errorf("failed to pad image file to 1 MiB:\n%w", err)
 		}
 	}
 


toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -132,6 +135,14 @@ func buildCosiFile(sourceDir string, outputFile string, partitions []outputParti
 			KnownInfo: partition,
 		}
 
+		outputPartitions = append(outputPartitions, Partition{
+			Path:         metadataImage.Image.Path,
+			PartUuid:     partition.PartUuid,
+			OriginalSize: partition.OriginalSize,
+			Label:        partition.PartLabel,
+			Number:       partition.PartitionNum + 1,
```
cwize1: `PartitionNum` is already 1-indexed. You don't need the `+ 1` here.

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -562,3 +586,26 @@ func handleBootloaderMetadata(bootloader *CosiBootloader) CosiBootloader {
...
+	defer cosiFile.Close()
+
+	currentSize, err := cosiFile.Seek(0, io.SeekEnd)
+	if err != nil {
+		return fmt.Errorf("failed to get current size of image file:\n%w", err)
+	}
+
+	padSize := int64(1024*1024) - (currentSize % int64(1024*1024))
+	if padSize != int64(1024*1024) {
+		_, err = cosiFile.Write(make([]byte, padSize))
```
cwize1: Use `cosiFile.Truncate` to expand the file size.

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -562,3 +586,26 @@ func handleBootloaderMetadata(bootloader *CosiBootloader) CosiBootloader {
...
+		if err != nil {
+			return fmt.Errorf("failed to pad image file to 1 MB:\n%w", err)
```
cwize1: `MiB`

`MB` means 10^6, not 2^20.

cwize1 (changes requested):

PR #517, opened 2025-12-12
Include COSI 1.2 metadata

fintelia:
This updates imagecustomizer to output [COSI 1.2](https://github.com/microsoft/azure-linux-image-tools/pull/516) files. 

~~Should not merge until the spec for COSI 1.2 is accepted.~~ COSI 1.2 is merged, so this'll be ready once the comment below is addressed.
 
---

### **Checklist**
- [ ] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
@@ -54,6 +54,7 @@ const (
 // Extract all partitions of connected image into separate files with specified format.
 func extractPartitions(imageLoopDevice string, outDir string, basename string, partitionFormat string,
 	imageUuid [randomization.UuidSize]byte, compressionLevel int, compressionLong int,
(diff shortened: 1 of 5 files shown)


toolkit/tools/pkg/imagecustomizerlib/extractpartitions.go
```diff
@@ -240,6 +241,7 @@ func constructOutputPartitionMetadata(diskPartition diskutils.PartitionInfo, par
 	partitionMetadata.Uuid = diskPartition.Uuid
 	partitionMetadata.PartUuid = diskPartition.PartUuid
 	partitionMetadata.Mountpoint = diskPartition.Mountpoint
+	partitionMetadata.OriginalSize = diskPartition.SizeInBytes
```
fintelia: This line is wrong. The partition has already been shrunk earlier during OS customization, but `originalSize` should hold the value from before partition shrinking.

PR #522, opened 2025-12-17
Fix mountbootpartition dracut module.

cwize1:
The mountbootpartition dracut module is used to mount the /boot partition during the initramfs, to allow the verity setup logic to access the verity hash signature file.

The way that the mountbootpartition dracut module is currently written is overly complex and cumbersome. But more importantly, there is a race condition where sometimes the /boot partition isn't mounted before the verity setup happens.

This change fixes this by handling the mount using a systemd `.mount` unit file, which is the proper way of mounting partitions during boot. This radically simplifies the logic and also allows the dependencies to be properly declared using `Before` and `After` in the unit file. In particular, a `Before=veritysetup-pre.target` is included, which fixes the race condition.

---

### **Checklist**

- [x] Tests added/updated


toolkit/tools/internal/resources/verity-signature/90mountbootpartition/mountbootpartition-generator.sh
```diff
@@ -1,81 +1,41 @@
...
-
-[Service]
-Type=oneshot
-ExecStart=$bootPartitionMonitorCmd
-RemainAfterExit=yes
-
-[Install]
-WantedBy=multi-user.target
+Description=/boot
+RequiresMountsFor=/dev/disk/by-uuid/$bootPartitionUuid
```
liulanze: Does `RequiresMountsFor=` have the same effect as the `After=dev-disk-by\\x2duuid-${bootPartitionUuid//-/\\x2d}.device` directive? They both take `partuuid`.

PR #694, opened 2026-04-13
Fix VMTests type annotations.

cwize1:
Python v3.12 doesn't support subscripts on user defined types. So, wrap the offending type declaration in a string.

---

### **Checklist**

- [x] Tests added/updated


liulanze (approved):
looks good

PR #698, opened 2026-04-14
Add Azure Linux 4.0 support

vinceaperri:
Add Azure Linux 4.0 (AZL4) support to the image customization tooling. The bulk of the change is a dedicated Azure Linux 4.0 distro handler plus a new Boot Loader Specification (BLS) parser/updater, because AZL4 differs from AZL3 in three fundamental ways: it uses `dnf` (not `tdnf`), it uses BLS entries (not `grub.cfg` `menuentry` blocks) as the source of truth for kernel command lines, and it ships per-arch `grub2-efi`/`shim` packages (not the `grub2-efi-binary[-noprefix]` umbrella names).

The PR also:

- Extends the `DistroHandler` interface so all bootloader/boot-config reads and writes are dispatched per-distro.
- Fixes a class of host-disk leakage when running `grub2-mkconfig` inside the image-customization chroot (`os-prober` is now force-disabled).
- Fixes a SELinux-relabel failure on images whose fstab mounts something at mnt (e.g. an Azure data-disk placeholder).
- Unskips 33 AZL4 tests that were previously stubbed out, and adds a regression test in `TestCustomizeImageNewUUIDs` for the `os-prober` leak.
- Adds an internal BLS Type 1 parser (`toolkit/tools/internal/bls`) and higher-level read/update helpers with extensive table-driven tests.
- Routes verity and UKI kernel-args rewrites through the distro handler so AZL4/Fedora update BLS entry files while AZL2/3/Ubuntu keep rewriting `grub.cfg`.
- Swaps `dnf info --installed` for `rpm -q` in `isPackageInstalled` to work on read-only chroots without writing a logfile inside the chroot.
- Fixes the Fedora handler: it now reads BLS by default (Fedora 42+) and stops pairing AZL3 grub-default with the AZL4 EFI stub.
- Fixes the ACL handler: grub/BLS methods now consistently return `fs.ErrNotExist` instead of mixed zero values that mislead callers.
- LiveISO "missing shim/grub" errors now name the correct per-distro/per-arch packages (`shim-x64`, `grub2-efi-x64` on AZL4 amd64).
- Backports the upstream systemd v258 [`ukify` section-padding fix](https://github.com/systemd/systemd/commit/ec1d031f3de02f84beca89e2b402d085fba62be4) onto the AZL3 functional-test host's `ukify` so the `TestCustomizeImageVerityUsrUkiRecustomize` passes for AZL4.
- Warns when `systemd-boot-unsigned` is detected on AZL4 (image will fail Secure Boot until a signed `systemd-boot` ships).
- Rejects ISO/PXE output formats in AZL4's `ValidateConfig` (not yet supported).

Notes:

- Package snapshot time (`--package-snapshot-time`, `os.packages.snapshotTime`) is intentionally skipped for AZL4 because the feature is a Microsoft addition to AZL3's `tdnf` that has not been ported to `dnf`. Tests `TestCustomizeImagePackagesSnapshotTime`, `TestCustomizeImagePackagesCliSnapshotTimeOverridesConfigFile`, and `TestCustomizeImagePackagesSnapshotTimeWithoutPreviewFlagFails` remain skipped for AZL4.
- ISO and PXE-* customization support is not implemented in this PR. It will be in a follow-up PR. For now, `test_min_change_efi_azl4_iso_bootstrap_output`, `test_min_change_efi_azl4_iso_full_os_output`, and `test_min_change_legacy_azl4_iso_output` remain skipped for AZL4.
- `systemd-boot-unsigned` will be removed once `systemd-boot` is available for AZL4. The code currently recognizes both for forwards-compatibility.
- The ukify patch will be upstreamed to Azure Linux 3.0, after which the workaround in the test workflows will be removed.
- Azure Linux 4.0's main repository only contains one kernel, so `TestCustomizeImageMultiKernel` remains skipped.
- Azure Linux 4.0's alpha release is shipped with unsigned RPMs right now, but when it's officially released the RPMs will be signed. Until then, `TestCustomizeImagePackagesAddOfflineLocalRepoNoGpgKey` remains skipped.
- The workarounds in the AZL4 test data to remove grub2-efi-x64/aa64 won't be necessary after AZL4 upgrades to shim 16.1+ (Fedora 44+).
- `cloud-native-azl4.repo` (or other repos) do not yet exist for AZL4, so `TestCustomizeImagePackagesUrlSource` remains skipped.


toolkit/tools/pkg/imagecustomizerlib/baseconfigs_test.go
```diff
@@ -76,7 +76,7 @@ func TestBaseConfigsFullRun(t *testing.T) {
 	}
 
 	if runtime.GOARCH == "arm64" {
```
cwize1: Support was added for ARM64 systemd-boot in August 2025. So, we can and should remove these checks. But it might make sense to do this in a different change.

toolkit/tools/pkg/imagecustomizerlib/customizeoverlays_test.go
```diff
@@ -91,8 +91,8 @@ func TestCustomizeImageOverlaysSELinux(t *testing.T) {
 }
 
 func testCustomizeImageOverlaysSELinuxHelper(t *testing.T, testName string, baseImageInfo testBaseImageInfo) {
-	if baseImageInfo.Version == baseImageVersionAzl3 {
-		t.Skip("Azure Linux 3.0 is missing policy.kern file")
+	if baseImageInfo.Version == baseImageVersionAzl3 || baseImageInfo.Version == baseImageVersionAzl4 {
```
cwize1: I wonder if this has been fixed now. It looks like `selinux-policy` does now contain a `policy.kern` file.

toolkit/tools/pkg/imagecustomizerlib/customizepackages_test.go
```diff
@@ -456,7 +456,7 @@ func TestCustomizeImagePackagesUrlSource(t *testing.T) {
 }
 
 func TestCustomizeImagePackagesBadRepo(t *testing.T) {
-	baseImageInfo := testBaseImageAzl3CoreEfi
+	baseImageInfo := testBaseImageAzl4CoreEfi
```
cwize1: Why was this changed to AZL4? Should the test be run for both versions?

Similar question elsewhere.

cwize1 (changes requested):

liulanze:
Please add a PR description for future reference.

test/vmtests/vmtests/imagecustomizer/test_min_change.py
```diff
@@ -157,6 +157,9 @@ def run_basic_checks(
         elif input_image_azl_release == 3:
             assert "ID=azurelinux" in os_release_text
             assert 'VERSION_ID="3.0"' in os_release_text
+        elif input_image_azl_release == 4:
+            assert "ID=azurelinux" in os_release_text
+            assert "VERSION_ID=4.0" in os_release_text
```
liulanze, line 162: nit: `'VERSION_ID="4.0"'` for consistency as above.

docs/imagecustomizer/developer-guide.md
```diff
@@ -41,6 +41,7 @@ sudo go test -C ./toolkit/tools ./...
 
    - [Azure Linux 2.0 core-efi](https://github.com/microsoft/azurelinux/blob/2.0/toolkit/imageconfigs/core-efi.json)
    - [Azure Linux 3.0 core-efi](https://github.com/microsoft/azurelinux/blob/3.0/toolkit/imageconfigs/core-efi.json)
+   - Azure Linux 4.0 core-efi
```
liulanze, line 44: Missing the hyperlink here?

test/vmtests/vmtests/imagecustomizer/test_min_change.py
vinceaperri: Actually, Azure Linux 4.0 does not quote the version ID. We'd have to push this upstream to get it changed.

docs/imagecustomizer/developer-guide.md
vinceaperri: Without an official release, I hesitate to add a URL now. I will track adding this as a task.

toolkit/tools/pkg/imagecustomizerlib/customizepackages_test.go
cwize1: I mainly don't want our tests to be running against AZL4 exclusively until it is officially released.

toolkit/tools/pkg/imagecustomizerlib/artifactsinputoutput_test.go
```diff
@@ -22,266 +23,274 @@ const (
...
-		true /*useBaseImageRpmRepos*/, "" /*packageSnapshotTime*/)
-	if !assert.NoError(t, err) {
-		return
-	}
-
-	espFiles := verifyAndSignOutputtedArtifacts(t, outputArtifactsDir, false)
-
-	// Use new buildDir to ensure the buildDir is created if it doesn't exist.
-	buildDirInject := filepath.Join(buildDir, "inject")
+	for _, baseImageInfo := range checkSkipForCustomizeDefaultAzureLinuxImages(t) {
```
cwize1: Please follow the pattern used in the other tests, where there is a `*Helper` function containing most of the logic of the test. This will reduce the size of the diff and will reduce the tab depth.

Similar comment elsewhere.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/main_test.go
```diff
@@ -262,35 +285,58 @@ func checkSkipForCustomizeImage(t *testing.T, baseImage testBaseImageInfo) strin
 	return *baseImage.Param
 }
 
-func findFirstAvailableImage(priorityList []testBaseImageInfo) (testBaseImageInfo, bool) {
+// findFirstAvailableImages returns at most two images: the highest priority image
+// from each of the two largest major Azure Linux versions provided.
+func findFirstAvailableImages(priorityList []testBaseImageInfo) []testBaseImageInfo {
```
cwize1: If you are going to run most tests against both AZL3 and AZL4, would it make sense to split them into separate GitHub jobs to parallelize them?

cwize1: FYI: This logic was originally added because some features had identical logic between AZL2 and AZL3. So, it was seen as a waste of time to run the test against both versions. But maybe it would make sense to just run all tests against all supported versions.

toolkit/tools/pkg/imagecustomizerlib/main_test.go
```diff
@@ -191,6 +213,7 @@ var (
 var (
 	baseImageCoreEfiAzl2          = flag.String(paramBaseImageAzl2CoreEfi, "", "An Azure Linux 2.0 core-efi image to use as a base image.")
 	baseImageCoreEfiAzl3          = flag.String(paramBaseImageAzl3CoreEfi, "", "An Azure Linux 3.0 core-efi image to use as a base image.")
+	baseImageCoreEfiAzl4          = flag.String(paramBaseImageAzl4CoreEfi, "", "An Azure Linux 4.0 core-efi image to use as a base image.")
 	baseImageBareMetalAzl2        = flag.String(paramBaseImageAzl2BareMetal, "", "An Azure Linux 2.0 bare-metal image to use as a base image.")
```
cwize1, line 215: I don't think `--base-image-bare-metal-azl2` and `--base-image-bare-metal-azl3` are being used anymore.

I added these when I thought I would need to download the images from MCR. But then I decided to use an Azure storage account to cache images.

vinceaperri: I can remove in another PR

toolkit/tools/pkg/imagecustomizerlib/baseconfigs_test.go
vinceaperri: Will do.

toolkit/tools/pkg/imagecustomizerlib/main_test.go
vinceaperri: That's a cleaner implementation to add test coverage. Thanks.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_fedora.go
```diff
@@ -127,3 +130,18 @@ func (d *fedoraDistroHandler) ConfigureDiskBootLoader(imageConnection *imageconn
 	return configureDiskBootLoader(imageConnection, rootMountIdType, bootType, selinuxConfig, kernelCommandLine,
 		currentSELinuxMode, true /* forceGrubMkconfig */)
 }
+
+func (d *fedoraDistroHandler) ExtractKernelToArgs(espDir string, bootDir string, buildDir string) (map[string]string, error) {
+	grubCfgPath := filepath.Join(bootDir, DefaultGrubCfgPath)
+	kernelToArgs, err := extractKernelToArgsFromGrub(grubCfgPath)
```
liulanze: The function `extractKernelToArgsFromGrub()` already joins the path, the line above `grubCfgPath` should be removed. Same comment for `distrohandler_ubuntu.go`

toolkit/tools/pkg/imagecustomizerlib/distrohandler_azurelinux.go
liulanze: Is `azl4` installs `systemd-boot-unsigned` instead?

toolkit/tools/imagegen/diskutils/filesystem.go
```diff
@@ -119,6 +126,17 @@ var (
 		},
 	}
 
+	// The default ext4 options used by an Azure Linux 4.0 image.
+	// See, the /etc/mke2fs.conf file in an Azure Linux 4.0 imag.
+	azl4Ext4Options = ext4Options{
+		BlockSize: 4096,
+		Features: []string{
+			"sparse_super", "large_file", "filetype", "resize_inode", "dir_index", "ext_attr", "has_journal", "extent",
+			"huge_file", "flex_bg", "metadata_csum", "metadata_csum_seed", "64bit", "dir_nlink", "extra_isize",
```
liulanze, line 137: Added `metadata_csum_seed` you need to make sure `e2fsprogs >= 1.47.0`, consider adding it to `ext4FeaturesE2fsprogsSupport = map[string]version.Version{ ......`

toolkit/tools/imagegen/diskutils/filesystem.go
```diff
@@ -119,6 +126,17 @@ var (
 		},
 	}
 
+	// The default ext4 options used by an Azure Linux 4.0 image.
+	// See, the /etc/mke2fs.conf file in an Azure Linux 4.0 imag.
```
liulanze: nit - image.

toolkit/tools/pkg/imagecustomizerlib/customizeoverlays_test.go
```diff
@@ -85,19 +85,19 @@ func TestCustomizeImageOverlays(t *testing.T) {
...
-			testCustomizeImageOverlaysSELinuxHelper(t, "TestCustomizeImageOverlaysSELinux"+baseImageInfo.Name, baseImageInfo)
+			testCustomizeImageOverlaysSELinuxHelper(t, baseImageInfo)
 		})
 	}
 }
 
-func testCustomizeImageOverlaysSELinuxHelper(t *testing.T, testName string, baseImageInfo testBaseImageInfo) {
+func testCustomizeImageOverlaysSELinuxHelper(t *testing.T, baseImageInfo testBaseImageInfo) {
 	if baseImageInfo.Version == baseImageVersionAzl3 {
 		t.Skip("Azure Linux 3.0 is missing policy.kern file")
```
liulanze: Let's extend the skip logic for azl4.

liulanze (changes requested):

toolkit/tools/pkg/imagecustomizerlib/distrohandler_fedora.go
vinceaperri: What line should be removed?

toolkit/tools/pkg/imagecustomizerlib/distrohandler_azurelinux.go
vinceaperri: Yes, unfortunately. Setting up production signing is usually the last thing that happens before GA.

toolkit/tools/pkg/imagecustomizerlib/customizeoverlays_test.go
vinceaperri: We're actually able to remove this skip logic now.

toolkit/tools/imagegen/diskutils/filesystem.go
vinceaperri: Thanks! This is definitely a bug.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_fedora.go
liulanze: Sorry, the `grubCfgPath := filepath.Join(bootDir, DefaultGrubCfgPath)` line, you will see the path has joint already in `extractKernelToArgsFromGrub()`, so no need to join the path again. Same issue for `distrohandler_ubuntu.go`

liulanze: Looks like this is fixed. Looks good, thanks. Please resolve the PR check pipeline failures.

liulanze (dismissed):
Please make sure the pipeline `Build (dev)` passed if you haven't done so.

liulanze (dismissed):

Britel:
Hi Vince — I've been validating Trident's downstream usage of this PR while we work on AZL4 enablement on our side, and have three pieces of feedback.

**TL;DR:** The dnf `logdir` fix in this PR is **correct and necessary** — it unblocks our entire AZL4 test image pipeline. But this PR doesn't compile in isolation, and the related stacked branches haven't picked up the fix yet.

---

### ✅ The dnf logdir fix is correct

The change at `packagemanager_dnf.go:193`:

```go
err := shell.NewExecBuilder("dnf", "info", "--setopt=logdir=/tmp", "--installed", packageName).
```

directly fixes a hard failure during MIC bootloader detection on AZL4 base images. Without it, any `customize` config that defines a `storage:` block smaller than the source image triggers MIC's partition resize → remount read-only → `dnf info --installed grub2-efi-x64` → `Cannot open log file: (30) - Read-only file system [/var/log/dnf5.log]` → bootloader detection returns false → fatal `unknown bootloader: none of [grub2-efi-x64 grub2-efi-aa64] or [systemd-boot systemd-boot-unsigned] found`.

AZL3 didn't hit this because tdnf doesn't write a log file by default. AZL4's switch to dnf5 surfaced it.

**I confirmed end-to-end** by patching this same one-line fix into the `multi-distro-grub` branch and rebuilding the MIC container. With the fix, Trident's `tests/images/testimages.py build trident-vm-grub-testimage-azl4` completes cleanly against the `core-efi-vhdx-4.0-amd64/4.0.20260420` base:

```
INFO[0053] Finished building COSI: ...trident-vm-grub-testimage-azl4.cosi
INFO[0060] Success!
```

Without the fix the same invocation fails with the read-only log file error described above.

### 🔴 PR #698 doesn't compile in isolation

Building from this branch HEAD (`6f90482e`):

```
$ cd toolkit/tools/imagecustomizer && go build .
# github.com/microsoft/azure-linux-image-tools/toolkit/tools/pkg/imagecustomizerlib
../pkg/imagecustomizerlib/partitionutils.go:1120:22: undefined: installutils
```

The call site at `partitionutils.go:1120` is:

```go
grubCfgFile, err := installutils.FindGrubCfgFile(bootDir)
```

but `partitionutils.go` doesn't import `installutils`. The function exists at `toolkit/tools/imagegen/installutils/installutils.go:133`, just isn't being pulled in.

The CI on this branch shows [Build AMD64 (5/7) FAILED](https://github.com/microsoft/azure-linux-image-tools/actions/runs/25522489734/job/74909908462) which I'm guessing is for the same reason.

Quick fix: add `"github.com/microsoft/azure-linux-image-tools/toolkit/tools/imagegen/installutils"` to the imports block in `partitionutils.go`.

### 🟡 Stacked branches missing the fix

Both `user/vinceaperri/multi-distro-grub` (#732) and `user/vinceaperri/azl4-tests` (#731) are missing the dnf logdir fix from this PR — `packagemanager_dnf.go` on those branches still has the unpatched call. They'll hit the same hard failure on any AZL4 customize with a partition resize.

Per `gh api .../compare/azl4-support...azl4-tests` that branch is "behind by 97" relative to azl4-support, so a rebase is overdue. Same story for multi-distro-grub.

Suggested order:
1. Land #698 first (with the import fix above)
2. Rebase #732 on the new main; it picks up the dnf fix transitively
3. Rebase #731 on the result

### 🟡 Lower-priority observation: blkid ambiguous on `--output-image-format raw`

Tangentially: with the dnf fix applied, `--output-image-format cosi` works end-to-end. But `--output-image-format raw` against the same AZL4 `core-efi-vhdx-4.0` base + same customize config fails earlier in the flow:

```
INFO[0000] Creating raw base image: /tmp/image.raw
FATA[0006] cannot validate target OS of the base image:
   failed to mount raw disk (/tmp/image.raw) as a loopback device:
   failed to get filesystem type of partition (/dev/loop0p1)
   exit status 2
```

`/dev/loop0p1` is the ESP (FAT32, 64 MiB). blkid exit 2 = ambiguous result. This may be a partition-table-vs-blkid race during initial loopback attach, or it may be specific to this base image. Not blocking for our path (we use `cosi`), but worth a look since you're touching this area heavily.

Happy to share full trace logs / repro environment for any of the above. Thanks for the AZL4 work — it's unblocking a chunk of Trident's path forward.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -351,8 +353,120 @@ func readFstabEntriesFromRootfs(rootfsPartition *diskutils.PartitionInfo,
...
+	}
+
+	// Always return an os.ErrNotExist if detection couldn't find an os-release file.
+	if err == nil {
+		err = os.ErrNotExist
+	}
+	return nil, fmt.Errorf("failed to find os-release file:\n%w", err)
+}
+
+// mountFstabPartitionReadonly mounts the partition referenced by fstab's mountTarget entry under
```
liulanze: Duplicated docstring, see line 440 and 443

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -64,11 +65,19 @@ const (
 	// SELinuxConfigDisabled is the string value to set SELinux to disabled in the /etc/selinux/config file.
 	SELinuxConfigDisabled = "disabled"
 
+	// FedoraGrubCfgRelPath is the path to the grub config file on Fedora and Azure Linux systems,
+	// relative to the /boot directory.
+	FedoraGrubCfgRelPath = "grub2/grub.cfg"
```
liulanze, line 70: Can we re-use the constants under customizeuki?

toolkit/tools/pkg/imagecustomizerlib/distrohandler_fedora.go
```diff
@@ -144,3 +145,29 @@ func (d *fedoraDistroHandler) ConfigureDiskBootLoader(imageConnection *imageconn
 		currentSELinuxMode, true /* forceGrubMkconfig */, d, resources.AssetsGrubDefFileAzl3,
 		installutils.FedoraGrubEnvRelPath, resources.AssetsGrubStubFileAzl4, installutils.GrubStubDirsAzl4)
```
Britel, line 179: 🐻‍❄️ **Karhu deep-review** (🔴 Blocking, High confidence)

**Fedora handler mixes AZL3 grub-default with AZL4 EFI stub.** Passes `AssetsGrubDefFileAzl3` for `/etc/default/grub` but `AssetsGrubStubFileAzl4` and `GrubStubDirsAzl4` for the EFI stub. The Azure Linux handler keeps both consistent (`distrohandler_azurelinux.go:188-198`).

The AZL3 default-file's templated placeholders (notably `{{.CryptoMountCommand}}`) are not what the AZL4-shaped stub expects. Combined with the AZL4 stub also lacking `{{.CryptoMountCommand}}`, this means encrypted-volume support silently no-ops on Fedora.

Commit `021a0765 fix: fedora missing efibootmgr and initial-setup (regression)` suggests Fedora regressed at some point during the PR — worth checking whether this is the same root cause.

**Fix:** introduce Fedora-specific assets (`AssetsGrubDefFileFedora`, `AssetsGrubStubFileFedora`), or pair both to AZL4 with a justifying comment if Fedora 42's BLS layout genuinely matches AZL4.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_fedora.go
```diff
@@ -144,3 +145,29 @@ func (d *fedoraDistroHandler) ConfigureDiskBootLoader(imageConnection *imageconn
...
 }
+
+func (d *fedoraDistroHandler) ReadGrubConfigLinuxArgs(bootDir string) (map[string][]grubConfigLinuxArg, error) {
+	return readKernelCmdlinesFromGrubCfg(bootDir, FedoraGrubCfgPath)
```
Britel: 🐻‍❄️ **Karhu deep-review** (🔴 Blocking, Medium confidence)

**Fedora `ReadGrubConfigLinuxArgs` will fail on Fedora 42 (BLS by default).** This reads inline `linux` lines from `grub.cfg`. Fedora 30+ defaults to `GRUB_ENABLE_BLSCFG=true` and emits no inline `linux` lines.

After commit `3187f31d` made `readKernelCmdlinesFromGrubCfg` strictly require at least one non-recovery `linux` line (`customizeuki.go:714`), Fedora images will hit:
```
failed to find any non-recovery linux command lines in grub.cfg
```

Same applies to `ReadNonRecoveryKernelCmdlines` and `UpdateBootConfigForVerity` below.

**Fix:** branch Fedora on BLS like AZL4 does — call `readKernelCmdlinesFromBLSEntries` directly, or detect via `FindBlsCfg` (`grubcfgutils.go:187`).

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
```diff
@@ -115,3 +116,29 @@ func (d *aclDistroHandler) ConfigureDiskBootLoader(imageConnection *imageconnect
 ) error {
 	return fmt.Errorf("bootloader configuration is not yet supported for ACL")
 }
+
+func (d *aclDistroHandler) ReadGrubConfigLinuxArgs(bootDir string) (map[string][]grubConfigLinuxArg, error) {
+	return readKernelCmdlinesFromGrubCfg(bootDir, FedoraGrubCfgPath)
```
Britel: 🐻‍❄️ **Karhu deep-review** (🔴 Blocking, Medium confidence)

**ACL handler returns grub paths despite ACL being grub-less.** `ReadGrub2ConfigFile` and `WriteGrub2ConfigFile` (lines 96-106) explicitly state ACL doesn't use GRUB, but `ReadGrubConfigLinuxArgs` (here), `ReadKernelCmdlines`, and `UpdateBootConfigForVerity` (line 138) unconditionally chase `FedoraGrubCfgPath`.

Verity/UKI flows are gated off via `ValidateConfig`, but read-only flows (extract-partitions, COSI metadata via `partitionutils.go:1084`) traverse this path. Result: missing grub.cfg on ACL → wrapped `ErrNotExist` → caller silently treats as 'no boot config' → kernel cmdline lost.

**Fix:** return `fs.ErrNotExist` (or an explicit 'not implemented for ACL' error) from these methods up-front, or implement BLS reading since ACL is BLS/UKI-based.

Same concern at `FindBootPartitionUuidFromEsp` (line 83).

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
```diff
@@ -692,14 +694,178 @@ func extractKernelToArgs(espPath string, bootDir string, buildDir string) (map[s
...
+
+			switch key {
+			case "linux":
+				if linux != "" {
+					return nil, fmt.Errorf("duplicate linux key in BLS entry (%s)", absPath)
+				}
+				linux = filepath.Base(value)
+			case "title":
+				title = value
+			case "efi", "uki", "uki-url":
```
Britel: 🐻‍❄️ **Karhu deep-review** (🔴 Blocking, High confidence — flagged by all 3 skeptic models)

**BLS parser hard-rejects valid `efi`/`uki`/`uki-url` keys, blocking UKI fallback.** The parser returns `fmt.Errorf("BLS entry uses 'uki' key, which is not supported")`. The wrapper `extractKernelToArgs` (~line 668) only falls through to `extractKernelCmdlineFromUkiEfis` on `fs.ErrNotExist` — so any AZL4 image already booting via UKI/BLS will fail re-customization, even though UKI extraction would have worked.

Native UKI BLS entries (Type-2, valid per the Boot Loader Spec) are also blocked.

**Fix:** either skip entries with `efi`/`uki`/`uki-url` keys and continue parsing other entries, or wrap the unsupported-key error so `extractKernelToArgs` falls through to UKI extraction. Returning `fs.ErrNotExist` when *only* UKI-shaped entries exist works cleanly with the existing fallback.

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -228,10 +236,130 @@ func installVerityMountBootPartitionDracutModule(installRoot string) error {
...
+	for idx, lineIndex := range optionsLineIndices {
+		trimmed := strings.TrimSpace(lines[lineIndex])
+		_, value, _ := strings.Cut(trimmed, " ")
+		if value == "" {
+			_, value, _ = strings.Cut(trimmed, "\t")
+		}
+
+		options := strings.TrimSpace(value)
+		var filteredTokens []string
+		for _, token := range strings.Fields(options) {
```
Britel: 🐻‍❄️ **Karhu deep-review** (🔴 Blocking, High confidence — flagged by all 3 skeptic models)

**`updateBLSEntryOptions` mishandles whitespace/quoting — read/write asymmetry corrupts BLS round-trips.** Three independent issues:

1. **Quoting:** `strings.Fields(options)` splits on whitespace only. The reader (`readKernelCmdlinesFromBLSEntries`) uses `grub.TokenizeConfig` which handles quoting. A value like `rd.cmdline="foo bar"` gets split into two tokens on write.

2. **Tab/space detection** (lines 303-305, 317-319): `strings.Cut(trimmed, " ")` then fallback `\t`. A line like `"options\tfoo=bar"` produces `key="options\tfoo=bar"` (no space match), the `options` line is silently skipped, and a new one is appended. Multiple `options` lines per spec means stale verity args coexist with new ones.

3. **Trailing newline:** `strings.Split`/`strings.Join` on `"\n"` may add or drop the trailing newline depending on input. Most tools tolerate it; signed BLS entries (UKI+pcrphase scenarios when SB lands) won't.

**Fix:** unify on `grub.TokenizeConfig` / `ParseCommandLineArgs` for both read and write. Use `strings.IndexFunc(trimmed, unicode.IsSpace)` for line-key detection. Preserve trailing newline.

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -228,10 +236,130 @@ func installVerityMountBootPartitionDracutModule(installRoot string) error {
...
+			continue
+		}
+
+		absPath := filepath.Join(entriesDir, entry.Name())
+		content, err := os.ReadFile(absPath)
+		if err != nil {
+			return fmt.Errorf("failed to read BLS entry file (%s):\n%w", absPath, err)
+		}
+
+		updatedContent, err := updateBLSEntryOptions(string(content), argsToRemove, newArgs)
```
Britel: 🐻‍❄️ **Karhu deep-review** (🔴 Blocking, Medium confidence)

**Verity BLS update touches recovery/rescue entries.** `updateBLSEntriesForVerity` iterates every `.conf` file (line 240) and calls `updateBLSEntryOptions` here with no recovery filter. Recovery entries get `root=/dev/mapper/root` and verity args injected. The recovery initramfs may not contain verity dracut modules, breaking the rescue path — defeating its purpose.

Recovery-skip logic exists in the read path (`customizeuki.go:790`) but not here.

**Fix:** apply the same recovery filter before mutation. Match both `recovery` and `rescue`, case-insensitively (the read-path filter is also too narrow — see separate comment at customizeuki.go:790).

toolkit/tools/pkg/osmodifierlib/modifydefaultgrub.go
```diff
@@ -68,38 +68,19 @@ func modifyDefaultGrub() error {
...
-				if arg.Name == "root" {
-					rootDevice = arg.Value
-				} else {
-					values = append(values, arg.Name+"="+arg.Value)
-				}
-			}
-		}
+	rootDevice := argMap["root"]
+	delete(argMap, "root")
+	for name, value := range argMap {
```
Britel: 🐻‍❄️ **Karhu deep-review** (🟡 Non-blocking, Medium confidence)

**Non-deterministic kernel arg ordering — reproducibility regression.** `argMap` from `ReadNonRecoveryKernelCmdlines` is a `map[string]string`. Go map iteration is randomized per spec. Each osmodifier run produces a different key order in `GRUB_CMDLINE_LINUX`; `grub2-mkconfig` reproduces that into the resulting `grub.cfg`, making images non-reproducible across osmodifier invocations.

The pre-PR code iterated `argTokens` (deterministic grub.cfg order). Also: the previous `len(lines) != 1` safety check that errored on multi-kernel images is gone — the new code silently picks whichever non-recovery entry lands first.

**Fix:** sort `values` alphabetically before passing to `UpdateKernelCommandLineArgs`, or change `ReadNonRecoveryKernelCmdlines` to return `[]kernelArg` preserving source order.

toolkit/tools/pkg/imagecustomizerlib/packagemanager_dnf.go
```diff
@@ -34,6 +34,13 @@ func (pm *dnfPackageManager) getCacheOnlyOptions() []string {
 	return []string{"--setopt=cacheonly=metadata"}
 }
 
+// getRefreshMetadataOptions returns DNF-specific options for the metadata refresh command.
+func (pm *dnfPackageManager) getRefreshMetadataOptions() []string {
```
Britel: 🐻‍❄️ **Karhu deep-review** (🟡 Non-blocking, High confidence — flagged by 3 models)

**`getRefreshMetadataOptions` is dead code — abstraction is bypassed.** This method (and the TDNF equivalent at `packagemanager_tdnf.go:39`, plus the interface declaration at `packagemanager_rpm.go:23`) has zero call sites. The actual `--setopt=skip_if_unavailable=False` injection is hardcoded in `customizepackages_rpm.go:285` (which the PR doesn't change).

A future engineer implementing a new `rpmPackageManagerHandler` will satisfy the interface and assume the option is wired through — but the real injection lives in `customizepackages_rpm.go`. Easy way to ship a missing or double-applied flag.

**Fix:** either wire `handler.getRefreshMetadataOptions()` into `refreshRpmPackageMetadata` and remove the hardcoded string in `customizepackages_rpm.go:285`, or remove the dead interface method.

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
```diff
@@ -692,14 +694,178 @@ func extractKernelToArgs(espPath string, bootDir string, buildDir string) (map[s
...
+				options += value
+			}
+		}
+
+		if linux == "" {
+			return nil, fmt.Errorf("BLS entry (%s) is missing 'linux' key", absPath)
+		}
+
+		if title == "" {
+			return nil, fmt.Errorf("BLS entry (%s) is missing 'title' key", absPath)
```
Britel: 🐻‍❄️ **Karhu deep-review** (🟡 Non-blocking, Medium confidence)

**`title` is now unconditionally required — regression vs. pre-refactor behavior.** Before the unification in `3187f31d`, the equivalent of `extractKernelCmdlineFromBLSEntries(skipRecovery=false)` only required `title` when `skipRecovery=true`. The new code rejects entries missing `title` for both code paths, even though the BLS spec allows it (systemd-boot falls back to filename).

Any AZL4 / Fedora image with a BLS entry missing `title:` fails the entire kernel-args read with no fallback. Affects extract / COSI / verity flows.

**Fix:** when `title == ""`, treat the entry as non-recovery (don't error), or skip with a debug log.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -575,7 +605,12 @@ func installGrubTemplateFile(assetFile, targetFile, installRoot, rootDevice, boo
 func CallGrubMkconfig(installChroot safechroot.ChrootInterface) (err error) {
 	ReportActionf("Running %s...", FedoraGrubMkconfigBinary)
 
-	return shell.NewExecBuilder(FedoraGrubMkconfigBinary, "-o", FedoraGrubCfgFile).
+	// Force-disable os-prober. grub2-mkconfig is run inside an image-customization
+	// chroot that has the host's /dev bind-mounted; without this, /etc/grub.d/30_os-prober
+	// would enumerate the build host's disks and inject menuentries pointing at the
+	// host's kernels and partitions into the customized image's grub.cfg. Set via env
+	// so the protection applies regardless of /etc/default/grub state in the target image.
+	return shell.NewExecBuilder("env", "GRUB_DISABLE_OS_PROBER=true", FedoraGrubMkconfigBinary, "-o", FedoraGrubCfgFile).
```
Britel: 🐻‍❄️ **Karhu deep-review** (🟡 Non-blocking, Medium confidence)

**`os-prober` disable uses `env` binary wrapper instead of process env.** This runs `env GRUB_DISABLE_OS_PROBER=true grub2-mkconfig …` inside the chroot. If the target image has `coreutils-single` swapped or `/usr/bin/env` symlinked exotically, the wrapper itself fails — and this affects all images, not just AZL4.

_The fix itself is correct and overdue_ (host kernel-entry leakage into customer images was a real bug). This is just an implementation refinement.

**Fix:** set the env var on the `shell.ExecBuilder` directly via an `.Env()` call instead of wrapping with `/usr/bin/env`. If the builder doesn't expose env-vars, add that capability — useful elsewhere too.

_(Worth calling out in release notes: this changes grub.cfg generation for ALL distros, which is the right thing but operators may notice different output.)_

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
```diff
@@ -692,14 +694,178 @@ func extractKernelToArgs(espPath string, bootDir string, buildDir string) (map[s
...
+		}
+
+		if strings.Contains(title, "recovery") {
```
Britel: 🐻‍❄️ **Karhu deep-review** (🟡 Non-blocking, Medium confidence)

**BLS recovery filter is substring-only and case-sensitive.** `strings.Contains(title, "recovery")` misses `Rescue`, `Recovery`, distro-specific rescue naming. Combined with the missing recovery filter on the write path (see `customizeverity.go:275`), recovery entries can be both included in reads AND mutated in writes.

**Fix:** lowercase compare, match both `recovery` and `rescue`. Even better: extract a shared helper used by both the read path here and the verity update path.

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
```diff
@@ -345,22 +347,20 @@ func prepareUkiHelper(ctx context.Context, buildDir string, uki *imagecustomizer
...
-			return fmt.Errorf("package (%s) is not installed:\n"+
-				"the following packages must be installed to use Uki: (%v)", pkg, requiredRpms)
+		if distroHandler.IsPackageInstalled(imageChroot, pkg) {
+			return nil
 		}
 	}
 
-	return nil
+	return fmt.Errorf("package (%s) is not installed:\n"+
+		"this package must be installed to use Uki", systemdBootPackages[0])
```
Britel: 🐻‍❄️ **Karhu deep-review** (🟡 Non-blocking, Medium confidence)

**Error message misnames the AZL4 package.** Always reads `"package (systemd-boot) is not installed"` because it hardcodes `systemdBootPackages[0]`. On AZL4 the satisfying package is `systemd-boot-unsigned`; the error directs the operator to install the wrong one.

**Fix:**
```go
return fmt.Errorf("package (%s) is not installed:\n"+
    "this package must be installed to use Uki", strings.Join(systemdBootPackages, " or "))
```

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -991,22 +1112,45 @@ func extractKernelCmdlineFromGrub(bootPartition diskutils.PartitionInfo, bootDir
...
+	return nil, false, fmt.Errorf("no kernel args found in boot config")
 }
 
-// Extracts the kernel args for each kernel from the grub.cfg file.
-// Returns a mapping from kernel version to list of kernel args.
-func extractKernelCmdlineFromGrubFile(grubCfgPath string) (map[string][]grubConfigLinuxArg, error) {
-	grubCfgContent, err := file.Read(grubCfgPath)
+// readGrubConfigLinuxArgsBestEffort probes a boot directory for a kernel command-line source
+// when no DistroHandler is available (e.g. os-release was not detectable on a USR-verity image).
+func readGrubConfigLinuxArgsBestEffort(bootDir string) (map[string][]grubConfigLinuxArg, error) {
```
Britel, line 1128: 🐻‍❄️ **Karhu deep-review** (🟡 Non-blocking, Medium confidence)

**`readGrubConfigLinuxArgsBestEffort` doesn't try BLS.** When `detectDistroFromRootfs` fails (e.g. verity image without os-release on rootfs), `distroHandler` is nil and this best-effort path probes only `grub2/grub.cfg` / `grub/grub.cfg`. AZL4 verity images (BLS-only, no `/boot/grub2/grub.cfg`) get `fs.ErrNotExist` and surface a confusing error rather than attempting BLS extraction.

Also: hybrid BLS + inline-`linux` grub.cfg files (legitimate on Fedora during BLS migration windows) are hard-rejected at line 1138-ish. Could prefer BLS with a debug log.

**Fix:** also probe `boot/loader/entries/*.conf` if grub.cfg probes all miss. Tolerate hybrid grub.cfg by preferring BLS.

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
```diff
@@ -345,22 +347,20 @@ func prepareUkiHelper(ctx context.Context, buildDir string, uki *imagecustomizer
 }
 
 func validateUkiDependencies(imageChroot *safechroot.Chroot, distroHandler DistroHandler) error {
-	// The following packages are required for the UKI feature:
-	// - "systemd-boot": Checked as a package dependency here to ensure installation,
-	//    but additional configuration is handled elsewhere in the UKI workflow.
-	requiredRpms := []string{"systemd-boot"}
+	systemdBootPackages := []string{"systemd-boot"}
+	if distroHandler.GetTargetOs() == targetos.TargetOsAzureLinux4 {
+		systemdBootPackages = []string{"systemd-boot", "systemd-boot-unsigned"}
```
Britel: 🐻‍❄️ **Karhu deep-review** (🟡 Non-blocking, Low confidence)

**UKI dependency check accepts `systemd-boot-unsigned` with no Secure Boot guard.** `systemdBootPackages = {"systemd-boot", "systemd-boot-unsigned"}` for AZL4; first match wins. An image with only `systemd-boot-unsigned` (which will fail Secure Boot verification on a SB-enabled VM) silently passes UKI validation.

Acknowledged in the PR description as a workaround pending signed `systemd-boot` for AZL4, but there's no assertion that prevents `unsigned` from leaking to a Secure Boot configuration.

**Fix:** either warn when only `unsigned` is present, or hard-error in `ValidateConfig` if `secureBoot` is requested AND only `systemd-boot-unsigned` would be installed.

Britel:
🐻‍❄️ **Karhu deep-review** (🔴 Blocking, Medium confidence — verified)

**`toolkit/tools/internal/resources/assets/azurelinux/azurelinux-4.0/efi/grub/grub.cfg` is missing the `{{.CryptoMountCommand}}` placeholder → encrypted boot silently no-ops on AZL4.**

_Posting as a general comment because GitHub won't accept a line comment on this brand-new file._

The AZL3 stub at `azurelinux-3.0/efi/grub/grub.cfg:1` starts with `{{.CryptoMountCommand}}`. The new AZL4 stub does not. `setGrubCfgEncryptedVolume` (`installutils.go:1319`) does a sed-style replacement of the placeholder; with no placeholder present, it matches zero occurrences and still returns `nil`.

Result: an AZL4 image build that enables encrypted root/boot has the encryption request silently dropped at this layer. Image boots, fails to find root, no build-time warning. Same bug reaches Fedora via the asset mismatch flagged at `distrohandler_fedora.go:146`.

**Fix:** pick one:
  - Add `{{.CryptoMountCommand}}` to this stub (matches AZL3 behavior)
  - Make `setGrubCfgEncryptedVolume` error when the placeholder is absent but `enableEncryptedVolume=true`
  - Gate AZL4 encrypted root behind an explicit `ValidateConfig` 'not yet supported' error

Britel:
🐻‍❄️ **Karhu deep-review** — suggestions (architectural / non-merge-blocking)

These are direction notes from the architect role across all three models. Independent of the line-level findings — none of these block merge, but the first cluster is worth landing before AZL5 work begins.

### Architecture (post-merge, before AZL5)

1. **Split `azureLinuxDistroHandler` per major version** (`azureLinux2DistroHandler`, `azureLinux3DistroHandler`, `azureLinux4DistroHandler`). Currently 8-13 raw `if d.version == "4.0"` branches inside one struct (lines 31, 57, 88, 97, 115, 129, 153, 182, 187, 206, 225, 239 in `distrohandler_azurelinux.go`). AZL2 vs AZL3 vs AZL4 differ in initramfs tool, package manager, _and_ boot-config format — three of the biggest axes the interface abstracts. ACL got its own file for less. Doing the split with three concrete versions in hand will produce a much better interface than waiting for AZL5 to force it.

2. **Pull orchestrator `GetTargetOs() == AzureLinux4` checks into interface methods.** `customizeuki.go:351` duplicates handler knowledge from `distrohandler_azurelinux.go:97`. SSoT violation. Add e.g. `RequiredSystemdBootPackages()` to the interface and delete the orchestrator branch.

3. **Introduce a typed `DistroCapabilities` / `DistroProfile` struct** to replace stringly-typed `ValidateConfig` rejections (e.g. ISO/PXE blocked via `fmt.Errorf` at `distrohandler_azurelinux.go:60`, ACL has 4 similar rejections at `distrohandler_acl.go:39-53`). Captures unsupported features as data, not prose — greppable and removable.

4. **Eliminate `ReadKernelCmdlines` copy-paste.** Four byte-for-byte identical 5-line implementations across `distrohandler_acl.go:3914`, `distrohandler_azurelinux.go:3994`, `distrohandler_fedora.go:4054`, `distrohandler_ubuntu.go:4097`. Either remove from interface and call `grubKernelArgsToStringMap(d.ReadGrubConfigLinuxArgs(...))` at callsites, or use an embedded base handler.

5. **Rename `ReadGrubConfigLinuxArgs` → `ReadBootConfigLinuxArgs`.** Method name says "Grub" but AZL4 implementation reads BLS entries from `/boot/loader/entries/*.conf`. Confusing for future readers tracing the call.

6. **Standardize `targetos` version match policy.** AZL exact-matches `"4.0"`, ACL prefix-matches `"3.*"`. An AZL `4.0.x` patch line would fail detection; ACL `3.0.20260421` works. Either prefix-match both with a documented "major.minor pins the handler" contract, or exact-match both.

### CI & Test architecture

7. **Matrix-ify `build.yml`.** 6 hand-rolled `{azl3, azl4} × {azl3-host, ubuntu2404-host} × {amd64, arm64}` job blocks should be a `strategy.matrix` with an `include` list. Will double again at AZL5.

8. **Test YAML composition.** `verity-*-{amd64,arm64}-{azl3,azl4}.yaml` proliferation (~16+ files for the verity matrix alone) will compound at AZL5. Either generate programmatically (Go struct → YAML) or use base + arch + distro overlay composition.

### Workaround hygiene (cheap, high leverage)

9. **Annotate every AZL4 workaround** with a removal trigger:
   - `systemd-boot-unsigned` package quirk
   - `skip_if_unavailable=False` for DNF
   - `unzip → jq` test substitution
   - `ukify-fix-insertion-of-padding-in-merged-sections.patch` CI patch
   - shim 16.1+ workaround for `grub2-efi-x64/aa64`
   - ISO/PXE blocked for AZL4 in `ValidateConfig`

   Format: `// AZL4 WORKAROUND: <why>. Remove when <condition>. Tracking: <issue link>.`

   A workaround you cannot find is a workaround you cannot remove.

### Pre-merge polish

- PR description says "switching from unzip to wget in tests" but the actual final state (after `f29a0192`) is `unzip → jq`. Description is stale.
- `mountFstabPartitionReadonly` in `partitionutils.go` has a duplicated docstring (merge artifact).
- The global `GRUB_DISABLE_OS_PROBER=true` change in `installutils.go:CallGrubMkconfig` silently changes behavior for ALL existing AZL2/AZL3/Fedora images. It's a correctness fix (host disks should never be in a guest grub.cfg), but worth calling out in release notes so operators don't file a regression when they notice different grub.cfg output.
- Decide whether `targetos.GetInstalledTargetOs` should treat unknown `VARIANT_ID` on Azure Linux as standard AZL or fail closed (`targetos.go:54-77`).

### Skipped tests with no "unskip when X" trigger

Not bugs, but worth tracking explicitly:
- `TestCustomizeImagePackagesAddOfflineLocalRepoWithGpgKey` — "AZL4 RPMs not yet GPG signed"
- `TestCustomizeImageMultiKernel` for AZL4 — "only one kernel version available"
- SELinux removal test for AZL4 — "selinux-policy ships by default"
- ARM64 UKI tests skipped with AZL3-specific message but condition is version-agnostic — should narrow to AZL3 only since AZL4 ships `systemd-boot-unsigned` on arm64 (e.g. `baseconfigs_test.go:79`).

### Strengths everyone agreed on

- DistroHandler interface expansion is the right seam.
- The `os-prober` env fix is real correctness — host kernel-entry leakage was a latent bug affecting all images.
- `detectDistroFromRootfs` cascading `/usr` and `/usr/lib` fallback is genuinely thoughtful for USR-verity images.
- `/run/_bindmountroot` improvement over `/mnt/_bindmountroot` collision is a clean win, with good comments.
- dnf vs tdnf abstraction in `rpmPackageManagerHandler` is clean.
- Tests are not shallow — `verifyVerityGrub` rewritten to call `distroHandler.ReadGrubConfigLinuxArgs` instead of regex parsing eliminates a test-only parser.
- 100-commit history is CI-driven calibration, not algorithmic churn.

_Per-agent reports retained locally — happy to share specific findings if useful._

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
vinceaperri: To reach this error, /boot/loader/entries/ has to contain a .conf with efi=/uki=/uki-url= at the moment the parser runs. None of imagecustomizer's flows produce such an entry:

- The grub-based AZL4 path lets grub-mkconfig + kernel-install manage /loader/entries/, which produces linux= entries the parser handles correctly.
- The UKI path (os.uki.mode: create) installs UKIs into /EFI/Linux/ and relies on systemd-boot's auto-discovery there. It never writes .conf BLS entries with efi=/uki=/uki-url=.

The AZL4 base images we customize don't ship such entries either. They ship entires with linux= only. The only ways to hit the error are

(a) hand-injecting an efi=/uki=/uki-url= entriy via additionalFiles even though systemd-boot already auto-discovers UKIs in /EFI/Linux/ and doesn't need the entry, or
(b) feeding in an image produced outside imagecustomizer that already contains one.

I don't think either is a scenario we need to support, so this error scenario remains valid.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
vinceaperri: We will throw an error in these functions. This is an improvement from the AZL3 code path ACL currently takes on main.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_fedora.go
vinceaperri: Will align with AZL4 in the Fedora distroHandler, which is an improvement from Fedora using the AZL3 code path on main.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_fedora.go
vinceaperri: Definitely an inconsistency. I've aligned to AZL4, which is an improvement from using AZL3 for Fedora on main.

toolkit/tools/pkg/imagecustomizerlib/packagemanager_dnf.go
vinceaperri: Removed dead code.

toolkit/tools/imagegen/installutils/installutils.go
vinceaperri: I will split this off into another PR for the sake of visibility. Right now, I've taken your suggestion and removed the "env" layering.

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
vinceaperri: Added warning.

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
vinceaperri: Fixed.

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
vinceaperri: Fixed. This was an existing weakness that was just extended into BLS-land. I've improved both the BLS and existing non-BLS code to make a case-insensitive compare against both recovery and rescue.

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
vinceaperri: fixed

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
vinceaperri: fixed

vinceaperri:
> 🐻‍❄️ **Karhu deep-review** (🔴 Blocking, Medium confidence — verified)
> 
> **`toolkit/tools/internal/resources/assets/azurelinux/azurelinux-4.0/efi/grub/grub.cfg` is missing the `{{.CryptoMountCommand}}` placeholder → encrypted boot silently no-ops on AZL4.**
> 
> _Posting as a general comment because GitHub won't accept a line comment on this brand-new file._
> 
> The AZL3 stub at `azurelinux-3.0/efi/grub/grub.cfg:1` starts with `{{.CryptoMountCommand}}`. The new AZL4 stub does not. `setGrubCfgEncryptedVolume` (`installutils.go:1319`) does a sed-style replacement of the placeholder; with no placeholder present, it matches zero occurrences and still returns `nil`.
> 
> Result: an AZL4 image build that enables encrypted root/boot has the encryption request silently dropped at this layer. Image boots, fails to find root, no build-time warning. Same bug reaches Fedora via the asset mismatch flagged at `distrohandler_fedora.go:146`.
> 
> **Fix:** pick one:
> 
> * Add `{{.CryptoMountCommand}}` to this stub (matches AZL3 behavior)
> * Make `setGrubCfgEncryptedVolume` error when the placeholder is absent but `enableEncryptedVolume=true`
> * Gate AZL4 encrypted root behind an explicit `ValidateConfig` 'not yet supported' error

This isn't blocking. CryptoMountCommand is never used by imagecustomizer.

toolkit/tools/pkg/osmodifierlib/modifydefaultgrub.go
vinceaperri: Fixed

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
vinceaperri: It does try BLS, just not when grub.cfg is missing. I've fixed the function to also try BLS when grub.cfg is missing.

.github/workflows/tests-functional.yml
```diff
@@ -68,6 +68,14 @@ jobs:
       env:
         HOST_ARCH: ${{ inputs.hostArch }}
 
+    - name: Patch ukify (AZL3)
+      if: inputs.hostDistro == 'azl3'
+      run: |
+        set -eux
+
+        UKIFY_PATH=$(which ukify)
+        sudo patch "$UKIFY_PATH" ./repo/.github/workflows/scripts/ukify-fix-insertion-of-padding-in-merged-sections.patch
```
cwize1, line 77: Is this patch being taken into AZL3?

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
```diff
@@ -692,14 +703,224 @@ func extractKernelToArgs(espPath string, bootDir string, buildDir string) (map[s
...
+	linuxLines := FindNonRecoveryLinuxLines(grubLines)
+	if len(linuxLines) == 0 {
+		return nil, fmt.Errorf("failed to find any non-recovery linux command lines in grub.cfg")
+	}
+
+	return formatKernelCmdlineFromLinuxLines(linuxLines)
+}
+
+// parseBLSLinuxValue parses the value following a "linux" key in a BLS entry.
+func parseBLSLinuxValue(lineTokens []grub.Token) (string, error) {
```
cwize1: Move all the BLS helper functions into their own file. e.g. `blsutils.go`.

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
```diff
@@ -692,14 +703,224 @@ func extractKernelToArgs(espPath string, bootDir string, buildDir string) (map[s
...
+			sb.WriteString(sw.Value)
+		}
+		valueParts = append(valueParts, sb.String())
+	}
+	return strings.Join(valueParts, " "), nil
+}
+
+// readKernelCmdlinesFromBLSEntries reads Boot Loader Specification (BLS) entries in {bootDir}/loader/entries/*.conf,
+// extracting a kernel-to-cmdline mapping for non-recovery entries.
+func readKernelCmdlinesFromBLSEntries(bootDir string) (map[string][]grubConfigLinuxArg, error) {
```
cwize1: BLS is a file that specifies a bunch of key-value pairs. It would be good to create a generic function that parses a BLS file into key-value pairs and then we can add other functions on top that interpret the BLS values.

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
```diff
@@ -692,14 +703,224 @@ func extractKernelToArgs(espPath string, bootDir string, buildDir string) (map[s
...
+			continue
+		}
+
+		absPath := filepath.Join(entriesDir, entry.Name())
+		content, err := os.ReadFile(absPath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read BLS entry file (%s):\n%w", absPath, err)
+		}
+
+		// BLS entries are grub-syntax key/value pairs, so we can leverage the grub parsing code here.
```
cwize1: It might be worth looking at the systemd's parsing logic to double check if this is the correct way of parsing BLS files.

For example, how does systemd's logic handle the `$ character?

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
```diff
@@ -344,23 +345,31 @@ func prepareUkiHelper(ctx context.Context, buildDir string, uki *imagecustomizer
...
+	systemdBootPackages []string,
+) error {
+	for _, pkg := range systemdBootPackages {
 		logger.Log.Debugf("Checking if package (%s) is installed", pkg)
-		installed := distroHandler.IsPackageInstalled(imageChroot, pkg)
-		if !installed {
-			return fmt.Errorf("package (%s) is not installed:\n"+
-				"the following packages must be installed to use Uki: (%v)", pkg, requiredRpms)
+		if distroHandler.IsPackageInstalled(imageChroot, pkg) {
+			if pkg == systemdBootUnsignedPackage {
```
cwize1: The value of `systemdBootUnsignedPackage` seems distro specific.

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
```diff
@@ -692,14 +703,224 @@ func extractKernelToArgs(espPath string, bootDir string, buildDir string) (map[s
...
+				args = append(args, lineArgs...)
+			}
+		}
+
+		if linux == "" {
+			return nil, fmt.Errorf("BLS entry (%s) is missing 'linux' key", absPath)
+		}
+
+		// Entries without titles are treated as normal entries.
+		if isRecoveryOrRescueTitle(title) {
```
cwize1: Would this be distro specific?

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
```diff
@@ -228,10 +237,136 @@ func installVerityMountBootPartitionDracutModule(installRoot string) error {
...
+//     on whitespace.
+//   - Per BLS spec an entry may contain multiple "options" lines whose values are
+//     concatenated. argsToRemove is applied to every options line; newArgs is appended
+//     only to the last one (so re-runs converge instead of duplicating).
+//   - Each existing options line is replaced byte-for-byte at the source location of its
+//     tokens, so surrounding lines, comments, indentation and the trailing newline are
+//     preserved exactly.
+//   - If no options line exists, a new one is appended, matching the file's existing
+//     trailing-newline convention.
+func updateBLSEntryOptions(content string, argsToRemove []string, newArgs []string) (string, error) {
```
cwize1: Considering that BLS is a declarative file format, it would probably be better to have a BLS parser/writer library instead of doing string manipulation like we have to do for grub.cfg files.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_azurelinux.go
```diff
@@ -83,25 +85,23 @@ func (d *azureLinuxDistroHandler) GetAllPackagesFromChroot(imageChroot safechroo
...
-	systemdBootPackages := []string{"systemd-boot"}
-	if d.version == "4.0" {
-		systemdBootPackages = []string{"systemd-boot", "systemd-boot-unsigned"}
-	}
-	for _, pkg := range systemdBootPackages {
-		if d.IsPackageInstalled(imageChroot, pkg) {
-			return BootloaderTypeSystemdBoot, nil
+	var grubEfiPackages []string
+	switch d.version {
+	case "4.0":
```
cwize1: It seems like almost every function has to special case AZL4. It might be good to create a dedicated DistroHandler for it.

toolkit/tools/pkg/imagecustomizerlib/distrohandler.go
```diff
@@ -17,6 +18,13 @@ const (
 	packageManagerTDNF = "tdnf"
 	packageManagerDNF  = "dnf"
 	packageManagerAPT  = "apt-get"
+
+	grubEfiPackageFedoraAmd64  = "grub2-efi-x64"
```
cwize1: Place these variables in their corresponding distrohandler file?

toolkit/tools/pkg/imagecustomizerlib/liveosisoartifactstore.go
```diff
@@ -367,7 +367,7 @@ func createIsoFilesStoreFromMountedImage(inputArtifactsStore *IsoArtifactsStore,
 
 	if filesStore.grubEfiPath == "" {
 		return nil, fmt.Errorf("failed to find the grub efi file (%s or %s):\n"+
-			"this file is provided by either the (grub2-efi-binary) or the (grub2-efi-binary-noprefix) package",
+			"this file is provided by the grub2-efi package (e.g. grub2-efi-binary, grub2-efi-x64, or grub2-efi-aa64)",
```
cwize1: It might be nice to make the package names in the error message distro specific.

toolkit/tools/pkg/imagecustomizerlib/packagemanager_dnf.go
```diff
@@ -181,7 +181,9 @@ func (pm *dnfPackageManager) createOutputCallback() func(string) {
 }
 
 func (pm *dnfPackageManager) isPackageInstalled(imageChroot safechroot.ChrootInterface, packageName string) bool {
-	err := shell.NewExecBuilder("dnf", "info", "--installed", packageName).
+	// dnf unconditionally opens /var/log/dnf.log for writing during info operations, so set logdir to a location we
+	// know is always writable, even in read-only chroots.
+	err := shell.NewExecBuilder("dnf", "info", "--setopt=logdir=/tmp", "--installed", packageName).
```
cwize1: Writing to a fixed location under /tmp is generally discouraged due to security concerns. (Another program can pre-claim the file and mess with it.)

toolkit/tools/pkg/imagecustomizerlib/imageutils.go
```diff
@@ -80,7 +82,15 @@ func connectToExistingImageHelper(imageConnection *imageconnection.ImageConnecti
 		return nil, nil, nil, fmt.Errorf("failed to read fstab entries from rootfs partition:\n%w", err)
 	}
 
-	partitionsLayout, verityMetadata, err := discoverPartitionLayout(fstabEntries, partitions, buildDir, ignoreOverlays)
+	if distroHandler == nil {
+		distroHandler, err = detectDistroFromRootfs(buildDir, rootfsPartition, rootfsPath, partitions, fstabEntries)
```
cwize1: I would move the `detectDistroFromRootfs` call into `getKernelCmdline` in `discoverPartitionLayout`. That way, the logic won't run unless it is needed.

Adding extra logic into a path used by all distros, increases the possibility that a new error will occur for a distro, when in previous IC versions that error wasn't seen.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -351,8 +353,120 @@ func readFstabEntriesFromRootfs(rootfsPartition *diskutils.PartitionInfo,
...
+	defer func() {
+		for i := len(fallbackMounts) - 1; i >= 0; i-- {
+			fallbackMounts[i].Close()
+		}
+	}()
+	for _, fstabTarget := range []struct {
+		mountTarget string
+		mountSubdir string
+	}{
+		{mountTarget: "/usr", mountSubdir: "usr"},
```
cwize1: Define the list on its own line.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -351,8 +353,120 @@ func readFstabEntriesFromRootfs(rootfsPartition *diskutils.PartitionInfo,
...
+	}
+	if !errors.Is(err, fs.ErrNotExist) {
+		return nil, fmt.Errorf("failed to get installed target OS from rootfs partition:\n%w", err)
+	}
+
+	// Cascading fallbacks: try /usr from fstab, then /usr/lib from fstab.
+	var fallbackMounts []*safemount.Mount
+	defer func() {
+		for i := len(fallbackMounts) - 1; i >= 0; i-- {
+			fallbackMounts[i].Close()
```
cwize1: `defer` is scoped to the function, not the `{}`. You can use `defer` in a loop and it'll likely do the right thing.

Also, `safemount.Mount` will correctly handle it if you call `Close` after `CleanClose`.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -351,8 +353,120 @@ func readFstabEntriesFromRootfs(rootfsPartition *diskutils.PartitionInfo,
 	return fstabEntries, nil
 }
 
+// detectDistroFromRootfs mounts the rootfs partition (and /usr or /usr/lib from fstab if needed)
+// read-only and returns a DistroHandler derived from /etc/os-release (or /usr/lib/os-release).
+// Returns nil if the os-release file cannot be found (e.g. USR-verity configured images).
+func detectDistroFromRootfs(buildDir string, rootfsPartition *diskutils.PartitionInfo, rootfsPath string,
```
cwize1, line 365: I think you might be able to simplify this function by using `getMostSpecificPath` to find which fstab entries have the `/etc` and `/usr/lib` directories.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -961,23 +1075,30 @@ func extractKernelCmdlineFromGrub(bootPartition diskutils.PartitionInfo, bootDir
...
 	defer bootPartitionMount.Close()
 
-	grubCfgPath := filepath.Join(tmpDirBoot, bootDirPath, DefaultGrubCfgPath)
-	kernelToArgs, err := extractKernelCmdlineFromGrubFile(grubCfgPath)
+	var kernelToArgs map[string][]grubConfigLinuxArg
+	bootDir := filepath.Join(tmpDirBoot, bootDirPath)
+	if distroHandler != nil {
+		kernelToArgs, err = distroHandler.ReadGrubConfigLinuxArgs(bootDir)
+	} else {
+		// Best-effort attempt to read the kernel args if no distro handler is available.
```
cwize1: Under what scenario does this occur?

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/customizeverity.go
vinceaperri: I've intentionally made the BLS path consistent with the grub path (see updateGrubConfigForVerity()). If we want to change IC behavior here, could we file it as a bug and address this in a separate PR?)

.github/workflows/tests-functional.yml
vinceaperri: I plan on proposing this, yes.

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
vinceaperri: This comment is wrong. systemd just treats them as line-based key-values errors. This will simplify the parsing, although we'll still need to parse the "options" values in a special way.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
vinceaperri: USR-verity on Fedora-like distros. The OS-detection mechanism in detectDistroFromRootfs needs os-release, but on Fedora this file is under /usr, and when it's under verity, it can't find the partition using fstab.

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
vinceaperri: It's BLS/GRUB specific, yes. Updated.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
vinceaperri: Updated comment.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_fedora.go
```diff
@@ -141,6 +157,50 @@ func (d *fedoraDistroHandler) ConfigureDiskBootLoader(imageConnection *imageconn
...
+	if err != nil {
+		return nil, err
+	}
+
+	return grubKernelArgsToStringMap(kernelToArgs), nil
+}
+
+func (d *fedoraDistroHandler) ReadNonRecoveryKernelCmdlines(bootDir string, argNames []string) (map[string]string, error) {
+	grubCfgPath := filepath.Join(bootDir, installutils.FedoraGrubCfgRelPath)
+	return readNonRecoveryKernelCmdlinesFromGrubCfg(grubCfgPath, argNames)
```
liulanze: This may not directly relate to Azl4. Fedora 42 seems does not have non-recovery `linux` lines, so this `readNonRecoveryKernelCmdlinesFromGrubCfg` will fail and then `updateGrubConfigForVerity` will silently fail to inject the verity args.

toolkit/tools/pkg/imagecustomizerlib/packagemanager_dnf.go
```diff
@@ -181,7 +181,10 @@ func (pm *dnfPackageManager) createOutputCallback() func(string) {
 }
 
 func (pm *dnfPackageManager) isPackageInstalled(imageChroot safechroot.ChrootInterface, packageName string) bool {
-	err := shell.NewExecBuilder("dnf", "info", "--installed", packageName).
```
cwize1, line 184: I think I recall that the code was originally using `rpm` and was switched to `tdnf` for some reason. It might be worth finding out why `tdnf` was preferred over `rpm`.

I think `rpm` might not be installed on all SKUs.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/packagemanager_dnf.go
vinceaperri: I've looked into this and found George's commit: 

```
commit 4f1cbac7537913b70d2fe1b25efa4dfc5e8639f1
Author: George Mileka <gmileka@users.noreply.github.com>
Date:   Tue Dec 3 18:42:06 2024 -0800

    Remove dependency on the 'rpm' package. (#12)
    
    MIC ISO flow uses 'rpm' to query information about installed packages.
    However, some input images do not have 'rpm' installed - only 'rpmlib'.
    
    MIC uses tdnf to install packages - so, it is safe to assume that tdnf
    is a required package for MIC operation.
    
    This fix switches from using 'rpm' to using 'tdnf' so no new
    dependencies are required for images without 'rpm' installed.
```

I've confirmed that on azl3 tdnf requires rpm-libs, but rpm-libs does **not** require rpm.

However, on azl4 dnf5 requires rpm-libs, which **does** require rpm.

So, as long as the coupling between dnf5 and rpm is maintained on azl4 through this dependency chain, it will be safe to use rpm as a drop-in replacement for dnf5.

Does this address your concern, or should we look more into how to facilitate dnf5 operations on read-only filesystems?

toolkit/tools/pkg/imagecustomizerlib/distrohandler_fedora.go
vinceaperri: Thanks. Since Fedora isn't even fully supported, I didn't think to align it with Azure Linux 4.0. However, there is no good reason not to after the azl4 code is stable.

toolkit/tools/pkg/imagecustomizerlib/packagemanager_dnf.go
cwize1: In that case, should we use `tdnf` for azl3 and `rpm` for azl4?

toolkit/tools/pkg/imagecustomizerlib/blsutils.go
```diff
@@ -0,0 +1,257 @@
...
+			val = tok[eq+1:]
+		}
+		args = append(args, grubConfigLinuxArg{Arg: tok, Name: name, Value: val})
+	}
+
+	for i := 0; i < len(value); i++ {
+		c := value[i]
+		switch {
+		case c == '"':
+			inQuotes = !inQuotes
```
cwize1, line 53: Is there character escaping?

cwize1 (approved):

liulanze (approved):

toolkit/tools/pkg/imagecustomizerlib/blsutils.go
vinceaperri: No, kernel comamnd line and BLS entry formats don't have escape sequences.

PR #702, opened 2026-04-16
Add support for inline verity.

cwize1:
Add support for verity where the data and hash tree are placed in the same partition, with the hash at the back.

Changes include:

- Populate the size and end of the partition type during validation, so that downstream code can easily use this data.

- The `mkfs` command doesn't have a common way to specify a filesystem size that doesn't fill the partition. So, switch to using the `mkfs.xxx` commands instead.

- The shrink partition logic needs to know the verity metadata since it needs to be able to shrink inline verity partitions correctly. Hence, the verity metadata collection was moved to be prior to the shrink partitions operation.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/imagecustomizerapi/storage.go
+++ b/toolkit/tools/imagecustomizerapi/storage.go
@@ -232,8 +232,8 @@ func (s *Storage) IsValid() error {
 
 					dataSize, err = calculateInlineVerityDataSize(uint64(partition.Size.Size))
 					if err != nil {
-						return fmt.Errorf("failed to caclulate inline verity device's (%d) data size:\n%w", err,
-							filesystem.DeviceId)
+						return fmt.Errorf("failed to caclulate inline verity device's (%s) data size:\n%w",
+							filesystem.DeviceId, err)
 					}
 
 					filesystem.Size = dataSize
--- a/toolkit/tools/imagecustomizerapi/storage.go
+++ b/toolkit/tools/imagecustomizerapi/storage.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/logger"
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/sliceutils"
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/verityutils"
 )
 
 type VerityPartitionsType int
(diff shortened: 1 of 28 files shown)


toolkit/tools/internal/verityutils/verityutils.go
```diff
@@ -6,30 +6,39 @@
...
-func ReadVeritySuperblock(hashPartitionPath string) (VeritySuperBlock, error) {
+func ReadVeritySuperblock(hashPartitionPath string, hashOffsetBytes uint64) (VeritySuperBlock, error) {
 	hashPartition, err := os.Open(hashPartitionPath)
 	if err != nil {
 		return VeritySuperBlock{}, fmt.Errorf("failed to open hash partition (%s) block device:\n%w", hashPartitionPath, err)
 	}
 	defer hashPartition.Close()
 
+	if hashOffsetBytes != 0 {
+		_, err := hashPartition.Seek(int64(hashOffsetBytes), io.SeekStart)
```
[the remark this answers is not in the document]
cwize1: The `Seek` function will error out if passed a negative number.

liulanze (dismissed):

toolkit/tools/imagecustomizerapi/storage.go
```diff
@@ -495,3 +510,38 @@ func addParentToDevice(deviceId string, deviceMap map[string]any, deviceParents
...
+func calculateInlineVerityDataSize(partitionSize uint64) (uint64, error) {
+	dataBlockSize := uint32(DefaultVerityDataBlockSize)
+	verityBlockSize := uint32(DefaultVerityHashBlockSize)
+	hashAlgorithm := DefaultVerityHashAlgorithm
+
+	left := uint64(0)
+	right := partitionSize / uint64(dataBlockSize)
+
+	// Binary search for the size that perfectly balances the data and the hash sizes.
+	for left <= right {
```
liulanze, line 524: I am thinking if we have a case that when `left` == `right`, so `dataBlocks` == `left` == `right`, and later if `totalBytes != partitionSize`, then is it an infinite loop?

toolkit/tools/imagecustomizerapi/storage.go
```diff
@@ -495,3 +510,38 @@ func addParentToDevice(deviceId string, deviceMap map[string]any, deviceParents
...
+		verityBlocks, err := verityutils.CalculateHashSizeInBlocks(dataBlocks, verityBlockSize, hashAlgorithm)
+		if err != nil {
+			return 0, err
+		}
+
+		totalBytes := verityBlocks*uint64(verityBlockSize) + dataBlocks*uint64(dataBlockSize)
+		if totalBytes == partitionSize {
+			left = dataBlocks
+			break
+		} else if totalBytes > partitionSize {
```
liulanze: Also, if `(left + right) / 2` equals to `left`, for this case:

```
} else if totalBytes > partitionSize {
    right = dataBlocks // right = dataBlocks - 1
} else {
    left = dataBlocks  // left = dataBlocks + 1
}
```

toolkit/tools/imagecustomizerapi/storage.go
cwize1: Fixed

toolkit/tools/imagecustomizerapi/storage.go
cwize1: Fixed

liulanze (approved):

PR #706, opened 2026-04-23
Fix sparse vmlinuz boot failure on XFS by changing cp --sparse to auto

liulanze:
Fix boot failure caused by sparse vmlinuz kernel files on XFS partitions.

When Image Customizer repartitions an image (via `storage:` config), it copies
all files from the old image to the new one using `cp --sparse always`. This
flag aggressively scans file content for zero-byte runs and converts them into
filesystem holes.

The AZL 3.0 vmlinuz kernel binary contains a legitimate run of zero bytes.
`cp --sparse always` turns this into a sparse hole in the destination file.
GRUB cannot read sparse files on XFS, causing a boot failure.

---

## Verification
Reproduced and verified with Azure Linux 3.0 core-efi image
(core-3.0.20260401.vhdx) using the `partitions-xfs-boot.yaml` test config.

**Before fix** (`--sparse always`):
```
filefrag -v /mnt/boot/vmlinuz-6.6.130.1-3.azl3
 ext:     logical_offset:        physical_offset: length:   expected: flags:
   0:        0..    3566:      34935..     38501:   3567:
   1:     3568..    3702:      38503..     38637:    135:             last,eof
```
Logical block 3567 is a sparse hole.

**After fix** (`--sparse auto`):
```
filefrag -v /mnt/boot/vmlinuz-6.6.130.1-3.azl3
 ext:     logical_offset:        physical_offset: length:   expected: flags:
   0:        0..    3702:      31809..     35511:   3703:             last,eof
```
Single contiguous extent, no holes.

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/customizepartitionsfilecopy.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/customizepartitionsfilecopy.go
@@ -68,7 +68,7 @@ func copyPartitionFilesWithOptions(sourceRoot, targetRoot string, noClobber bool
 	// `-a` ensures unix permissions, extended attributes (including SELinux), and sub-directories (-r) are copied.
 	// `--no-dereference` ensures that symlinks are copied as symlinks.
 	copyArgs := []string{
-		"--verbose", "-a", "--no-dereference", "--sparse", "always",
+		"--verbose", "-a", "--no-dereference", "--sparse", "auto",
 		sourceRoot, targetRoot,
 	}
 
--- a/toolkit/tools/pkg/imagecustomizerlib/customizepartitionsfilecopy.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/customizepartitionsfilecopy.go
@@ -68,7 +68,7 @@ func copyPartitionFilesWithOptions(sourceRoot, targetRoot string, noClobber bool
 	// `-a` ensures unix permissions, extended attributes (including SELinux), and sub-directories (-r) are copied.
 	// `--no-dereference` ensures that symlinks are copied as symlinks.
 	copyArgs := []string{
-		"--verbose", "-a", "--no-dereference", "--sparse", "auto",
+		"--verbose", "-a", "--no-dereference",
 		sourceRoot, targetRoot,
 	}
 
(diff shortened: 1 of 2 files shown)


toolkit/tools/pkg/imagecustomizerlib/customizepartitionsfilecopy.go
```diff
@@ -68,7 +68,7 @@ func copyPartitionFilesWithOptions(sourceRoot, targetRoot string, noClobber bool
 	// `-a` ensures unix permissions, extended attributes (including SELinux), and sub-directories (-r) are copied.
 	// `--no-dereference` ensures that symlinks are copied as symlinks.
 	copyArgs := []string{
-		"--verbose", "-a", "--no-dereference", "--sparse", "always",
+		"--verbose", "-a", "--no-dereference", "--sparse", "auto",
```
vinceaperri: Have you considered --sparse=never? It would be good to avoid the ambiguity of a "crude heuristic" according to my `cp --help`, but I'm also not sure why --sparse=always was here in the first place. Perhaps just to save space.

toolkit/tools/pkg/imagecustomizerlib/customizepartitionsfilecopy.go
```diff
@@ -68,7 +68,7 @@ func copyPartitionFilesWithOptions(sourceRoot, targetRoot string, noClobber bool
...
 		sourceRoot, targetRoot,
 	}
```
vinceaperri, line 74: Got some test failures.

vinceaperri (changes requested):

toolkit/tools/pkg/imagecustomizerlib/customizepartitionsfilecopy.go
vinceaperri: If you want to stick with `auto`, as that's the default value, why no remove `--sparse auto` altogether?

liulanze: `--sparse=never` would forcefully expand any legitimately sparse files in the image, which wastes space unnecessarily. Agreed on removing the flag entirely.

vinceaperri (approved):

PR #708, opened 2026-04-23
Add Phase 0 support: mount, recognize, and validate container images

liulanze:
Enables Image Customizer to mount and recognize container images without modifying them, laying the foundation for full support in subsequent phases. This adds distro detection via VARIANT_ID=azurecontainerlinux in os-release (with /usr/lib/os-release fallback per the os-release(5) spec), a new aclDistroHandler with specific ESP path (boot instead of boot/efi) and systemd-boot bootloader detection, and IC-specific fstab discovery at /usr/share/ic/etc/fstab for USR partition. All hardcoded EspDir usages are replaced with distroHandler.GetEspDir() via a new method on the DistroHandler interface. UKI file discovery is broadened from vmlinuz-*.efi to *.efi to support acl.efi naming, and clearBtrfsReadOnlyProperties is added to handle btrfs subvolumes sealed with ro=true at build time.

Test plan
Verified go vet and go test pass for all affected packages. Ran end-to-end no-op image customization (mount, detect as UKI boot type, output VHD) successfully. Need to verify no regression on Azure Linux 3.0 images.

---

### **Checklist**
- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/internal/targetos/targetos.go
+++ b/toolkit/tools/internal/targetos/targetos.go
@@ -55,9 +55,9 @@ func GetInstalledTargetOs(rootfs string) (TargetOs, error) {
 
 		switch variantId {
 		case "azurecontainerlinux":
-			// ACL uses VERSION_ID like "3.0.YYYYMMDD" (e.g. "3.0.20260421").
-			// Accept any version that starts with "3.0.".
-			if !strings.HasPrefix(versionId, "3.0.") {
+			// ACL currently sets VERSION_ID to the full version string (e.g.
+			// "3.0.20260421") Accept any version that starts with "3."
+			if !strings.HasPrefix(versionId, "3.") {
 				return "", fmt.Errorf("unknown VERSION_ID (%s) for Azure Container Linux in os-release", versionId)
 			}
 			return TargetOsAzureContainerLinux3, nil
--- a/toolkit/tools/internal/targetos/targetos.go
+++ b/toolkit/tools/internal/targetos/targetos.go
@@ -16,12 +16,12 @@ import (
 type TargetOs string
 
 const (
-	TargetOsAzureLinux2 TargetOs = "azl2"
-	TargetOsAzureLinux3 TargetOs = "azl3"
+	TargetOsAzureLinux2          TargetOs = "azl2"
+	TargetOsAzureLinux3          TargetOs = "azl3"
 	TargetOsAzureContainerLinux3 TargetOs = "acl3"
-	TargetOsFedora42    TargetOs = "fedora42"
-	TargetOsUbuntu2204  TargetOs = "ubuntu2204"
-	TargetOsUbuntu2404  TargetOs = "ubuntu2404"
+	TargetOsFedora42             TargetOs = "fedora42"
+	TargetOsUbuntu2204           TargetOs = "ubuntu2204"
+	TargetOsUbuntu2404           TargetOs = "ubuntu2404"
 )
 
 func GetInstalledTargetOs(rootfs string) (TargetOs, error) {
--- a/toolkit/tools/internal/targetos/targetos.go
+++ b/toolkit/tools/internal/targetos/targetos.go
@@ -6,6 +6,7 @@ package targetos
 import (
 	"fmt"
 	"path/filepath"
+	"strings"
 
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/envfile"
 )
@@ -15,15 +16,20 @@ type TargetOs string
 const (
 	TargetOsAzureLinux2 TargetOs = "azl2"
 	TargetOsAzureLinux3 TargetOs = "azl3"
+	TargetOsAcl         TargetOs = "acl"
 	TargetOsFedora42    TargetOs = "fedora42"
 	TargetOsUbuntu2204  TargetOs = "ubuntu2204"
 	TargetOsUbuntu2404  TargetOs = "ubuntu2404"
 )
 
 func GetInstalledTargetOs(rootfs string) (TargetOs, error) {
+	// Try /etc/os-release first, then fall back to /usr/lib/os-release.
 	fields, err := envfile.ParseEnvFile(filepath.Join(rootfs, "etc/os-release"))
 	if err != nil {
-		return "", fmt.Errorf("failed to read /etc/os-release file:\n%w", err)
+		fields, err = envfile.ParseEnvFile(filepath.Join(rootfs, "usr/lib/os-release"))
+		if err != nil {
+			return "", fmt.Errorf("failed to read os-release (tried /etc/os-release and /usr/lib/os-release):\n%w", err)
+		}
 	}
 
 	distroId := fields["ID"]
@@ -40,12 +46,29 @@ func GetInstalledTargetOs(rootfs string) (TargetOs, error) {
 		}
 
 	case "azurelinux":
-		switch versionId {
-		case "3.0":
-			return TargetOsAzureLinux3, nil
+		variantId := fields["VARIANT_ID"]
+
+		switch variantId {
+		case "azurecontainerlinux":
+			// ACL uses VERSION_ID like "3.0.YYYYMMDD" (e.g. "3.0.20260421").
+			// Accept any version that starts with "3.0".
+			if !strings.HasPrefix(versionId, "3.0") {
+				return "", fmt.Errorf("unknown VERSION_ID (%s) for Azure Container Linux in /etc/os-release", versionId)
+			}
+			return TargetOsAcl, nil
+
+		case "":
+			// Standard Azure Linux.
+			switch versionId {
+			case "3.0":
+				return TargetOsAzureLinux3, nil
+
+			default:
+				return "", fmt.Errorf("unknown VERSION_ID (%s) for Azure Linux in /etc/os-release", versionId)
+			}
 
 		default:
-			return "", fmt.Errorf("unknown VERSION_ID (%s) for Azure Linux in /etc/os-release", versionId)
+			return "", fmt.Errorf("unknown VARIANT_ID (%s) for Azure Linux in /etc/os-release", variantId)
 		}
 
 	case "fedora":
--- a/toolkit/tools/internal/targetos/targetos.go
+++ b/toolkit/tools/internal/targetos/targetos.go
@@ -18,7 +18,7 @@ type TargetOs string
 const (
 	TargetOsAzureLinux2 TargetOs = "azl2"
 	TargetOsAzureLinux3 TargetOs = "azl3"
-	TargetOsAcl         TargetOs = "acl"
+	TargetOsAzureContainerLinux3 TargetOs = "acl3"
 	TargetOsFedora42    TargetOs = "fedora42"
 	TargetOsUbuntu2204  TargetOs = "ubuntu2204"
 	TargetOsUbuntu2404  TargetOs = "ubuntu2404"
@@ -56,14 +56,14 @@ func GetInstalledTargetOs(rootfs string) (TargetOs, error) {
 		switch variantId {
 		case "azurecontainerlinux":
 			// ACL uses VERSION_ID like "3.0.YYYYMMDD" (e.g. "3.0.20260421").
-			// Accept any version that starts with "3.0".
-			if !strings.HasPrefix(versionId, "3.0") {
+			// Accept any version that starts with "3.0.".
+			if !strings.HasPrefix(versionId, "3.0.") {
 				return "", fmt.Errorf("unknown VERSION_ID (%s) for Azure Container Linux in os-release", versionId)
 			}
-			return TargetOsAcl, nil
+			return TargetOsAzureContainerLinux3, nil
 
-		case "":
-			// Standard Azure Linux.
+		default:
+			// Standard Azure Linux (or unknown variant — treat as standard).
 			switch versionId {
 			case "3.0":
 				return TargetOsAzureLinux3, nil
@@ -71,9 +71,6 @@ func GetInstalledTargetOs(rootfs string) (TargetOs, error) {
 			default:
 				return "", fmt.Errorf("unknown VERSION_ID (%s) for Azure Linux in os-release", versionId)
 			}
-
-		default:
-			return "", fmt.Errorf("unknown VARIANT_ID (%s) for Azure Linux in os-release", variantId)
 		}
 
 	case "fedora":
--- a/toolkit/tools/internal/targetos/targetos.go
+++ b/toolkit/tools/internal/targetos/targetos.go
@@ -4,7 +4,9 @@
 package targetos
 
 import (
+	"errors"
 	"fmt"
+	"io/fs"
 	"path/filepath"
 	"strings"
 
@@ -26,6 +28,9 @@ func GetInstalledTargetOs(rootfs string) (TargetOs, error) {
 	// Try /etc/os-release first, then fall back to /usr/lib/os-release.
 	fields, err := envfile.ParseEnvFile(filepath.Join(rootfs, "etc/os-release"))
 	if err != nil {
+		if !errors.Is(err, fs.ErrNotExist) {
+			return "", fmt.Errorf("failed to read /etc/os-release:\n%w", err)
+		}
 		fields, err = envfile.ParseEnvFile(filepath.Join(rootfs, "usr/lib/os-release"))
 		if err != nil {
 			return "", fmt.Errorf("failed to read os-release (tried /etc/os-release and /usr/lib/os-release):\n%w", err)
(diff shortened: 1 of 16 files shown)


toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
```diff
@@ -0,0 +1,89 @@
...
+
+// aclDistroHandler implements DistroHandler for Azure Container Linux (ACL).
+// ACL uses systemd-boot + UKI (no GRUB) and has an immutable /usr with dm-verity.
+type aclDistroHandler struct {
+	packageManager rpmPackageManagerHandler
+}
+
+func newAclDistroHandler() *aclDistroHandler {
+	return &aclDistroHandler{
+		packageManager: newTdnfPackageManager("3.0"),
```
jiria, line 25: Is this accurate?

jiria (dismissed):

toolkit/tools/internal/targetos/targetos.go
```diff
@@ -15,15 +16,20 @@ type TargetOs string
...
 	TargetOsUbuntu2204  TargetOs = "ubuntu2204"
 	TargetOsUbuntu2404  TargetOs = "ubuntu2404"
 )
 
 func GetInstalledTargetOs(rootfs string) (TargetOs, error) {
+	// Try /etc/os-release first, then fall back to /usr/lib/os-release.
 	fields, err := envfile.ParseEnvFile(filepath.Join(rootfs, "etc/os-release"))
 	if err != nil {
-		return "", fmt.Errorf("failed to read /etc/os-release file:\n%w", err)
+		fields, err = envfile.ParseEnvFile(filepath.Join(rootfs, "usr/lib/os-release"))
```
cwize1, line 34: Only check the other path on file-not-found error.

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -429,7 +429,7 @@ func extractUkiEntriesIfPresent(chrootDir, buildDir string) ([]SystemDBootEntry,
 
 	var entries []SystemDBootEntry
 	for kernelName, cmdline := range cmdlines {
-		efiPath := filepath.Join("/boot/efi/EFI/Linux", fmt.Sprintf("%s.efi", kernelName))
+		efiPath := filepath.Join("/"+distroHandler.GetEspDir(), "EFI/Linux", fmt.Sprintf("%s.efi", kernelName))
```
cwize1: `filepath.Join(“/“, distroHandler.GetEspDir(), "EFI/Linux", fmt.Sprintf("%s.efi", kernelName))`

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
```diff
@@ -965,13 +965,17 @@ func getKernelNameFromUki(ukiPath string) (string, error) {
...
+		// Standard UKI naming: vmlinuz-<version>.efi → vmlinuz-<version>
+		return "vmlinuz-" + matches[1], nil
 	}
 
-	// Reconstruct kernel name (vmlinuz-<version>, e.g., vmlinuz-6.6.51.1-5.azl3)
-	kernelName := "vmlinuz-" + matches[1]
-	return kernelName, nil
+	// Non-standard UKI naming (e.g., acl.efi): use filename without .efi extension
+	if strings.HasSuffix(fileName, ".efi") {
+		return strings.TrimSuffix(fileName, ".efi"), nil
```
cwize1, line 975: Is the name of the kernel in `/usr/lib/modules` actually `acl`?

toolkit/tools/pkg/imagecustomizerlib/imageutils.go
```diff
@@ -381,11 +385,55 @@ func createPartIdToPartUuidMap(partIDToDevPathMap map[string]string, diskPartiti
...
 
 func extractOSRelease(imageConnection *imageconnection.ImageConnection) (string, error) {
+	// Try /etc/os-release first, then fall back to /usr/lib/os-release.
+	// The fallback is per the os-release(5) spec and is needed for distros like
+	// ACL where /etc is an overlay that may not be mounted during customization.
 	osReleasePath := filepath.Join(imageConnection.Chroot().RootDir(), "etc/os-release")
 	data, err := file.Read(osReleasePath)
 	if err != nil {
-		return "", fmt.Errorf("failed to read /etc/os-release:\n%w", err)
+		osReleasePath = filepath.Join(imageConnection.Chroot().RootDir(), "usr/lib/os-release")
```
cwize1, line 397: Only check other path if error is file-not-found.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -203,6 +203,17 @@ func findFstabInRoot(diskPartition diskutils.PartitionInfo, tmpDir string) ([]st
 		matchingPaths = append(matchingPaths, "")
 	}
 
+	// ACL places an fstab at /usr/share/ic/etc/fstab on the USR partition. This
```
cwize1: To prevent this from applying to other distros, check the partition type UUID.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
liulanze: Yes, this is accurate. ACL is based on Azure Linux 3.0 and uses `tdnf`. The "3.0" is the --releasever value passed to tdnf for package repo resolution, matching AZL3's repo structure.

toolkit/tools/internal/targetos/targetos.go
liulanze: Fixed.

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
liulanze: Fixed.

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
liulanze: no, the kernel in /usr/lib/modules/ is not named acl. it would be something like 6.6.x.y-z.acl. The acl return value works for Phase 0 because it's only used as a map key in the SELinux cmdline extraction path. 

For Phase 2 (create/modify), this function will need to be reworked to extract the actual kernel version from the UKI binary rather than inferring it from the filename.

toolkit/tools/pkg/imagecustomizerlib/imageutils.go
liulanze: Fixed.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
liulanze: ACL's USR partition uses the generic Linux partition type UUID (0fc63daf-8483-4772-8e79-3d69d8477de4), not the Discoverable Partitions Specification USR UUID (8484680c-9521-48c6-9c11-b0720656f69e), so a partition type check doesn't work here. The share/ic/etc/fstab path is unique to ACL so the path itself is sufficient to prevent false matches on other distros.

toolkit/tools/pkg/imagecustomizerlib/imageutils.go
```diff
@@ -381,11 +385,55 @@ func createPartIdToPartUuidMap(partIDToDevPathMap map[string]string, diskPartiti
...
+			continue
+		}
+
+		mountPath := filepath.Join(imageConnection.Chroot().RootDir(), mp.GetTarget())
+
+		// Check if the subvolume is read-only.
+		stdout, _, err := shell.Execute("btrfs", "property", "get", "-ts", mountPath, "ro")
+		if err != nil {
+			// Not all btrfs mounts have subvolumes; skip on error.
+			continue
```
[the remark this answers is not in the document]
liulanze: Fixed.

toolkit/tools/internal/targetos/targetos.go
```diff
@@ -40,12 +46,29 @@ func GetInstalledTargetOs(rootfs string) (TargetOs, error) {
...
-			return TargetOsAzureLinux3, nil
+		variantId := fields["VARIANT_ID"]
+
+		switch variantId {
+		case "azurecontainerlinux":
+			// ACL uses VERSION_ID like "3.0.YYYYMMDD" (e.g. "3.0.20260421").
+			// Accept any version that starts with "3.0".
+			if !strings.HasPrefix(versionId, "3.0") {
+				return "", fmt.Errorf("unknown VERSION_ID (%s) for Azure Container Linux in /etc/os-release", versionId)
+			}
```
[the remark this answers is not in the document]
liulanze: Fxied.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func (d *aclDistroHandler) GetTargetOs() targetos.TargetOs {
+	return targetos.TargetOsAcl
+}
+
+func (d *aclDistroHandler) ValidateConfig(rc *ResolvedConfig) error {
+	// TODO: Block unsupported operations (repartitioning, GRUB, initramfs regen, etc.)
+	return nil
+}
```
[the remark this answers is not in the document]
liulanze: Fixed.

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
```diff
@@ -965,13 +965,17 @@ func getKernelNameFromUki(ukiPath string) (string, error) {
...
+	}
```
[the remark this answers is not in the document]
liulanze: No fix needed. This is the same concern as the earlier reviewer comment about acl.efi → getKernelVersion.

bfjelds:
is there a plan to support ACL in osmodifier codepaths as well? (i imagine a lot of this will apply to both IC and OSM, but am curious if OSM is in the support-ACL plans)

liulanze:
> is there a plan to support ACL in osmodifier codepaths as well? (i imagine a lot of this will apply to both IC and OSM, but am curious if OSM is in the support-ACL plans)

The current focus is IC support for ACL. @jiria helps to follow up please.

toolkit/tools/internal/targetos/targetos.go
```diff
@@ -36,16 +47,33 @@ func GetInstalledTargetOs(rootfs string) (TargetOs, error) {
 			return TargetOsAzureLinux2, nil
 
 		default:
-			return "", fmt.Errorf("unknown VERSION_ID (%s) for CBL-Mariner in /etc/os-release", versionId)
+			return "", fmt.Errorf("unknown VERSION_ID (%s) for CBL-Mariner in os-release", versionId)
```
vinceaperri, line 50: You could save to a variable the actual abspath read so you can embed in the error message here (instead of "os-release")? Would be more informative for the user.

toolkit/tools/internal/targetos/targetos.go
```diff
@@ -36,16 +47,33 @@ func GetInstalledTargetOs(rootfs string) (TargetOs, error) {
...
-		switch versionId {
-		case "3.0":
-			return TargetOsAzureLinux3, nil
+		variantId := fields["VARIANT_ID"]
+
+		switch variantId {
+		case "azurecontainerlinux":
+			// ACL uses VERSION_ID like "3.0.YYYYMMDD" (e.g. "3.0.20260421").
+			// Accept any version that starts with "3.0".
+			if !strings.HasPrefix(versionId, "3.0") {
```
vinceaperri: This is non-standard. Azure Linux 3.0 base images put this in the VERSION field:

```
NAME="Microsoft Azure Linux"
VERSION="3.0.20260401"
ID=azurelinux
VERSION_ID="3.0"
PRETTY_NAME="Microsoft Azure Linux 3.0"
```

Have you raised this as an issue with ACL?

vinceaperri: nit: I would check for the `3.0.` prefix

toolkit/tools/internal/targetos/targetos.go
vinceaperri: Similar comment in the 6 other error strings below.

toolkit/tools/internal/targetos/targetos.go
```diff
@@ -36,16 +47,33 @@ func GetInstalledTargetOs(rootfs string) (TargetOs, error) {
...
+			case "3.0":
+				return TargetOsAzureLinux3, nil
+
+			default:
+				return "", fmt.Errorf("unknown VERSION_ID (%s) for Azure Linux in os-release", versionId)
+			}
 
 		default:
-			return "", fmt.Errorf("unknown VERSION_ID (%s) for Azure Linux in /etc/os-release", versionId)
+			return "", fmt.Errorf("unknown VARIANT_ID (%s) for Azure Linux in os-release", variantId)
```
vinceaperri: Do we need to error here? Previously, we just ignored VARIANT_ID altogether. I wouldn't want to introduce errors for customers who happen to use a non-ACL variant of Azure Linux. For backwards-compatibility reasons, I would have the "" case block be the default block instead.

toolkit/tools/internal/targetos/targetos.go
```diff
@@ -15,15 +18,23 @@ type TargetOs string
 const (
 	TargetOsAzureLinux2 TargetOs = "azl2"
 	TargetOsAzureLinux3 TargetOs = "azl3"
+	TargetOsAcl         TargetOs = "acl"
```
vinceaperri: nit: TargetOsAzureContainerLinux3 would be a clearer name, and acl3 a clearer value.

toolkit/tools/internal/targetos/targetos.go
jiria, line 50: Aside, should we rename CBL-Mariner here?

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
jiria: To confirm, where will be the tdnf binary loaded from?

vinceaperri (changes requested):

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
liulanze: The tdnf binary comes from inside the image, that IC runs all package manager commands chrooted into the mounted image (`Chroot(imageChroot.ChrootDir())`), so it uses the tdnf that ships in the ACL image itself. That said, for Phase 0/1 this is unused since `ValidateConfig` blocks all package operations on ACL.

toolkit/tools/internal/targetos/targetos.go
liulanze: This is pre-existing code, not our change. The case "mariner" branch handles CBL-Mariner 2.0 images where ID=mariner in os-release, so "CBL-Mariner" is accurate here. We can rename it in a separate cleanup PR if desired.

liulanze: Good suggestion, but these error messages are pre-existing, not part of our ACL changes. Happy to do that improvement in a separate cleanup PR to keep this one focused.

toolkit/tools/internal/targetos/targetos.go
liulanze: ACL's os-release is generated in `uki_install.sh` and sets `VERSION_ID="${IMAGE_VERSION_ID}"` which resolves to the full version string like `3.0.20260421`. We used `strings.HasPrefix(versionId, "3.0")` to handle this. @jiria do we want to align the `VERSION_ID` to 3.0?

liulanze: Fixed.

toolkit/tools/internal/targetos/targetos.go
liulanze: Fxied.

toolkit/tools/internal/targetos/targetos.go
liulanze: Fixed.

toolkit/tools/internal/targetos/targetos.go
liulanze: If ACL aligns VERSION_ID to "3.0", we can do that in a separate ACL-side PR. Our IC code handles both formats today, so this is non-blocking.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
jiria: Sure, my point is there is no tdnf in the image. Fine with me if that helps to unblock the initial phase, just wanted to call it out.

jiria (approved):

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
liulanze: Noted, thanks.

vinceaperri (approved):

PR #710, opened 2026-04-24
Fix: SELinux mode incorrectly detected as disabled when no kernel args present

vinceaperri:
getSELinuxModeFromLinuxArgs() treats the absence of SELinux kernel command-line arguments (security=, selinux=, enforcing=) as Disabled. No args, however, means the mode is determined by /etc/selinux/config, and should return Default to trigger the fallback path in GetSELinuxMode().
    
This causes selinuxSetFiles() to skip setfiles relabeling on any image that relies on config without explicit kernel cmdline SELinux args. Files modified by imagecustomizer are left with unlabeled_t contexts. If SELinux is actually enforcing at boot (per the config file), services are denied access to these files, causing dbus-broker, systemd-logind, sshd, and login to fail.
    
The bug was masked on prior Azure Linux versions:
    
- AZL2: No SELinux at all (no config, no kernel args). The incorrect Disabled result happens to be harmless since there's nothing to relabel.
- AZL3: Explicitly sets selinux=0 on the kernel command line, so the function correctly returns Disabled via the explicit check.
- AZL4: First image that omits SELinux kernel args but has SELINUX=enforcing in config. The bug causes setfiles to be skipped, producing broken images.

This also updates 3.0-default-grub test file to include selinux=0, matching the actual file shipped by AZL3 (3.0.20260414)

### DIFF
--- a/test/vmtests/vmtests/osmodifier/test_osmodifier.py
+++ b/test/vmtests/vmtests/osmodifier/test_osmodifier.py
@@ -315,11 +315,7 @@ def test_osmodifier_boot_config(
     assert "overlay" in grub_cfg
 
     sestatus_output = ssh_client.run("sudo sestatus").stdout
-    sestatus = dict(
-        line.split(":", 1)
-        for line in sestatus_output.splitlines()
-        if ":" in line
-    )
+    sestatus = dict(line.split(":", 1) for line in sestatus_output.splitlines() if ":" in line)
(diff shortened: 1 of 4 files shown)


liulanze (dismissed):

cwize1:
If I recall correctly, on AZL2 no SELinux args did mean it was disabled. But this changed for AZL3.

PR #713, opened 2026-04-29
Add Phase 1 ACL support: ValidateConfig for mutable partition operations

liulanze:
Expand ACL's ValidateConfig to block unsupported operations (packages,
additionalDirs, overlays, SELinux, kernelCommandLine) while allowing
mutable partition ops (additionalFiles, services, hostname, users,
groups, modules, scripts, UKI passthrough).

Note: hash-offset verity support originally in this commit is now
covered by the inline verity PR (#702).


liulanze:
Will merge the phase 1 config validation and phase 2 as one PR.

PR #723, opened 2026-05-06
azl4: Use Distro-Aware isPackageInstalled in LiveOS code and TestBaseConfigsFullRun

vinceaperri:
Removes isPackageInstalled() which uses tdnf so we can execute these code paths for distros that don't use it.

This function is safe to remove. It's a duplicate of `func (pm *tdnfPackageManager) isPackageInstalled`


toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -204,7 +204,7 @@ func createLiveOSFromRawHelper(ctx context.Context, buildDir string, inputArtifa
 	defer rawImageConnection.Close()
 
 	// Detect the distro handler for the image.
-	distroHandler, err := NewDistroHandlerFromChroot(rawImageConnection.Chroot())
+	distroHandler, err = NewDistroHandlerFromChroot(rawImageConnection.Chroot())
```
liulanze: I think we can remove the dead `distroHandler` parameter from both `createLiveOSFromRaw` and `createLiveOSFromRawHelper`. The `im.distroHandler` passed, which is silently discarded, like `createLiveOSFromRawHelper` immediately overwrites it.

vinceaperri: Done!

liulanze (approved):

PR #724, opened 2026-05-06
azl4: Create Distro-Aware FindBootPartitionUuidFromEsp for output.artifacts API support

vinceaperri:
The grub.cfg stubs on the EFI partitions are in a different location and require a different regular expression to grep for the boot partition UUID on AZL4, so this PR introduces DistroHandler.FindBootPartitionUuidFromEsp to handle that.

Example azl4 grub.cfg stub (from Alpha 2 release):

```
search --fs-uuid --set=root 3190eea2-a4b1-4399-9679-e0840cf8eb75
set prefix=($root)/boot/grub2
configfile ($root)/boot/grub2/grub.cfg
```

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/distrohandler_fedora.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/distrohandler_fedora.go
@@ -93,6 +93,7 @@ func (d *fedoraDistroHandler) GetEspDir() string {
 }
 
 func (d *fedoraDistroHandler) FindBootPartitionUuidFromEsp(espMountDir string) (string, error) {
+	// Reading Fedora's grub.cfg stub is not supported, so for now just use Azure Linux 3.0's values.
 	return readBootPartitionUuidFromGrubCfg(filepath.Join(espMountDir, espGrubCfgPathAzl3), bootPartitionRegexAzl3)
 }
 
--- a/toolkit/tools/pkg/imagecustomizerlib/distrohandler_fedora.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/distrohandler_fedora.go
@@ -6,6 +6,7 @@ package imagecustomizerlib
 import (
 	"context"
 	"fmt"
+	"path/filepath"
 	"slices"
 
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/imagecustomizerapi"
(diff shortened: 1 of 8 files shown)


toolkit/tools/pkg/imagecustomizerlib/distrohandler_fedora.go
```diff
@@ -91,6 +92,10 @@ func (d *fedoraDistroHandler) GetEspDir() string {
 	return "boot/efi"
 }
 
+func (d *fedoraDistroHandler) FindBootPartitionUuidFromEsp(espMountDir string) (string, error) {
+	return readBootPartitionUuidFromGrubCfg(filepath.Join(espMountDir, espGrubCfgPathAzl3), bootPartitionRegexAzl3)
```
liulanze, line 97: Should here be used `espGrubCfgPathAzl3` or `espGrubCfgPathAzl4`?

toolkit/tools/pkg/imagecustomizerlib/distrohandler_ubuntu.go
```diff
@@ -112,6 +113,10 @@ func (d *ubuntuDistroHandler) GetEspDir() string {
 	return "boot/efi"
 }
 
+func (d *ubuntuDistroHandler) FindBootPartitionUuidFromEsp(espMountDir string) (string, error) {
+	return readBootPartitionUuidFromGrubCfg(filepath.Join(espMountDir, espGrubCfgPathAzl3), bootPartitionRegexAzl3)
```
liulanze, line 118: Is Ubuntu using the same Grub path as Azl? I could be wrong, but remember I saw `EFI/ubuntu/grub.cfg` once before.

vinceaperri: I just kept the functionality as-is for other distributions. They used azl3-specific paths before this PR, and this PR doesn't change it with the addition of the distroHandler function. Let me add a comment to address this specifically.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_fedora.go
vinceaperri: Fedora doesn't support this code path, so I just kept the values as-is.

liulanze (approved):

PR #725, opened 2026-05-06
Add ACL support: package install, UKI create mode, initramfs regen, and boot flow fixes.

liulanze:
## Summary

Enable Phase 2 Image Customizer support for Azure Container Linux (ACL), covering package installation, initramfs regeneration, UKI create mode, SELinux, and kernel command line modification.

### Testing

- `go vet ./pkg/imagecustomizerlib/...` — pass
- `go test ./pkg/imagecustomizerlib/...` — pass
- E2E validation pending after PR-B merges (ACL image build + IC customization with UKI create mode)

---

### **Checklist**
- [x] Tests added/updated


toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
```diff
@@ -100,13 +247,36 @@ func (d *aclDistroHandler) WriteGrub2ConfigFile(grub2Config string,
 }
 
 func (d *aclDistroHandler) RegenerateInitramfs(ctx context.Context, imageChroot *safechroot.Chroot) error {
-	return fmt.Errorf("initramfs regeneration is not yet supported for ACL")
+	logger.Log.Infof("Regenerating initramfs for ACL")
+
+	// Validate dracut is available in the image.
+	if !d.IsPackageInstalled(imageChroot, "dracut") {
```
vinceaperri: This check is only done in ACL, but other distros use dracut too. Please either remove it or lift it to a shared helper. I'd prefer removing it, since the dracut exec will fail immediately after with a clear error anyway.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
```diff
@@ -31,8 +39,27 @@ func (d *aclDistroHandler) GetTargetOs() targetos.TargetOs {
 }
 
 func (d *aclDistroHandler) ValidateConfig(rc *ResolvedConfig) error {
-	// ACL Phase 0: only mount/recognize/passthrough is supported.
-	// Block operations that would fail with confusing errors later.
+	// ACL supports mutable partition modifications and package installation
```
vinceaperri: Can you put this information in the distribution support table instead?

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
```diff
@@ -57,17 +98,123 @@ func (d *aclDistroHandler) ManagePackages(ctx context.Context, buildDir string,
 	config *imagecustomizerapi.OS, imageChroot *safechroot.Chroot, toolsChroot *safechroot.Chroot,
 	rpmsSources []string, useBaseImageRpmRepos bool, snapshotTime imagecustomizerapi.PackageSnapshotTime,
 ) error {
-	return managePackagesRpm(
-		ctx, buildDir, baseConfigPath, config, imageChroot, toolsChroot, rpmsSources, useBaseImageRpmRepos,
-		snapshotTime, d.packageManager)
+	// ACL has no tdnf inside the image. Run tdnf from the host with --installroot
```
vinceaperri: This is great. tdnf inside the image has been a longstanding dependency, so dropping it for ACL is a real win. Could this code be factored into the managePackagesRpm() as a new packageManager implementation instead? Nothing about driving tdnf from the host with --installroot is ACL-specific, so keeping it distro-agnostic would enable other distros in the future and would fit the existing pattern of the code. --installroot is already used in managePackagesRpm when toolsChroot != nil.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
```diff
@@ -57,17 +98,123 @@ func (d *aclDistroHandler) ManagePackages(ctx context.Context, buildDir string,
...
+		if err == nil {
+			return true
+		}
+	}
+	return false
 }
 
 func (d *aclDistroHandler) GetAllPackagesFromChroot(imageChroot safechroot.ChrootInterface) ([]OsPackage, error) {
-	return getAllPackagesFromChrootRpm(imageChroot)
+	// ACL does not ship an RPM database in the image, so rpm -qa is not available.
```
vinceaperri: This is only used when the output is COSI/bare-metal. Do you intend to support this? I don't see any declaration of support for this or any qualifications on that support, and appears unrelated to the PR, so I would continue returning getAllPackagesFromChrootRpm(imageChroot) here (and presumably throwing an error).

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
```diff
@@ -100,13 +247,36 @@ func (d *aclDistroHandler) WriteGrub2ConfigFile(grub2Config string,
...
+		Chroot(imageChroot.ChrootDir()).
+		Execute()
+	if err != nil {
+		return fmt.Errorf("failed to rebuild initramfs for ACL:\n%w", err)
+	}
+
+	return nil
 }
 
 func (d *aclDistroHandler) ConfigureDiskBootLoader(imageConnection *imageconnection.ImageConnection,
```
vinceaperri, line 275: Likewise, I wouldn't mask the error here either. It should never run anyway because create image and resetting the bootloader is not supported for ACL. Your previous error was better because it didn't mask the issue (that this code path was being called on an ACL image).

toolkit/tools/pkg/imagecustomizerlib/customizeos.go
```diff
@@ -54,6 +54,20 @@ func doOsCustomizations(ctx context.Context, rc *ResolvedConfig, imageConnection
 			if err != nil {
 				return err
 			}
+
+			// For non-GRUB distros, save UKI cmdline early so handleBootLoader/
```
vinceaperri: What do you mean by save UKI cmdline eary? When would it otherwise be saved? Is it being saved twice now in 'create' mode?

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
```diff
@@ -57,17 +98,123 @@ func (d *aclDistroHandler) ManagePackages(ctx context.Context, buildDir string,
...
+
+	return nil
 }
 
 func (d *aclDistroHandler) IsPackageInstalled(imageChroot safechroot.ChrootInterface, packageName string) bool {
-	return d.packageManager.isPackageInstalled(imageChroot, packageName)
+	// ACL does not ship an RPM database or tdnf in the image, so package
+	// queries via rpm/tdnf are not possible. Instead, check for known
+	// binaries or files that each package provides.
+	knownPackageFiles := map[string][]string{
```
vinceaperri, line 170: This logic should be encapsulated in a new packageManager type. The contents of knownPackageFiles is the only ACL-specific thing here, and could be used to instantiate a new packageManager type in the distroHandler constructor.

vinceaperri (changes requested):

toolkit/tools/pkg/imagecustomizerlib/customizeos.go
liulanze: Non-GRUB distros have no grub.cfg, so we need to extract the cmdline from the existing UKIs and save it to uki-kernel-info.json before handleBootLoader/handleSELinux run, they read and modify the cmdline through that file. prepareUkiHelper then reads it back with the modifications applied.

So yes, it's written twice: first as the initial extraction, then updated with the final kernel-to-initramfs mapping.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
liulanze: Should we wait until ACL is to public release? The ValidateConfig code itself already enforces the restrictions via error returns, so the comment is just documentation.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
liulanze: Agree this should eventually be distro-agnostic. For this PR, I'd prefer to keep it in the ACL handler since it's the only consumer right now and the error paths are ACL-specific.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
liulanze: Same here. For now it's only used by ACL and is pretty small (~28 lines). I'll bundle this with the host-side tdnf refactor in the follow-up. Does that work?

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
liulanze: Fixed. Thanks.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
liulanze: Fixed, thanks.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
liulanze: Fixed.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
vinceaperri: Sounds good!

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
vinceaperri: Sounds good!

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
vinceaperri: It's currently a documentation gap for those who will be using ACL with the next release. That being said, we also shouldn't add "Yes" to the table until tests are added for the distribution and feature combination, so we can leave out until tests are added. We could add it with Preview for the features listed here.

toolkit/tools/pkg/imagecustomizerlib/customizeselinux.go
```diff
@@ -120,7 +124,9 @@ func UpdateSELinuxModeInConfigFile(selinuxMode imagecustomizerapi.SELinuxMode, i
 	return nil
 }
 
-func selinuxSetFiles(ctx context.Context, selinuxMode imagecustomizerapi.SELinuxMode, imageChroot *safechroot.Chroot) error {
+func selinuxSetFiles(ctx context.Context, selinuxMode imagecustomizerapi.SELinuxMode, imageChroot *safechroot.Chroot,
+	distroHandler DistroHandler,
```
vinceaperri: distroHandler is only used in this function to call distroHandler.GetSELinuxConfigDir(). Why not just pass the value of distroHandler.GetSELinuxConfigDir() to this function?

toolkit/tools/pkg/imagecustomizerlib/distrohandler_azurelinux.go
```diff
@@ -118,6 +118,10 @@ func (d *azureLinuxDistroHandler) FindBootPartitionUuidFromEsp(espMountDir strin
 	return readBootPartitionUuidFromGrubCfg(filepath.Join(espMountDir, espGrubCfgPath), bootPartitionRegex)
 }
 
+func (d *azureLinuxDistroHandler) GetSELinuxConfigDir() string {
+	return "etc/selinux"
```
vinceaperri: Could you add a constant for this shared value?

toolkit/tools/pkg/imagecustomizerlib/customizeselinux.go
```diff
@@ -81,24 +81,28 @@ func handleSELinux(ctx context.Context, buildDir string, selinuxMode imagecustom
...
 	if err != nil {
 		return imagecustomizerapi.SELinuxModeDefault, err
 	}
 
 	return selinuxMode, nil
 }
 
-func UpdateSELinuxModeInConfigFile(selinuxMode imagecustomizerapi.SELinuxMode, imageChroot safechroot.ChrootInterface) error {
+func UpdateSELinuxModeInConfigFile(selinuxMode imagecustomizerapi.SELinuxMode, imageChroot safechroot.ChrootInterface,
+	distroHandler DistroHandler,
```
vinceaperri: distroHandler is only used in this function to call distroHandler.GetSELinuxConfigDir(). Similarly, could we pass this value instead?

toolkit/tools/pkg/imagecustomizerlib/customizeos.go
```diff
@@ -54,6 +54,20 @@ func doOsCustomizations(ctx context.Context, rc *ResolvedConfig, imageConnection
...
+			// handleSELinux can modify it via uki-kernel-info.json.
+			bootloaderType, err := distroHandler.DetectBootloaderType(imageChroot)
+			if err != nil {
+				return err
+			}
+
+			if bootloaderType != BootloaderTypeGrub {
```
vinceaperri: nit: Keeping the expression positive would make it clearer: `bootloaderType == BootloaderTypeSystemdBoot`

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
```diff
@@ -221,6 +221,39 @@ func extractAndSaveUkiCmdline(buildDir string, imageChroot *safechroot.Chroot, d
 	return nil
 }
 
+// extractAndSaveUkiCmdlineForCreateMode extracts kernel cmdline from existing
+// UKIs and their addons, saving to uki-kernel-info.json before handleBootLoader/
+// handleSELinux run. Needed for non-GRUB distros in UKI create mode.
+func extractAndSaveUkiCmdlineForCreateMode(buildDir string, imageChroot *safechroot.Chroot, distroHandler DistroHandler) error {
```
vinceaperri: distroHandler is only used to call distroHandler.GetEspDir(). Can you just pass the value of espDir instead?

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
```diff
@@ -318,13 +351,32 @@ func prepareUkiHelper(ctx context.Context, buildDir string, uki *imagecustomizer
...
 	kernelInfo := make(map[string]UkiKernelInfo)
 
 	for kernel, initramfs := range kernelToInitramfs {
-		cmdline, exists := kernelToArgs[kernel]
-		if !exists {
-			return fmt.Errorf("no command line arguments found for kernel (%s)", kernel)
+		var cmdline string
+		if existErr == nil {
+			if existingInfo, ok := existingKernelInfo[kernel]; ok {
+				// Use cmdline from early extraction (potentially modified by handleBootLoader/handleSELinux).
```
vinceaperri: This comment is redundant with l. 355 above.

toolkit/tools/pkg/imagecustomizerlib/customizeos.go
vinceaperri: It would be good to explain here why the modify path does not need a `bootloaderType == BootloaderTypeSystemdBoot` gate, or to explain above why create needs one above.

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
vinceaperri: This looks very similar to extractAndSaveUkiCmdline. The only behavior difference I see is that modify reads cmdline from a single hardcoded path, while create mode delegates to extractKenrelCmdlineFromUkiEfis, which reads the main UKI plus every addon in order. Everything else is stuff you can unify around (ensuring the directory exists, initramfs zero value, and "no UKI files found" hard error). Could they be unified?

toolkit/tools/pkg/imagecustomizerlib/customizeos.go
vinceaperri: This pre-existing the PR, but should modify mode not read the main UKI's .cmdline? What if it's non-empty? What about modifying an image with multiple addons? Apologies if this has already been discussed during design phase.

vinceaperri (changes requested):

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
liulanze: Fixed.

toolkit/tools/pkg/imagecustomizerlib/customizeselinux.go
liulanze: Fixed.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_azurelinux.go
liulanze: Fixed.

toolkit/tools/pkg/imagecustomizerlib/customizeselinux.go
liulanze: Fixed.

toolkit/tools/pkg/imagecustomizerlib/customizeos.go
liulanze: Fixed.

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
liulanze: Fixed.

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
liulanze: Fixed.

toolkit/tools/pkg/imagecustomizerlib/customizeos.go
liulanze: Fixed.

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
liulanze: Fixed.

toolkit/tools/pkg/imagecustomizerlib/customizeos.go
liulanze: The unified function now uses `extractKernelCmdlineFromUkiEfis`, which reads the main UKI `.cmdline` and all addons in lexicographic order, mirroring systemd-boot's behavior.

docs/imagecustomizer/api/distribution-support.md
```diff
@@ -10,76 +10,76 @@ The following tables show which APIs are supported for each distribution.
...
+| &emsp;[--cosi-compression-level](./cli/customize.md#--cosi-compression-levellevel)                  | Yes             | No                  | No                        |
+| &emsp;[--output-selinux-policy-path](./cli/customize.md#--output-selinux-policy-pathdirectory-path) | Yes             | No                  | No                        |
+| &emsp;[--config-file](./cli/customize.md#--config-filefile-path)                                    | Yes             | Preview             | Preview                   |
+| &emsp;[--rpm-source](./cli/customize.md#--rpm-sourcepath)                                           | Yes             | No                  | No                        |
+| &emsp;[--disable-base-image-rpm-repos](./cli/customize.md#--disable-base-image-rpm-repos)           | Yes             | No                  | No                        |
+| &emsp;[--package-snapshot-time](./cli/customize.md#--package-snapshot-time)                         | Yes             | No                  | No                        |
```
jiria, line 45: Is this an expensive feature to keep?

docs/imagecustomizer/api/distribution-support.md
```diff
@@ -10,76 +10,76 @@ The following tables show which APIs are supported for each distribution.
...
+| &emsp;&emsp;`vhd`                                                                                   | Yes             | Preview             | Preview                   |
+| &emsp;&emsp;`vhdx`                                                                                  | Yes             | Preview             | Preview                   |
+| &emsp;[--cosi-compression-level](./cli/customize.md#--cosi-compression-levellevel)                  | Yes             | No                  | No                        |
+| &emsp;[--output-selinux-policy-path](./cli/customize.md#--output-selinux-policy-pathdirectory-path) | Yes             | No                  | No                        |
+| &emsp;[--config-file](./cli/customize.md#--config-filefile-path)                                    | Yes             | Preview             | Preview                   |
+| &emsp;[--rpm-source](./cli/customize.md#--rpm-sourcepath)                                           | Yes             | No                  | No                        |
```
jiria, line 43: Is this an expensive feature to keep?

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
```diff
@@ -58,13 +81,117 @@ func (d *aclDistroHandler) ManagePackages(ctx context.Context, buildDir string,
...
+		// systemd-boot: on ACL the signed systemd-boot binary is installed as
+		// grubx64.efi (shim's second-stage) on the ESP. BOOTX64.EFI is shim.
+		"systemd-boot": {
+			"boot/EFI/BOOT/grubx64.efi",
```
jiria: Is there nothing more unique to key off (perhaps in addition to)?

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
```diff
@@ -105,13 +243,27 @@ func (d *aclDistroHandler) WriteGrub2ConfigFile(grub2Config string,
 }
 
 func (d *aclDistroHandler) RegenerateInitramfs(ctx context.Context, imageChroot *safechroot.Chroot) error {
-	return fmt.Errorf("initramfs regeneration is not yet supported for ACL")
+	logger.Log.Infof("Regenerating initramfs for ACL")
+
+	ctx, span := startRegenerateInitramfsSpan(ctx)
+	defer span.End()
+
+	err := shell.NewExecBuilder("dracut", "--force", "--regenerate-all").
```
jiria, line 263: Does this currently all work and the resulting image works as expected?

docs/imagecustomizer/api/distribution-support.md
liulanze: Not expensive I think, the "No" is that it hasn't been explicitly validated e2e for ACL yet rather than a real implementation gap. We can flip it to "Yes" once we add a test for it.

docs/imagecustomizer/api/distribution-support.md
liulanze: More expensive than --rpm-source. Supporting it would require a host-side tdnf config equivalent.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
liulanze: Fixed.

PR #731, opened 2026-05-08
azl4: Add functional and vm tests

vinceaperri:
- Tests
    - Artifacts
        - core-efi VHDX for arm64 and amd64 in the image cache, taken from the Alpha 2 release of Azure Linux 4.0
        - offline RPMs taken from the Alpha 2 container release at azlpubstagingacroxz2o4gw.azurecr.io
    - Functional tests
        - Separate azl3/2 workflow jobs from azl4
        - Tests are skipped where Azure Linux 4.0 fails, but the test is still updated to work for Azure Linux 4.0 when support is soon added.
    - VM tests
        - Added a workflow step for osmodifer run
        - Integrated azl4 with existing imagecustomizer run
- Documentation
    - Add Azure Linux 4.0 to the distribution support page.


test/vmtests/vmtests/osmodifier/test_osmodifier.py
```diff
@@ -36,7 +38,13 @@ def setup_vm_with_osmodifier(
     libvirt_conn: libvirt.virConnect,
     session_close_list: List[Closeable],
 ) -> Tuple[SshClient, Path, Path]:
-    config_path = TEST_CONFIGS_DIR.joinpath("osmodifier-vm-config.yaml")
+    if distro_id == "azurelinux" and version_id == "4.0":
```
liulanze, line 41: This first `if` skips for AZL4, so the second `if` block that would assign the AZL4 config path is unreachable. Also, when AZL4 support is eventually added and the skip is removed, the else branch will fall back to the AZL3 config I think which is wrong.

test/vmtests/vmtests/osmodifier/test_osmodifier.py
```diff
@@ -258,25 +266,29 @@ def test_user_creation_config(
     result.check_exit_code()
 
     output = result.stdout.strip()
-    assert "sudo" in output, f"'sudo' not found in groups: {output}"
+    assert "wheel" in output, f"'sudo' not found in groups: {output}"
```
liulanze: nit - `assert "wheel" in output, f"'wheel' not found in groups: {output}"`

toolkit/tools/pkg/imagecustomizerlib/customizepackages_test.go
```diff
@@ -422,20 +507,54 @@ func TestCustomizeImagePackagesDiskSpace(t *testing.T) {
 	assert.ErrorContains(t, err, "failed to install packages ([gcc])")
 }
 
+// installPackageDiskSpaceConfigFile returns the install-package-disk-space test config file appropriate for the
+// given base image version (azl3 vs azl4) and host architecture.
+func installPackageDiskSpaceConfigFile(t *testing.T, baseImageInfo testBaseImageInfo) string {
```
liulanze, line 512: How is AZL2 handled here?

.github/workflows/tests-vmtests-osmodifier.yml
```diff
@@ -180,6 +185,25 @@ jobs:
...
             INPUT_IMAGE="${INPUT_IMAGE}" \
+            DISTRO_ID="azurelinux" \
+            VERSION_ID="3.0" \
+            SSH_PRIVATE_KEY_FILE=~/.ssh/id_ed25519
+
+      - name: Run tests (AZL4)
+        env:
+          IMAGE_CUSTOMIZER_CONTAINER_TAG: ${{ steps.importContainer.outputs.containerTag }}
+          OSMODIFIER_BIN: ${{ steps.osmodifier_bin.outputs.path }}
+          INPUT_IMAGE: ${{ steps.setup_image.outputs.azl4_path  }}
```
liulanze: nit - double space before }}

liulanze (changes requested):

toolkit/tools/pkg/imagecustomizerlib/customizepackages_test.go
vinceaperri: Added

test/vmtests/vmtests/osmodifier/test_osmodifier.py
vinceaperri: So the first if block is a temporary measure until I am able to add the changes in to remove this. It will be removed once osmodifier supports azl4. 

The second block does correctly fall back to azl3 config, since that is the only other distro-version that osmodifier is tested against. The config file used is just a renamed from osmodifier-vm-config.yaml. I added the -azl3 suffix to better distinguish it and to recognize how the packages declared in this YAML are actually azl3-specific, not applicable to ubuntu or fedora.

liulanze (approved):

PR #732, opened 2026-05-08
azl4: Support distro-aware grub stub assets and multiple EFI stub dirs

vinceaperri:
Plumb per-distro grub assets and EFI stub install directories through `installEfiBootloader` so Azure Linux 4.0 can use its own `/etc/default/grub` template and install stub `grub.cfg` files under `EFI/fedora` and `EFI/BOOT` instead of `boot/grub2`.


liulanze (commented):
LGTM overall, but could we add an integration test verifying the stub grub.cfg files land in the expected ESP locations for both distros? This PR introduces new install layout and adds new asset templates that we may want to have test coverage.

toolkit/tools/internal/resources/resources.go
```diff
@@ -9,8 +9,11 @@ import (
 
 const (
 	// Assets
-	AssetsGrubCfgFile = "assets/grub2/grub.cfg"
-	AssetsGrubDefFile = "assets/grub2/grub"
+	AssetsGrubCfgFile      = "assets/grub2/grub.cfg"
```
cwize1: I suspect different distros could have different files. So, I would make the distro/distro-version a directory instead of a filename suffix.

liulanze (approved):
lgtm

PR #742, opened 2026-05-21
fix: osmodifier: error when 'root=' is missing from non-recovery boot entry

vinceaperri:
In modifydefaultgrub.go, extractValuesFromGrubConfig previously returned an empty rootDevice when the non-recovery boot entry's kernel command line did not contain a 'root=' argument. SetRootDevice would then silently set GRUB_DEVICE with an empty value into /etc/default/grub, producing an unbootable default grub config.

Return an explicit error so the failure surfaces immediately to the caller instead of being papered over.


liulanze (approved):
LGTM

PR #757, opened 2026-06-03
support converting ACL images to COSI format

bfjelds:
## Summary
 
 Enables `imagecustomizer convert --output-image-format cosi` for Azure Container Linux (ACL) images. Previously this failed because ACL images lack `rpm`/`tdnf` CLI binaries and have a read-only rootfs.
 
 ## Problem
 
 Three issues blocked ACL → COSI conversion:
 
 1. **No `rpm` binary** — `GetAllPackagesFromChroot` runs `rpm -qa` inside the chroot to build the COSI package manifest. ACL images don't ship rpm.
 2. **No `tdnf` binary** — `DetectBootloaderType` probes for the `systemd-boot` package via `tdnf info`. ACL images don't ship tdnf, so detection silently fails and errors with `"unknown bootloader"`.
 3. **Read-only rootfs** — The convert path sets up default chroot mounts (`/dev`, `/proc`, `/sys`), which requires creating directories on the rootfs. ACL's btrfs+dm-verity rootfs is read-only, so `MkdirAll` fails.
 
 ## Changes
 
 **`distrohandler_acl.go`**
 - `DetectBootloaderType`: Return `systemd-boot` directly. ACL always uses systemd-boot + UKI (the handler already blocks GRUB in three other methods).
 - `ValidateUkiDependencies`: Return nil. ACL always ships systemd-boot.
 - `IsPackageInstalled`: Use host `rpm -q --root --dbpath` instead of in-chroot `tdnf info`.
 - `GetAllPackagesFromChroot`: Use host `rpm -qa --root --dbpath`, gracefully returning an empty list if the RPM DB doesn't exist (common for minimal ACL images).
 
 **`customizepackages_rpm.go`**
 - Add `getAllPackagesFromChrootRpmViaHost()` and `isPackageInstalledViaHostRpm()` helpers that query the image's RPM DB using the host's rpm binary with `--root` and explicit `--dbpath /var/lib/rpm`.
 - Extract `parseRpmQueryOutput()` shared parser (also handles empty output).
 
 **`artifactsinputoutput.go`**
 - Set `includeDefaultMounts=false` in `prepareImageConversionData`. The convert path only reads files and queries the RPM DB — it doesn't execute programs in the chroot, so `/dev`/`/proc`/`/sys` mounts are unnecessary.
 
 ## Testing
 
 - Unit tests for `parseRpmQueryOutput` (valid, empty, malformed input) and ACL handler hardcoded returns (`DetectBootloaderType`, `ValidateUkiDependencies`).
 - Manual: converted an ACL VHD to COSI format successfully.
 - No changes to Azure Linux 2/3/4 or Fedora handlers — existing behavior unaffected.

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
@@ -89,9 +89,15 @@ func (d *aclDistroHandler) IsPackageInstalled(imageChroot safechroot.ChrootInter
 
 func (d *aclDistroHandler) GetAllPackagesFromChroot(imageChroot safechroot.ChrootInterface) ([]OsPackage, error) {
 	// ACL images do not ship the rpm CLI, so the standard in-chroot
-	// rpm -qa query would fail. Use the host's rpm binary with --root
-	// to query the image's RPM database instead.
-	return getAllPackagesFromChrootRpmViaHost(imageChroot)
+	// rpm -qa query would fail. Try the host's rpm binary with --root
+	// to query the image's RPM database. If the DB doesn't exist (common
+	// for minimal ACL images that strip the RPM DB), return an empty list.
+	packages, err := getAllPackagesFromChrootRpmViaHost(imageChroot)
+	if err != nil {
+		logger.Log.Warnf("Could not query RPM DB for ACL image, returning empty package list: %v", err)
+		return nil, nil
+	}
+	return packages, nil
 }
 
 func (d *aclDistroHandler) DetectBootloaderType(imageChroot safechroot.ChrootInterface) (BootloaderType, error) {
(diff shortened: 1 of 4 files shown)


toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
```diff
@@ -81,21 +81,34 @@ func (d *aclDistroHandler) ManagePackages(ctx context.Context, buildDir string,
...
 func (d *aclDistroHandler) GetAllPackagesFromChroot(imageChroot safechroot.ChrootInterface) ([]OsPackage, error) {
-	return getAllPackagesFromChrootRpm(imageChroot)
+	// ACL images do not ship the rpm CLI, so the standard in-chroot
+	// rpm -qa query would fail. Try the host's rpm binary with --root
+	// to query the image's RPM database. If the DB doesn't exist (common
+	// for minimal ACL images that strip the RPM DB), return an empty list.
+	packages, err := getAllPackagesFromChrootRpmViaHost(imageChroot)
+	if err != nil {
+		logger.Log.Warnf("Could not query RPM DB for ACL image, returning empty package list: %v", err)
+		return nil, nil
```
bfjelds, line 98: seems a little overkill to provide getAllPackagesFromChrootRpmViaHost AND be ok with returning an empty set.  if empty set is acceptable for ACL, maybe just do that?

cwize1:
Fixed by #771 and #775

PR #759, opened 2026-06-04
Fix LiveISO build handle of kernel files.

cwize1:
The LiveISO build needs the ability to split files relating to the kernel from other files. Currently, this logic is a basic regex that detects if the filename contains a AZL3 formatted kernel version string. However, this is not very robust and fails if the image is using a 3rd party kernel (e.g. LSG NPI).

This change fixes this problem by collecting a list of all the kernel versions and then detecting which files contain one of the kernel version strings. This should make the logic more robust, including for other distros.

In addition, this change cleans-up/merges some string constants.

---

### **Checklist**

- [x] Tests added/updated


toolkit/tools/pkg/imagecustomizerlib/liveosisoutils.go
```diff
@@ -43,9 +43,8 @@ const (
 
 	initrdImage = "initrd.img"
 
-	// In vhd(x)/qcow images, the kernel is named 'vmlinuz-<version>'.
-	// In the ISO image, the kernel is named 'vmlinuz'.
-	vmLinuzPrefix     = "vmlinuz"
+	// In vhd(x)/qcow/iso images, the kernel is named 'vmlinuz-<version>'.
+	vmLinuzPrefix     = "vmlinuz-"
 	initramfsPrefix   = "initramfs-"  // Azl3
 	initrdPrefix      = "initrd.img-" // CBL-Mariner
```
vinceaperri: Since this PR drops the AZL3-specific regex, are the `// Azl3` and `// CBL-Mariner` labels still accurate? Both prefixes are also used by other distros.

toolkit/tools/pkg/imagecustomizerlib/liveosisoartifactstore.go
```diff
@@ -112,26 +113,37 @@ func getSELinuxMode(buildDir string, imageChroot *safechroot.Chroot, distroHandl
...
-func storeIfKernelSpecificFile(filesStore *IsoFilesStore, targetPath string) bool {
+func storeIfKernelSpecificFile(filesStore *IsoFilesStore, targetPath string, kernelVersions []string) bool {
 	scheduleAdditionalFile := true
 
 	baseFileName := filepath.Base(targetPath)
 
-	matches := kernelVersionRegEx.FindStringSubmatch(baseFileName)
-	if len(matches) <= 1 {
+	kernelVersion, hasKernelVersion := sliceutils.FindValueFunc(kernelVersions, func(kernelVersion string) bool {
+		return strings.Contains(targetPath, kernelVersion)
```
vinceaperri: Should this match `baseFileName` rather than `targetPath`?

gmileka (dismissed):

toolkit/tools/pkg/imagecustomizerlib/liveosisoartifactstore.go
liulanze: +1

cwize1: Fixed

toolkit/tools/pkg/imagecustomizerlib/liveosisoutils.go
cwize1: The labels are still accurate. But I've updated them to include other distros.

vinceaperri (approved):

liulanze (approved):

PR #762, opened 2026-06-05
Run grub2-install inside the install chroot to use target image's modules

vinceaperri:
## Problem

Image Customizer installs the legacy (BIOS) bootloader by running `grub2-install` on the **build host**, which copies grub modules from the *host's* `grub2-pc` into the target image. When the host's grub significantly differs from the target's grub, the copied modules don't match the `grub.cfg` the target's own `grub2-mkconfig` produces, and the image fails to boot. Concretely, the AZL3-based IC container building an AZL4 image produces a disk whose `grub.cfg` does `insmod blscfg` while `/boot/grub2/i386-pc/` was populated from AZL3 (no BLS modules), so it errors with `file '.../blscfg.mod' not found` on boot.

## Fix

Install the legacy bootloader from the **target's own `grub2-install`, run inside the install chroot**, so the bootloader and its modules always come from the target. This requires the `grub2-install` command (and its `i386-pc` modules) in the target image — added via `os.packages.install`:

- Fedora, Azure Linux: `grub2-pc`
- Ubuntu: `grub2-common`

Behavior when the target does not ship them:

| Target distro version | Behavior |
|---|---|
| `grub2-install` present (any version) | Run `grub2-install` in the chroot; modules come from the target. |
| AZL2 / AZL3, not present | Fall back to the host's `grub2-install` with a **warning**. This preserves existing configs but is **deprecated** and will be removed in a future version. |
| All other distro versions, not present | Fail with a clear error directing the user to install the package. |

## Container: explicitly install `systemd-udev`

The IC container's host-side disk setup calls `udevadm settle` (`waitForDevicesToSettle`). `udevadm` is provided by the **`systemd-udev`** package, which until now arrived only transitively via `grub2`. `systemd-udev` is now listed explicitly in `imagecustomizer.Dockerfile` so `udevadm` can never silently disappear if the grub2 dependency changes.

## Testing

- **Functional:** `TestCustomizeImagePartitionsLegacy` covers the chroot path (config installs `grub2-pc`); `TestCustomizeImagePartitionsLegacy_TargetMissingPackages` removes the packages and asserts the AZL2/AZL3 host fallback still builds a valid legacy image; the AZL4 no-`grub2-pc` config asserts the error.
- **VM tests:** `test_legacy_bootloader_reset_azl{2,3,4}` boot a hard-reset legacy image built with `grub2-pc` (chroot path); `test_legacy_bootloader_reset_fallback_azl{2,3}` boot one built without it (host fallback).

## Docs

`storage.md` documents the `bootType: legacy` package requirement, the per-distro package names, and the deprecated AZL2/3 host fallback.

### DIFF
--- a/docs/imagecustomizer/api/configuration/storage.md
+++ b/docs/imagecustomizer/api/configuration/storage.md
@@ -54,7 +54,7 @@ Supported options:
 
   Install the following packages (using the `os.packages.install` API):
 
-  - Fedora, Azure Linux 4.0: `grub2-tools`, `grub-pc-modules`
+  - Fedora, Azure Linux 4.0: `grub2-tools`, `grub2-pc-modules`
   - Azure Linux 2.0, 3.0: `grub2`, `grub2-pc`
 
   For Azure Linux 2.0 and 3.0, if the `grub2-install` command is not available in the
--- a/docs/imagecustomizer/api/configuration/storage.md
+++ b/docs/imagecustomizer/api/configuration/storage.md
@@ -52,10 +52,10 @@ Supported options:
 
   The `grub2-install` command must be installed in the target image.
 
-  Install the following package (using the `os.packages.install` API):
+  Install the following packages (using the `os.packages.install` API):
 
-  - Fedora, Azure Linux: `grub2-pc`
-  - Ubuntu: `grub2-common`
+  - Fedora, Azure Linux 4.0: `grub2-tools`, `grub-pc-modules`
+  - Azure Linux 2.0, 3.0: `grub2`, `grub2-pc`
 
   For Azure Linux 2.0 and 3.0, if the `grub2-install` command is not available in the
   target OS, then the `grub2-install` command from the container (or build host) is
--- a/docs/imagecustomizer/api/configuration/storage.md
+++ b/docs/imagecustomizer/api/configuration/storage.md
@@ -50,6 +50,15 @@ Supported options:
 
 - `legacy`: Support booting from BIOS firmware.
 
+  The `grub2-install` command must be installed in the target image.
+
+  Install the following package (using the `os.packages.install` API):
+
+  - Fedora, Azure Linux: `grub2-pc`
+  - Ubuntu: `grub2-common`
+
+  For Azure Linux 2.0 and 3.0, if the `grub2-install` command is not available in the target OS, then the `grub2-install` command from the container (or build host) is used instead. However, this behavior is deprecated and will be removed in a future version.
+
   When this option is specified, the partition layout must contain a partition with the
   `bios-grub` flag.
 
--- a/docs/imagecustomizer/api/configuration/storage.md
+++ b/docs/imagecustomizer/api/configuration/storage.md
@@ -57,7 +57,10 @@ Supported options:
   - Fedora, Azure Linux: `grub2-pc`
   - Ubuntu: `grub2-common`
 
-  For Azure Linux 2.0 and 3.0, if the `grub2-install` command is not available in the target OS, then the `grub2-install` command from the container (or build host) is used instead. However, this behavior is deprecated and will be removed in a future version.
+  For Azure Linux 2.0 and 3.0, if the `grub2-install` command is not available in the
+  target OS, then the `grub2-install` command from the container (or build host) is
+  used instead. However, this behavior is deprecated and will be removed in a future
+  version.
 
   When this option is specified, the partition layout must contain a partition with the
   `bios-grub` flag.
--- a/toolkit/tools/imagegen/installutils/installutils.go
+++ b/toolkit/tools/imagegen/installutils/installutils.go
@@ -106,8 +106,12 @@ const (
 )
 
 const (
-	rootMountPoint = "/"
-	bootMountPoint = "/boot"
+	rootMountPoint   = "/"
+	bootMountPoint   = "/boot"
+	grub2BootDir     = "/boot/grub2"
+	grub2InstallName = "grub2-install"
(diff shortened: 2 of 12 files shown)


cwize1 (dismissed):

toolkit/tools/pkg/imagecustomizerlib/testdata/legacyboot-config.yaml
```diff
@@ -21,5 +21,12 @@ storage:
       path: /
 
 os:
+  packages:
+    install:
+      # grub2-pc must be installed in the target image because grub2-install
+      # runs inside the install chroot, and the core-efi base test images do not
+      # ship with it.
+      - grub2-pc
```
cwize1, line 29: This represents a breaking change. So, it is something we need to be intentional about.

cwize1 (changes requested):

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -971,29 +968,10 @@ func installLegacyBootloader(installChroot *safechroot.Chroot, bootDevPath strin
 			return
 		}
 	}
-	installBootDir := filepath.Join(installChroot.RootDir(), bootDir)
-	grub2InstallBootDirArg := fmt.Sprintf("%s=%s", bootDirArg, installBootDir)
 
-	installName := grub2InstallName
```
liulanze, line 977: Now we removed this, the falling back logic to the latter for Debian-style targets is removed. if the legacy boot customization for Ubuntu or Debian base images can be now failed with "exec not found". This seems a regression for Debian/Ubuntu.

vinceaperri: Yes, will add back.

docs/imagecustomizer/api/configuration/storage.md
```diff
@@ -53,6 +53,12 @@ Supported options:
   When this option is specified, the partition layout must contain a partition with the
   `bios-grub` flag.
 
+  When `grub2-pc` is installed in the target image, the legacy bootloader is installed by
```
cwize1: It might be good to cleanly separate what the user needs to do from the explanation. This will make it easier to read it at a glance.

For example:

```Markdown
The `grub2-install` or `grub-install` command must be installed in the target image.

Install the following package (using the `os.packages.install` API):

- Fedora, Azure Linux: `grub2-pc`
- Ubuntu: `grub2-common`

For Azure Linux 2.0 and 3.0, if the `grub2-install` command is not available in the target OS, then the `grub2-install` command from the container (or build host) is used instead. However, this behavior is deprecated and will be removed in a future version.
```

vinceaperri: Done (note: grub-install isn't used in the target image, only grub2-install).

cwize1: Ubuntu calls the command `grub-install`. So, you'll need to fix this logic for Ubuntu. But feel free to do this in a subsequent change.

cwize1 (dismissed):

docs/imagecustomizer/api/configuration/storage.md
```diff
@@ -50,6 +50,15 @@ Supported options:
...
 - `legacy`: Support booting from BIOS firmware.
 
+  The `grub2-install` command must be installed in the target image.
+
+  Install the following package (using the `os.packages.install` API):
+
+  - Fedora, Azure Linux: `grub2-pc`
+  - Ubuntu: `grub2-common`
+
+  For Azure Linux 2.0 and 3.0, if the `grub2-install` command is not available in the target OS, then the `grub2-install` command from the container (or build host) is used instead. However, this behavior is deprecated and will be removed in a future version.
```
cwize1: nit: Line wrap at 88

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -952,28 +959,94 @@ func InstallBootloader(installChroot *safechroot.Chroot, encryptEnabled bool, bo
...
+			Execute()
+		if err != nil {
+			return
+		}
+	case allowHostGrubInstallFallback:
+		if err = installLegacyBootloaderFromHost(installChroot, bootDevPath, missingPackages); err != nil {
+			return
+		}
+	default:
+		package_s := "package"
```
cwize1: I don't know if it is worth having `package_s`. This is a dev tool, not an app. So, I don't think anyone will mind if there is unnecessary pluralization.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -987,20 +1060,33 @@ func installLegacyBootloader(installChroot *safechroot.Chroot, bootDevPath strin
...
 
-	installGrub2BootDir := filepath.Join(installChroot.RootDir(), grub2BootDir)
-	err = shell.ExecuteLive(squashErrors, "chmod", "-R", "go-rwx", installGrub2BootDir)
-	return
+	installBootDir := filepath.Join(installChroot.RootDir(), "/boot")
+	bootDirArg := fmt.Sprintf("%s=%s", "--boot-directory", installBootDir)
+	return shell.ExecuteLive(false /*squashErrors*/, installName, "--target=i386-pc", bootDirArg, bootDevPath)
+}
+
+func grubInstallCommandExistsInChroot(installChroot *safechroot.Chroot, command string) bool {
```
cwize1: #775 added a `shell.LookPathChroot` function, which you could use here.

cwize1 (changes requested):

cwize1 (approved):

PR #763, opened 2026-06-05
docs: Add Host Kernel Modules concepts page

vinceaperri:
Document the kernel modules Image Customizer requires the host kernel to provide (loop, ext4, vfat, btrfs, xfs for core functionality; dm_mod and dm_verity for verity images; squashfs and iso9660 for ISO / Live OS work).

Also adds a link to the new page from the concepts landing page and from the prerequisites section of both the container and binary quick-start guides, since the container shares the host kernel and the requirement applies to both usage paths.

### DIFF
--- a/docs/imagecustomizer/concepts/README.md
+++ b/docs/imagecustomizer/concepts/README.md
@@ -16,3 +16,4 @@ following scenarios:
 * ISO: Live-ISO creation. More details in [ISO Support](./iso.md)
 * PXE: Creating a PXE bootable image. More details in [PXE Support](./pxe.md)
 * Verity protected images: [Guidelines for creating a verity-protected root filesystem](./verity/verity.md)
+* Host kernel modules: [Kernel modules the host must provide](./host-kernel-modules.md)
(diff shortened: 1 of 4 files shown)


toolkit/tools/imagecustomizer/container/imagecustomizer.Dockerfile
```diff
@@ -12,7 +12,7 @@ ENV AZURE_MONITOR_CONNECTION_STRING=${AZ_MON_CONN_STR}
 RUN tdnf update -y && \
    tdnf install -y azurelinux-repos-cloud-native && \
    tdnf update -y && \
-   tdnf install -y qemu-img rpm coreutils util-linux systemd openssl \
+   tdnf install -y qemu-img rpm coreutils util-linux systemd systemd-udev openssl \
```
cwize1: This is from the other PR?

cwize1 (changes requested):

toolkit/tools/imagecustomizer/container/imagecustomizer.Dockerfile
vinceaperri: Yes, removed.

cwize1 (approved):

liulanze (approved):

PR #769, opened 2026-06-09
Bump max supported version of mkfs.xfs.

cwize1:
---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/imagegen/diskutils/filesystem.go
+++ b/toolkit/tools/imagegen/diskutils/filesystem.go
@@ -412,7 +412,9 @@ var (
 
 	// The maximum version of mkfs.xfs that is currently supported.
 	// This is used to prevent issues with newer versions of mkfs.xfs default enabling new features.
-	maxMkfsXfsVersion = version.Version{6, 15}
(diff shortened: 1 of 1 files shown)


toolkit/tools/imagegen/diskutils/filesystem.go
```diff
@@ -661,7 +663,8 @@ func getXfsFileSystemOptions(hostKernelVersion version.Version, options fileSyst
 		return nil, err
 	}
 
-	if mkfsXfsVersion.Gt(maxMkfsXfsVersion) {
+	mkfsXfsVersionMinorVer := mkfsXfsVersion[:min(2, len(mkfsXfsVersion))]
```
liulanze, line 666: For example `7.0.0` compares to `6.15.0` vs `7.0` compares to `6.15`, and we choose the latter correct?

liulanze (approved):

toolkit/tools/imagegen/diskutils/filesystem.go
cwize1: Yeah, it is so that `7.0.1` is treated as ok when `mkfsXfsVersion` is `7.0`.

PR #771, opened 2026-06-09
Allow ACL image to be mounted as read-only.

cwize1:
The ACL image doesn't include placeholder directories for the standard linux directories (e.g. /dev) or even its own mounts (e.g. /oem). This causes problems for IC when mounting the ACL as read-only because some of the directories are missing for the mounts.

To workaround this issue, an overlay is used to create the placeholder directories underneath the root partition. That way, IC has directories it can attach the mounts to.

To facilitate this workaround, IC needs to know what the target distro is earlier in the process: in-between partition discovery and mounting.

Also, the DistroHandler instance is now passed between more components, instead of being (re-)discovered in multiple places.

Also, ensure there is a single place where all the potential `os-release` paths are listed.

Also, log when a DistroHandler instance is created and which one. This could help with debugging issues relating to the picked DistroHandler implementation.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/imageutils.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/imageutils.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io/fs"
 	"path/filepath"
+	"slices"
 	"sort"
 	"strings"
 
@@ -16,12 +17,15 @@ import (
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/imagegen/configuration"
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/imagegen/diskutils"
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/imagegen/installutils"
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/envfile"
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/file"
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/imageconnection"
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/logger"
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/safechroot"
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/safemount"
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/shell"
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/sliceutils"
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/targetos"
 	"go.opentelemetry.io/otel"
 	"golang.org/x/sys/unix"
 )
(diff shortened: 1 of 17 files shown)


toolkit/tools/pkg/imagecustomizerlib/selinuxpolicyoutput.go
```diff
@@ -49,7 +49,7 @@ func readSelinuxType(chrootDir string) (string, error) {
 }
 
 func outputSelinuxPolicy(ctx context.Context, outputDir string, buildDir string, buildImage string,
-	partitionsLayout []fstabEntryPartNum,
+	partitionsLayout []fstabEntryPartNum, distroHander DistroHandler,
```
vinceaperri: `distroHandler`

toolkit/tools/pkg/imagecustomizerlib/imageutils.go
```diff
@@ -447,3 +478,77 @@ func clearBtrfsReadOnlyProperties(imageConnection *imageconnection.ImageConnecti
...
+			uintptr(fstabEntry.VfsOptions), fstabEntry.FsOptions, true /*makeAndDeleteDir*/)
+		if err != nil {
+			return targetos.TargetOs{}, err
+		}
+		defer mount.Close()
+
+		osReleasePath := filepath.Join(mountDir, relativePath)
+		osReleaseFileExists, err := file.PathExists(osReleasePath)
+		if err != nil {
+			return targetos.TargetOs{}, fmt.Errorf("failed to check if os-release file exists (path='%s):\n%w",
```
vinceaperri: single quote is unterminated.

toolkit/tools/pkg/imagecustomizerlib/imageutils.go
```diff
@@ -447,3 +478,77 @@ func clearBtrfsReadOnlyProperties(imageConnection *imageconnection.ImageConnecti
...
+			if err != nil {
+				return targetos.TargetOs{}, err
+			}
+
+			continue
+		}
+
+		fields, err := envfile.ParseEnvFile(osReleasePath)
+		if err != nil {
+			return targetos.TargetOs{}, fmt.Errorf("failed to read os-relase file (path='%s):\n%w", osReleasePath,
```
vinceaperri: `os-release`, and the single quote is unterminated.

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -590,25 +598,89 @@ func partitionLayoutToMountPoints(layout []fstabEntryPartNum, diskPartitions []d
 
 		var mountPoint *safechroot.MountPoint
 		if fstabEntry.Target == "/" {
-			mountPoint = safechroot.NewPreDefaultsMountPoint(
+			mountPoint := safechroot.NewPreDefaultsMountPoint(
```
vinceaperri, line 600: Outer `var mountPoint` is only used by the else branch now, so this `:=` shadows it. Remove the outer declaration and use `:=` there too?

cwize1: Fixed

toolkit/tools/pkg/imagecustomizerlib/imageutils.go
cwize1: Fixed

toolkit/tools/pkg/imagecustomizerlib/imageutils.go
cwize1: Fixed

toolkit/tools/pkg/imagecustomizerlib/selinuxpolicyoutput.go
cwize1: Fixed

vinceaperri (approved):

PR #775, opened 2026-06-15
ACL: Don't fail when collecting the list of packages.

cwize1:
The ACL image commonly doesn't contain an RPM db file or even the rpm command. This causes a `convert` to COSI command to fail. To prevent this failure, just populate an empty package list in the COSI file when either the RPM db or rpm command are missing.

Also, hardcode the bootloader type in the COSI file to systemd-boot for the ACL image, since again it isn't possible to query the list of packages for ACL and ACL always uses systemd-boot.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/internal/shell/shell.go
+++ b/toolkit/tools/internal/shell/shell.go
@@ -158,6 +158,12 @@ func MustExecuteLive(command string, args ...string) {
 	}
 }
 
+// Tries to find a commnad within a chroot.
+// Returns an error iff. the command is not found.
+func LookPathChroot(command string, rootDir string) (string, error) {
+	return chrootLookPath(command, rootDir, chrootPathDirs)
+}
+
 func trackAndStartProcess(cmd *exec.Cmd, capabilities []uintptr) (err error) {
 	logger.Log.Debugf("Executing: %v", cmd.Args)
 
--- a/toolkit/tools/internal/shell/shell.go
+++ b/toolkit/tools/internal/shell/shell.go
@@ -158,7 +158,7 @@ func MustExecuteLive(command string, args ...string) {
 	}
 }
 
-// Tries to find a commnad within a chroot.
+// Tries to find a command within a chroot.
 // Returns an error iff. the command is not found.
 func LookPathChroot(command string, rootDir string) (string, error) {
 	return chrootLookPath(command, rootDir, chrootPathDirs)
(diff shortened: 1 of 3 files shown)


toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
```diff
@@ -100,12 +102,40 @@ func (d *aclDistroHandler) IsPackageInstalled(imageChroot safechroot.ChrootInter
...
+	_, err := shell.LookPathChroot("rpm", imageChroot.ChrootDir())
+	if err != nil {
+		// RPM command is not found.
+		return nil, nil
+	}
+
+	// Check if the rpm database is available.
+	exists, err := file.PathExists(filepath.Join(imageChroot.RootDir(), "/var/lib/rpm"))
+	if err != nil {
+		// RPM command is not found.
```
liulanze: RPMDB is not found?

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
```diff
@@ -100,12 +102,40 @@ func (d *aclDistroHandler) IsPackageInstalled(imageChroot safechroot.ChrootInter
...
+		return nil, fmt.Errorf("failed to check if rpm db exists:\n%w", err)
+	}
+
+	if !exists {
+		// RPM database doesn't exist.
+		return nil, nil
+	}
+
+	// Get the list of packages.
+	packages, err := getAllPackagesFromChrootRpm(imageChroot)
```
liulanze, line 128: Is this block can be:

```
return getAllPackagesFromChrootRpm(imageChroot)
```

toolkit/tools/internal/shell/shell.go
```diff
@@ -158,6 +158,12 @@ func MustExecuteLive(command string, args ...string) {
 	}
 }
 
+// Tries to find a commnad within a chroot.
```
liulanze: nit - command

liulanze (dismissed):

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
cwize1: It is better to have the error handling so that it is easier to add/change code in the future.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
cwize1: Fixed

toolkit/tools/internal/shell/shell.go
cwize1: Fixed

liulanze (approved):

PR #776, opened 2026-06-15
Add --tools-dir to customize, rename create's --tools-file to --tools-dir

liulanze:
ACL images do not include a package manager (tdnf/dnf) in the base image
because `/usr` is immutable and verity-protected. The `customize` subcommand
previously had no way to provide an external package manager, making package
operations on ACL images impossible.

This PR mirrors the `--tools-file` support that already exists on the `create`
subcommand, bringing the same capability to `customize`.

### DIFF
--- a/docs/imagecustomizer/api/cli/create.md
+++ b/docs/imagecustomizer/api/cli/create.md
@@ -129,7 +129,7 @@ dependencies) needed to install packages during image creation.
 For instructions on how to create this directory, see:
 [How to create the tools directory](../../how-to/create-tools-dir.md)
 
-Added in v1.2. Renamed from `--tools-file` in v1.5.
+Added in v1.5.
 
 ## --distro=DISTRO
 
--- a/docs/imagecustomizer/api/cli/create.md
+++ b/docs/imagecustomizer/api/cli/create.md
@@ -119,13 +119,15 @@ RPMs from an existing RPM repo (such as packages.microsoft.com). Using a cloned 
 
 Added in v1.2.
 
-## --tools-file=PATH
+## --tools-dir=DIRECTORY-PATH
 
 Required.
 
-Specifies the path to a tools file in `.tar.gz` format.
+Specifies the path to a directory that provides the tools (tdnf/dnf and their
+dependencies) needed to install packages during image creation.
 
-This file should contain the TDNF tar package (or an equivalent), which is used to manage package dependencies and facilitate installation workflows.
+For instructions on how to create this directory, see:
+[How to create the tools directory](../../how-to/create-tools-dir.md)
 
 Added in v1.2.
 
--- a/docs/imagecustomizer/api/cli/create.md
+++ b/docs/imagecustomizer/api/cli/create.md
@@ -129,7 +129,7 @@ dependencies) needed to install packages during image creation.
 For instructions on how to create this directory, see:
 [How to create the tools directory](../../how-to/create-tools-dir.md)
 
-Added in v1.2.
+Added in v1.2. Renamed from `--tools-file` in v1.5.
 
 ## --distro=DISTRO
 
--- a/docs/imagecustomizer/how-to/create-image.md
+++ b/docs/imagecustomizer/how-to/create-image.md
@@ -174,23 +174,19 @@ As such, the example configuration files below must specify `create` in the [pre
    For documentation on the supported configuration options, see:
    [Supported configuration](../api/configuration/configuration.md)
 
-2. Download the tools file.
+2. Create the tools directory.
 
-   The tools file contains bootstrap utilities needed to create an image from scratch.
+   The tools directory contains the bootstrap utilities (tdnf and its dependencies)
+   needed to install packages during image creation.
 
-   To create a tools file for a new Azure Linux 3.0:
+   For step-by-step instructions, see:
+   [How to create the tools directory](../how-to/create-tools-dir.md)
 
-    ```bash
-    ./toolkit/tools/internal/testutils/testrpms/create-tools-file.sh \
-      "mcr.microsoft.com/azurelinux/base/core:3.0" "$HOME/staging/azure-linux-3.0-tools.tar.gz"
-    ```
+   For Azure Linux 3.0, the tools directory should be created from the
+   `mcr.microsoft.com/azurelinux/base/core:3.0` container image.
 
-   To create a tools file for a new Fedora 42:
-
-    ```bash
-    ./toolkit/tools/internal/testutils/testrpms/create-tools-file.sh \
-      "quay.io/fedora/fedora:42" "$HOME/staging/fedora-42-tools.tar.gz"
-    ```
+   For Fedora 42, the tools directory should be created from the
+   `quay.io/fedora/fedora:42` container image.
 
 3. Create a repository configuration file and, if needed, a GPG key file.
 
@@ -246,7 +242,7 @@ As such, the example configuration files below must specify `create` in the [pre
       mcr.microsoft.com/azurelinux/imagecustomizer:latest create \
         --distro azurelinux \
         --distro-version 3.0 \
-        --tools-file /mnt/staging/azure-linux-3.0-tools.tar.gz \
+        --tools-dir /mnt/staging/tools-azurelinux-3.0 \
         --rpm-source /mnt/staging/azure-linux-3.0-rpms.repo \
         --config-file /mnt/staging/config-azl3.yaml \
         --build-dir /build \
@@ -265,7 +261,7 @@ As such, the example configuration files below must specify `create` in the [pre
       mcr.microsoft.com/azurelinux/imagecustomizer:latest create \
         --distro fedora \
         --distro-version 42 \
-        --tools-file /mnt/staging/fedora-42-tools.tar.gz \
+        --tools-dir /mnt/staging/tools-fedora-42 \
         --rpm-source /mnt/staging/fedora-42-rpms.repo \
         --config-file /mnt/staging/config-fedora.yaml \
         --build-dir /build \
@@ -302,8 +298,8 @@ As such, the example configuration files below must specify `create` in the [pre
 
     - `--distro-version 42`: Specify the Fedora version to create the image for.
 
-    - `--tools-file /mnt/staging/fedora-42-tools.tar.gz`: Use the host's
-      `$HOME/staging/fedora-42-tools.tar.gz` file as the bootstrapping tools file.
+    - `--tools-dir /mnt/staging/tools-fedora-42`: Use the host's
+      `$HOME/staging/tools-fedora-42` directory as the bootstrapping tools directory.
 
     - `--rpm-source /mnt/staging/fedora-42-rpms.repo`: Use the host's
       `$HOME/staging/fedora-42-rpms.repo` file for installing packages.
--- a/docs/imagecustomizer/how-to/create-image.md
+++ b/docs/imagecustomizer/how-to/create-image.md
@@ -182,12 +182,6 @@ As such, the example configuration files below must specify `create` in the [pre
    For step-by-step instructions, see:
    [How to create the tools directory](../how-to/create-tools-dir.md)
 
-   For Azure Linux 3.0, the tools directory should be created from the
-   `mcr.microsoft.com/azurelinux/base/core:3.0` container image.
-
-   For Fedora 42, the tools directory should be created from the
-   `quay.io/fedora/fedora:42` container image.
-
 3. Create a repository configuration file and, if needed, a GPG key file.
 
    The repository configuration file (`.repo` file) tells the Image Customizer where to download RPM packages from.
--- a/docs/imagecustomizer/how-to/create-tools-dir.md
+++ b/docs/imagecustomizer/how-to/create-tools-dir.md
@@ -1,19 +1,18 @@
 ---
-title: Create Tools Directory
+title: Tools Directory
 parent: How To
 nav_order: 9
 has_toc: false
 ---
 
-# Creating the Tools Directory
+# Tools Directory
 
-The `--tools-dir` flag used by the Image Customizer
-[create](../api/cli/create.md) and [customize](../api/cli/customize.md) subcommands
-requires a directory that contains a package manager (tdnf or dnf) and its runtime
-dependencies.
+Both the `create` and `customize` subcommands of the Image Customizer accept a
+`--tools-dir` flag. The directory it points to must contain a package manager
+(tdnf or dnf) and its runtime dependencies.
 
-This directory is used as a bootstrap environment to install packages into the target
-image. It is not added to the image itself.
+Image Customizer uses this directory as a chroot to run utilities needed during
+image customization.
 
 ## Prerequisites
 
--- a/docs/imagecustomizer/how-to/create-tools-dir.md
+++ b/docs/imagecustomizer/how-to/create-tools-dir.md
@@ -1,0 +1,107 @@
+---
+title: Create Tools Directory
+parent: How To
+nav_order: 9
+has_toc: false
+---
+
+# Creating the Tools Directory
+
+The `--tools-dir` flag used by the Image Customizer
+[create](../api/cli/create.md) and [customize](../api/cli/customize.md) subcommands
+requires a directory that contains a package manager (tdnf or dnf) and its runtime
+dependencies.
+
+This directory is used as a bootstrap environment to install packages into the target
+image. It is not added to the image itself.
+
+## Prerequisites
+
+- Linux host
+- Docker (or equivalent container engine) installed
+
+## How It Works
+
+The tools directory is created by exporting the filesystem of a container image that
+already contains the package manager you need. Image Customizer then uses this directory
+as a chroot environment, mounting the target image inside it at `/_imageroot` and
+running `tdnf --installroot=/_imageroot`.
+
+## Instructions
+
+### Azure Linux 3.0
+
+The Azure Linux base core container includes tdnf and all its dependencies.
+
+```bash
+# Pull the container image
+docker pull mcr.microsoft.com/azurelinux/base/core:3.0
+
+# Create a temporary container (does not start it)
+docker create --name ic-tools-azurelinux-3.0 mcr.microsoft.com/azurelinux/base/core:3.0
+
+# Export the container filesystem and extract it into a directory
+mkdir -p "$HOME/staging/tools-azurelinux-3.0"
+docker export ic-tools-azurelinux-3.0 | tar -x -C "$HOME/staging/tools-azurelinux-3.0"
+
+# Remove the temporary container
+docker rm ic-tools-azurelinux-3.0
+```
+
+The tools directory is now at `$HOME/staging/tools-azurelinux-3.0`.
+
+### Fedora 42
+
+The Fedora base container includes dnf and all its dependencies.
+
+```bash
+# Pull the container image
+docker pull quay.io/fedora/fedora:42
+
+# Create a temporary container (does not start it)
+docker create --name ic-tools-fedora-42 quay.io/fedora/fedora:42
+
+# Export the container filesystem and extract it into a directory
+mkdir -p "$HOME/staging/tools-fedora-42"
+docker export ic-tools-fedora-42 | tar -x -C "$HOME/staging/tools-fedora-42"
+
+# Remove the temporary container
+docker rm ic-tools-fedora-42
+```
+
+The tools directory is now at `$HOME/staging/tools-fedora-42`.
+
+## Reusing the Tools Directory
+
+The tools directory can be reused across multiple Image Customizer runs as long as the
+container image it was created from has not changed. You do not need to recreate it each
+time.
+
+To pick up updates (e.g. a newer version of tdnf), pull the container image again and
+repeat the steps above to recreate the directory.
+
+## Using the Tools Directory
+
+Pass the tools directory to Image Customizer with the `--tools-dir` flag.
+
+For example, when running Image Customizer via Docker:
+
+```bash
+docker run \
+  --rm \
+  --privileged=true \
+  -v /dev:/dev \
+  -v "$HOME/staging:/mnt/staging:z" \
+  mcr.microsoft.com/azurelinux/imagecustomizer:latest customize \
+    --tools-dir /mnt/staging/tools-azurelinux-3.0 \
+    ...
+```
+
+Note that `$HOME/staging` is mounted into the container at `/mnt/staging`, so the tools
+directory path inside the container is `/mnt/staging/tools-azurelinux-3.0`.
+
+## See Also
+
+- [create subcommand](../api/cli/create.md)
+- [customize subcommand](../api/cli/customize.md)
+- [Creating a new image from scratch](../how-to/create-image.md)
--- a/toolkit/tools/imagecustomizer/main.go
+++ b/toolkit/tools/imagecustomizer/main.go
@@ -23,7 +23,7 @@ type CreateCmd struct {
 	BuildDir            string   `name:"build-dir" help:"Directory to run build out of." required:""`
 	ConfigFile          string   `name:"config-file" help:"Path of the image creation config file." required:""`
 	RpmSources          []string `name:"rpm-source" help:"Path to a RPM repo config file or a directory containing RPMs." required:""`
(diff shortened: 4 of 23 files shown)


docs/imagecustomizer/how-to/create-image.md
```diff
@@ -174,23 +174,19 @@ As such, the example configuration files below must specify `create` in the [pre
...
+   For step-by-step instructions, see:
+   [How to create the tools directory](../how-to/create-tools-dir.md)
 
-    ```bash
-    ./toolkit/tools/internal/testutils/testrpms/create-tools-file.sh \
-      "mcr.microsoft.com/azurelinux/base/core:3.0" "$HOME/staging/azure-linux-3.0-tools.tar.gz"
-    ```
-
-   To create a tools file for a new Fedora 42:
+   For Azure Linux 3.0, the tools directory should be created from the
```
cwize1: I would specify the requirements only in `create-tools-dir.md`, instead of duplicating some of the info here.

docs/imagecustomizer/how-to/create-tools-dir.md
```diff
@@ -0,0 +1,107 @@
...
+parent: How To
+nav_order: 9
+has_toc: false
+---
+
+# Creating the Tools Directory
+
+The `--tools-dir` flag used by the Image Customizer
+[create](../api/cli/create.md) and [customize](../api/cli/customize.md) subcommands
+requires a directory that contains a package manager (tdnf or dnf) and its runtime
```
cwize1: This sentence doesn't make much sense, particularly around `subcommands requires` .

toolkit/tools/internal/testutils/testutils.go
cwize1: Delete `GetDownloadedToolsFile`?

toolkit/tools/pkg/imagecustomizerlib/createimage_test.go
```diff
@@ -169,23 +169,21 @@ func TestCreateImageRawNoTar(t *testing.T) {
 }
 
 func testCreateImageRawNoTar(t *testing.T, name string, version string, configFile string) {
-	testutils.CheckSkipForCustomizeImageRequirements(t)
```
cwize1, line 174: Why was `testutils.CheckSkipForCustomizeImageRequirements(t)` removed?

toolkit/tools/pkg/imagecustomizerlib/customizeos.go
```diff
@@ -26,14 +27,22 @@ var ErrUkiKernelModified = NewImageCustomizerError("UKI:KernelModified",
...
+	toolsChroot *safechroot.Chroot,
 ) error {
 	var err error
 
 	imageChroot := imageConnection.Chroot()
 
 	buildTime := time.Now().Format(buildTimeFormat)
 
-	resolvConf, err := overrideResolvConf(imageChroot)
+	// When a tools chroot is used, network access (for package downloads) runs
```
cwize1: User scripts (e.g. `postCustomization`) might also want to access the network. So, we should do this for both chroots.

I would handle the `toolsChroot`'s resolve conf file when the chroot is initialized.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -673,8 +674,31 @@ func customizeImageHelper(ctx context.Context, rc *ResolvedConfig, partitionsCus
 
 	readOnlyVerity := rc.Storage.ReinitializeVerity != imagecustomizerapi.ReinitializeVerityTypeAll
 
+	// If a tools directory is provided, copy it to the build directory and initialize a tools chroot from it.
```
cwize1: I'd be inclined to just use the directory passed to us. That way, we can save time by avoiding copying files.

docs/imagecustomizer/how-to/create-tools-dir.md
```diff
@@ -0,0 +1,107 @@
...
+dependencies.
+
+This directory is used as a bootstrap environment to install packages into the target
```
cwize1: You don't need to handle this in this change but there are some other tools that would be good to run from the tools chroot when provided, including `useradd`, `setfiles`, etc.

Some customers may want to keep their OS as small as possible. So, if we can run the tools outside the target OS chroot, then those tools don't need to be installed in the target OS.

cwize1 (changes requested):

docs/imagecustomizer/api/cli/create.md
```diff
@@ -119,13 +119,15 @@ RPMs from an existing RPM repo (such as packages.microsoft.com). Using a cloned
...
 
-Specifies the path to a tools file in `.tar.gz` format.
+Specifies the path to a directory that provides the tools (tdnf/dnf and their
+dependencies) needed to install packages during image creation.
 
-This file should contain the TDNF tar package (or an equivalent), which is used to manage package dependencies and facilitate installation workflows.
+For instructions on how to create this directory, see:
+[How to create the tools directory](../../how-to/create-tools-dir.md)
 
 Added in v1.2.
```
vinceaperri: This renames the existing `--tools-file` flag to `--tools-dir`. Should this be 'Added in v1.5' (or note the rename) rather than 'Added in v1.2'?

toolkit/tools/imagecustomizer/main.go
```diff
@@ -22,7 +22,7 @@ type CreateCmd struct {
 	BuildDir            string   `name:"build-dir" help:"Directory to run build out of." required:""`
 	ConfigFile          string   `name:"config-file" help:"Path of the image creation config file." required:""`
 	RpmSources          []string `name:"rpm-source" help:"Path to a RPM repo config file or a directory containing RPMs." required:""`
-	ToolsTar            string   `name:"tools-file" help:"Path to tdnf/dnf worker tarball" required:""`
+	ToolsDir            string   `name:"tools-dir" help:"Path to a directory containing tdnf/dnf and its dependencies." required:""`
```
vinceaperri, line 26: This renames the existing `create --tools-file` flag to `--tools-dir`, a breaking change. Worth calling out in the PR title so we know to include it in the release notes.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
```diff
@@ -75,7 +75,11 @@ func (d *aclDistroHandler) ValidateConfig(rc *ResolvedConfig) error {
 			len(pkgs.Remove) > 0 || len(pkgs.RemoveLists) > 0 ||
 			len(pkgs.Update) > 0 || len(pkgs.UpdateLists) > 0 ||
 			pkgs.UpdateExistingPackages {
-			return fmt.Errorf("package operations are not yet supported for ACL")
+			if rc.Options.ToolsDir == "" {
```
vinceaperri, line 80: --tools-dir is repeated: "ACL package operations require --tools-dir: ACL images do not include a package manager" conveys the same thing.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
vinceaperri: This tools-chroot setup is duplicated in `CustomizeImageHelperCreate`. A shared helper would be nice here.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -673,8 +674,31 @@ func customizeImageHelper(ctx context.Context, rc *ResolvedConfig, partitionsCus
...
+	var toolsChroot *safechroot.Chroot
+	mountBaseDir := rc.BuildDirAbs
+	mountPointName := "imageroot"
+	if rc.Options.ToolsDir != "" {
+		toolsChrootDir := filepath.Join(rc.BuildDirAbs, toolsRoot)
+		if err := os.MkdirAll(toolsChrootDir, os.ModePerm); err != nil {
```
vinceaperri: Unlike create's validateToolsDir, customize never validates --tools-dir exists / is a directory here.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizeroptions.go
```diff
@@ -28,6 +28,7 @@ type ImageCustomizerOptions struct {
 	PackageSnapshotTime     imagecustomizerapi.PackageSnapshotTime
 	ImageCacheDir           string
 	CosiCompressionLevel    *int
+	ToolsDir                string
```
vinceaperri, line 31: I don't see a test exercising the new `customize --tools-dir` ACL package-ops path, only the create-side rename is covered. Can we add one?

docs/imagecustomizer/how-to/create-tools-dir.md
liulanze: Can you explain more where make less sense? Do you want me to restructure it or remove it? I've done a restructure this time.

toolkit/tools/internal/testutils/testutils.go
vinceaperri: +1 It is now unused.

toolkit/tools/pkg/imagecustomizerlib/createimage_test.go
```diff
@@ -46,7 +46,7 @@ func testCreateImageRaw(t *testing.T, name string, version string, configFile st
 	// get RPM sources
 	downloadedRpmsRepoFile := testutils.GetDownloadedRpmsRepoFile(t, testutilsDir, "azurelinux", version, false, true)
 	rpmSources := []string{downloadedRpmsRepoFile}
-	toolsFile := testutils.GetDownloadedToolsFile(t, testutilsDir, "azurelinux", version, true)
+	toolsFile := testutils.GetDownloadedToolsDir(t, testutilsDir, "azurelinux", version, true)
```
vinceaperri: These `toolsFile` vars now hold the result of `GetDownloadedToolsDir` (a directory). Rename to `toolsDir` here and at lines 115 and 235.

vinceaperri (changes requested):

toolkit/tools/pkg/imagecustomizerlib/createimage_test.go
liulanze: The test was turned into a validation of the error path (the empty tools-dir --> “tools directory is required for image creation” error). It no longer mounts a chroot or runs qemu, so none of CheckSkipForCustomizeImageRequirements’s preconditions apply.

toolkit/tools/pkg/imagecustomizerlib/customizeos.go
liulanze: Fixed, please review again.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
liulanze: Try fixing it, please have a review.

docs/imagecustomizer/how-to/create-tools-dir.md
liulanze: Thanks, agreed, will track as follow-up

docs/imagecustomizer/api/cli/create.md
liulanze: Fixed.

toolkit/tools/pkg/imagecustomizerlib/createimage_test.go
liulanze: Fixed, thanks.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizeroptions.go
liulanze: Added `TestAclValidateConfigPackageOpsRequireToolsDir()`, please review.

docs/imagecustomizer/api/cli/create.md
```diff
@@ -119,15 +119,17 @@ RPMs from an existing RPM repo (such as packages.microsoft.com). Using a cloned
...
-Specifies the path to a tools file in `.tar.gz` format.
+Specifies the path to a directory that provides the tools (tdnf/dnf and their
+dependencies) needed to install packages during image creation.
 
-This file should contain the TDNF tar package (or an equivalent), which is used to manage package dependencies and facilitate installation workflows.
+For instructions on how to create this directory, see:
+[How to create the tools directory](../../how-to/create-tools-dir.md)
 
-Added in v1.2.
+Added in v1.2. Renamed from `--tools-file` in v1.5.
```
cwize1: This property behaves differently. So, I would just say `Added in v1.5`.

toolkit/tools/internal/testutils/testutils.go
```diff
@@ -102,16 +103,34 @@ func GetDownloadedRpmsDir(t *testing.T, testutilsDir string, distro string, dist
...
 			"please run toolkit/tools/internal/testutils/testrpms/download-test-utils.sh -d %s -t %s -s %t",
 			toolsFilePath, distro, distroVersion, createImage)
 	}
-	return toolsFilePath
+
+	if err := os.MkdirAll(toolsDirPath, os.ModePerm); err != nil {
+		t.Fatalf("failed to create tools directory (%s): %v", toolsDirPath, err)
+	}
+
+	cmd := exec.Command("tar", "-x", "-z", "-f", toolsFilePath, "-C", toolsDirPath)
```
cwize1: Why is untarring needed here? Doesn't `download-test-utils.sh` do this already?

toolkit/tools/pkg/imagecustomizerlib/createimage_test.go
cwize1: I would probably leave it in, since the test is still calling `CreateImageWithConfigFile` even if we expect the test to error out before using chroot, etc.

Also, `CheckSkipForCustomizeImageRequirements` includes a check for `testing.Short()` which is how we separate the UTs from the functional tests.

toolkit/tools/pkg/imagecustomizerlib/createimagehelper.go
```diff
@@ -56,12 +53,6 @@ func CustomizeImageHelperCreate(ctx context.Context, rc *ResolvedConfig, tarFile
 		return nil, "", err
 	}
 
-	// Close the tools chroot and image connection.
-	err = toolsChroot.Close(false)
```
cwize1, line 60: We still need to check for the error code in the success path.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -673,8 +673,31 @@ func customizeImageHelper(ctx context.Context, rc *ResolvedConfig, partitionsCus
...
+		}
+		var cleanup func()
+		var err error
+		toolsChroot, cleanup, err = initToolsChroot(ctx, rc.Options.ToolsDir)
+		if err != nil {
+			return nil, nil, nil, "", err
+		}
+		// cleanup must run AFTER imageConnection.Close so the image unmounts
+		// before the tools chroot directory is torn down.
+		defer cleanup()
```
cwize1: Same comment here. We need to check the error codes during the success path.

cwize1 (changes requested):

toolkit/tools/internal/testutils/testutils.go
liulanze: Fixed, thanks.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
liulanze: Done, reverted the createimagehelper.go change.

toolkit/tools/pkg/imagecustomizerlib/createimagehelper.go
```diff
@@ -76,7 +72,7 @@ func doOsCustomizationsCreate(
 	imageChroot := imageConnection.Chroot()
 	buildTime := time.Now().Format(buildTimeFormat)
 
-	resolvConf, err := overrideResolvConf(toolsChroot)
+	resolvConf, err := overrideResolvConf(imageChroot)
```
cwize1, line 78: Remove `overrideResolvConf` and `restoreResolvConf` since these are now handled by `initToolsChroot`?

toolkit/tools/pkg/imagecustomizerlib/toolschroot.go
```diff
@@ -0,0 +1,40 @@
...
+	if err != nil {
+		if closeErr := toolsChroot.Close(false); closeErr != nil {
+			logger.Log.Warnf("Failed to close tools chroot (%s) after resolv.conf override failure: %v",
+				toolsDir, closeErr)
+		}
+		return nil, nil, fmt.Errorf("failed to override resolv.conf in tools chroot:\n%w", err)
+	}
+
+	cleanup := func() {
+		if restoreErr := restoreResolvConf(ctx, toolsResolvConf, toolsChroot); restoreErr != nil {
```
cwize1: Ideally, the error code for `restoreResolvConf` should also be checked during the success case.

It might be worth having `initToolsChroot` return a struct and adding a `Close` and `CleanClose` functions to that struct.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/createimagehelper.go
liulanze: These handle different chroots I think. `initToolsChroot` overrides `resolv.conf` in the tools chroot so that `tdnf` can resolve repo URLs. This call overrides `resolv.conf` in the image chroot, the `toolsDir/_imageroot/etc/resolv.conf`, so user `postCustomization/finalizeCustomization` scripts and RPM `%post` scriptlets running inside `--installroot` have DNS. But I will add comment to clarify.

toolkit/tools/pkg/imagecustomizerlib/toolschroot.go
liulanze: Refactored initToolsChroot.

toolkit/tools/pkg/imagecustomizerlib/createimagehelper.go
cwize1: Ah, sorry. I missed that you changed this to `imageChroot`.

And I guess the `restoreResolvConf` below was wrong since it is already pointing to `imageChroot`.

toolkit/tools/pkg/imagecustomizerlib/toolschroot.go
```diff
@@ -0,0 +1,40 @@
...
+			logger.Log.Warnf("Failed to restore resolv.conf in tools chroot (%s): %v",
+				toolsDir, restoreErr)
+		}
+		if closeErr := toolsChroot.Close(false); closeErr != nil {
```
vinceaperri: `Close(false)` makes safechroot `os.RemoveAll` the chroot root, which is now the user's real `--tools-dir`, so it gets deleted after every run and breaks the reuse promised in create-tools-dir.md. Use `Close(true)` here and in the error path above so only the mounts are torn down.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -673,8 +672,30 @@ func customizeImageHelper(ctx context.Context, rc *ResolvedConfig, partitionsCus
...
+	if rc.Options.ToolsDir != "" {
+		if err := validateToolsDir(rc.Options.ToolsDir); err != nil {
+			return nil, nil, nil, "", err
+		}
+		var err error
+		toolsChroot, err = initToolsChroot(rc.Options.ToolsDir)
+		if err != nil {
+			return nil, nil, nil, "", err
+		}
+		// Close must run AFTER imageConnection.Close so the image unmounts
```
cwize1: It is typical to destroy resources in the reverse order that they were created. So, I don't think this comment is needed.

toolkit/tools/pkg/imagecustomizerlib/toolschroot.go
```diff
@@ -0,0 +1,86 @@
+// Copyright (c) Microsoft Corporation.
```
cwize1, line 1: I would write it like this:

```golang
// Copyright (c) Microsoft Corporation.
// Licensed under the MIT License.

package imagecustomizerlib

import (
	"context"
	"fmt"

	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/logger"
	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/safechroot"
)

// ToolsChroot wraps a safechroot used to run tdnf/dnf against a target image
// via --installroot. It owns the resolv.conf override on the tools chroot.
type ToolsChroot struct {
	ctx        context.Context
	chroot     *safechroot.Chroot
	resolvConf *resolvConfInfo
	toolsDir   string
}

func initToolsChroot(ctx context.Context, toolsDir string) (*ToolsChroot, error) {
	t := &ToolsChroot{
		ctx:      ctx,
		toolsDir: toolsDir,
	}

	err := t.initToolsChroot()
	if err != nil {
		t.Close()
		return nil, err
	}

	return t, nil
}

func (t *ToolsChroot) initToolsChroot() error {
	t.chroot = safechroot.NewChroot(t.toolsDir, true)
	if err := t.chroot.Initialize("", nil, nil, true); err != nil {
		return fmt.Errorf("failed to initialize tools chroot from %s:\n%w", t.toolsDir, err)
	}

	resolvConf, err := overrideResolvConf(t.chroot)
	if err != nil {
		return fmt.Errorf("failed to override resolv.conf in tools chroot:\n%w", err)
	}
	t.resolvConf = &resolvConf

	return nil
}

// Chroot returns the underlying safechroot. Nil-safe.
func (t *ToolsChroot) Chroot() *safechroot.Chroot {
	return t.chroot
}

// CleanClose restores resolv.conf and closes the chroot, returning the first
// error. Use this in the success path. Safe to call multiple times.
func (t *ToolsChroot) CleanClose() error {
	if err := restoreResolvConf(t.ctx, *t.resolvConf, t.chroot); err != nil {
		return fmt.Errorf("failed to restore resolv.conf in tools chroot:\n%w", err)
	}
	t.resolvConf = nil

	if err := t.chroot.Close(true); err != nil {
		return fmt.Errorf("failed to initialize tools chroot from %s:\n%w", t.toolsDir, err)
	}

	return nil
}

// Close is a best-effort cleanup intended for `defer`. Errors are logged. Safe
// to call multiple times.
func (t *ToolsChroot) Close() {
	if t.resolvConf != nil {
		if err := restoreResolvConf(t.ctx, *t.resolvConf, t.chroot); err != nil {
			logger.Log.Warnf("Failed to restore resolv.conf in tools chroot (%s): %v", t.toolsDir, err)
		}
		t.resolvConf = nil
	}

	if t.chroot != nil {
		if err := t.chroot.Close(true); err != nil {
			logger.Log.Warnf("Failed to close tools chroot (%s): %v", t.toolsDir, err)
		}
	}
}
```

If you track the state of each resource separately, then it means the `Close` function can run against a partially constructed or partially deconstructed `ToolsChroot` instance. This allows you to use the `Close` function to clean up the `ToolsChroot` when something goes wrong.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/toolschroot.go
```diff
@@ -0,0 +1,86 @@
...
+// via --installroot. It owns the resolv.conf override on the tools chroot.
+type ToolsChroot struct {
+	chroot     *safechroot.Chroot
+	resolvConf resolvConfInfo
+	toolsDir   string
+	closed     bool
+}
+
+func initToolsChroot(toolsDir string) (*ToolsChroot, error) {
+	chroot := safechroot.NewChroot(toolsDir, true)
```
vinceaperri: This wraps the user's real `--tools-dir`, so every `Close(false)` below (lines 31, 69, 83) makes safechroot `os.RemoveAll` it, deleting the dir (and the shared `tools-*-dir` test fixture) after each create and customize run despite the reuse promised in create-tools-dir.md. Use `Close(true)` so only the mounts are torn down.

vinceaperri (dismissed):

toolkit/tools/pkg/imagecustomizerlib/toolschroot.go
liulanze: Good catch, fixed in the latest revision. Both Close() and CleanClose() now use chroot.Close(true).

cwize1 (approved):

PR #779, opened 2026-06-16
Add `--setfiles-context` API.

cwize1:
Customizing an OS with SELinux on a Fedora dev box currently hits errors because the kernel will block you from applying labels to files that don't exist in the kernel's active policy (among other issues). To fix this problem, the `setfiles` call needs to be run under the `setfiles_mac_t` label, which has special permissions to ignore the active policy. (This label is used by Fedora's Anaconda installer.)

This change exposes a new CLI API that allows the user to specify the label to run the `setfiles` call under. This will allow Fedora users to specify the required `setfiles_mac_t` label.

Exposing this as an API helps keep the design simple since it avoids logic for stuff like detecting if SELinux is enabled, what the correct label to use for the current build host, etc.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/docs/imagecustomizer/api/cli/customize.md
+++ b/docs/imagecustomizer/api/cli/customize.md
@@ -298,13 +298,16 @@ This option is useful when running on a build system with SELinux enabled.
 
 Added in v1.5.
 
-### Fedora
+### selinux-policy (e.g. Fedora)
 
-When running on a Fedora system, use the value: `system_u:system_r:setfiles_mac_t:s0`
+When running on a build system with SELinux enabled and that uses the
+[selinux-policy](https://github.com/fedora-selinux/selinux-policy) policy (e.g. Fedora),
+use the value: `system_u:system_r:setfiles_mac_t:s0`
 
-The `setfiles_mac_t` label allows SELinux labels to be applied that aren't present in
-the build host's SELinux rules, which is super useful when customizing images that have
-a different distro / distro version than the build host.
+The `setfiles_mac_t` label is allowed to use the `CAP_MAC_ADMIN` capability, which
+permits SELinux labels to be applied that aren't present in the build host's SELinux
+rules. This is super useful when customizing images that have a different distro /
+distro version than the build host.
 
 Also, you need to run the commands:
 
@@ -315,3 +318,12 @@ Also, you need to run the commands:
 2. `sudo semanage permissive -a systemd_hwdb_t`
 
    `setfiles_mac_t` doesn't have the correct permissions to label `/etc/udev/hwdb.bin`.
+
+### refpolicy
+
+When running on a build system that uses the
+[refpolicy](https://github.com/SELinuxProject/refpolicy) policy (e.g. Azure Linux 3.0),
+the only thing you can do is disable SELinux.
+
+This SELinux policy does not have a label with permissions for the `CAP_MAC_ADMIN`
+capability.
--- a/docs/imagecustomizer/api/cli/customize.md
+++ b/docs/imagecustomizer/api/cli/customize.md
@@ -308,3 +308,5 @@ Also, you need to run the commands:
    This allows transitions from `unconfined_t` to `setfiles_mac_t`.
 
 2. `sudo semanage permissive -a systemd_hwdb_t`
+
+   `setfiles_mac_t` doesn't have the correct permissions to label `/etc/udev/hwdb.bin`.
--- a/docs/imagecustomizer/api/cli/customize.md
+++ b/docs/imagecustomizer/api/cli/customize.md
@@ -287,7 +287,12 @@ Added in v1.5.
 
 ## --setfiles-context
 
-The SELinux context to use when running the `setfiles` command.
+The SELinux context to use when running the `setfiles` command, which is used to apply
+the security context labels on all the files in the OS when SELinux is enabled in the
+target OS.
+
+If omitted, the `setfiles` command is run in the default security context chosen by the
+kernel.
 
 This option is useful when running on a build system with SELinux enabled.
 
--- a/docs/imagecustomizer/api/cli/customize.md
+++ b/docs/imagecustomizer/api/cli/customize.md
@@ -284,3 +284,27 @@ For instructions on how to create this directory, see:
 [How to create the tools directory](../../how-to/create-tools-dir.md)
 
 Added in v1.5.
+
+## --setfiles-context
+
+The SELinux context to use when running the `setfiles` command.
+
+This option is useful when running on a build system with SELinux enabled.
+
+Added in v1.5.
+
+### Fedora
+
+When running on a Fedora system, use the value: `system_u:system_r:setfiles_mac_t:s0`
+
+The `setfiles_mac_t` label allows SELinux labels to be applied that aren't present in
+the build host's SELinux rules, which is super useful when customizing images that have
+a different distro / distro version than the build host.
+
+Also, you need to run the commands:
+
+1. `sudo semanage permissive -a setfiles_mac_t`
+
+   This allows transitions from `unconfined_t` to `setfiles_mac_t`.
+
+2. `sudo semanage permissive -a systemd_hwdb_t`
--- a/toolkit/tools/imagegen/installutils/installutils.go
+++ b/toolkit/tools/imagegen/installutils/installutils.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"runtime"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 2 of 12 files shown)


toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -911,6 +898,42 @@ func SELinuxRelabelFiles(installChroot safechroot.ChrootInterface, mountPointToF
...
+
+	// We only want to print basic info, filter out the real output unless at trace level (Execute call handles that)
+	files := 0
+	onStdout := func(line string) {
+		files++
+		if (files % 1000) == 0 {
+			logger.Log.Debugf("SELinux: labelled %d files", files)
+		}
+	}
+	err := shell.NewExecBuilder("setfiles", "-m", "-v", "-r", targetRootPath, fileContextPath, targetPath).
```
vinceaperri, line 911: Should add a comment here saying `.Capabilities()` should never be added to this call, since it would fork setfiles onto a separate thread (see trackAndStartProcess), which would silently drop the label since the exec context only applies to children forked on the locked thread.

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -911,6 +898,42 @@ func SELinuxRelabelFiles(installChroot safechroot.ChrootInterface, mountPointToF
...
 
+func callSetFiles(targetRootPath string, fileContextPath string, targetPath string, chrootDir string,
+	setFilesContext string,
+) error {
+	if setFilesContext != "" {
+		runtime.LockOSThread()
+		defer runtime.UnlockOSThread()
+		err := setSELinuxExecContext(setFilesContext)
+		if err != nil {
+			return fmt.Errorf("failed to set setfiles exec context (%s)", setFilesContext)
```
vinceaperri: The underlying error is dropped. Wrap it with `:\n%w`?

toolkit/tools/imagegen/installutils/installutils.go
```diff
@@ -911,6 +898,42 @@ func SELinuxRelabelFiles(installChroot safechroot.ChrootInterface, mountPointToF
...
+		}
+
+		defer resetSELinuxExecContext()
```
vinceaperri: resetSELinuxExecContext's error is dropped, then UnlockOSThread returns this dirty thread to the pool. On reset failure, maybe skip the unlock so the runtime retires the thread.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizeroptions.go
```diff
@@ -28,6 +28,7 @@ type ImageCustomizerOptions struct {
 	PackageSnapshotTime     imagecustomizerapi.PackageSnapshotTime
 	ImageCacheDir           string
 	CosiCompressionLevel    *int
+	SetFilesContext         string
```
vinceaperri, line 32: I don't see a test exercising a non-empty setfiles-context. Can the new exec-context path be covered?

docs/imagecustomizer/api/cli/customize.md
```diff
@@ -266,3 +266,27 @@ You may enable this feature by adding `input-image-oci` to the
...
+the build host's SELinux rules, which is super useful when customizing images that have
+a different distro / distro version than the build host.
+
+Also, you need to run the commands:
+
+1. `sudo semanage permissive -a setfiles_mac_t`
+
+   This allows transitions from `unconfined_t` to `setfiles_mac_t`.
+
+2. `sudo semanage permissive -a systemd_hwdb_t`
```
vinceaperri, line 318: Add an explanation for the second command?

vinceaperri (changes requested):

toolkit/tools/pkg/imagecustomizerlib/imagecustomizeroptions.go
cwize1: Not easily. It requires the host to have SELinux enabled. And the label to use depends on the active SELinux policy.

toolkit/tools/imagegen/installutils/installutils.go
cwize1: I moved setting the SELinux context into the `shell` package.

toolkit/tools/imagegen/installutils/installutils.go
cwize1: Fixed

toolkit/tools/imagegen/installutils/installutils.go
cwize1: The code now uses a goroutine and always throws away the OS thread.

docs/imagecustomizer/api/cli/customize.md
cwize1: Fixed

toolkit/tools/internal/shell/shell.go
```diff
@@ -178,40 +178,51 @@ func trackAndStartProcess(cmd *exec.Cmd, capabilities []uintptr) (err error) {
...
+	// thread.
 	runtime.LockOSThread()
 
-	err := setOSThreadCapabilities(capabilities)
-	if err != nil {
-		return fmt.Errorf("failed to set process capabilities:\n%w", err)
+	if selinuxContext != "" {
+		err := setSELinuxExecContext(selinuxContext)
+		if err != nil {
+			return fmt.Errorf("failed to set process SELinux exec context (%s)\n%w", selinuxContext, err)
```
vinceaperri: Missing colon before `\n%w` like in the capabilities error below

toolkit/tools/internal/shell/execbuilder.go
```diff
@@ -74,18 +75,24 @@ func (b ExecBuilder) WorkingDirectory(path string) ExecBuilder {
...
 
-// Chroot sets
+// Chroot sets the root directory to run the command under.
 func (b ExecBuilder) Chroot(chrootDir string) ExecBuilder {
 	b.chrootDir = chrootDir
 	return b
 }
 
-// Chroot sets
+// Capabilities sets capabilities the command will be be restricted to.
```
vinceaperri: will be restricted to

toolkit/tools/internal/shell/selinux.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	return nil
+}
+
+// Clears the explicit override of the SELinux context to use for child processes started by the current OS thread.
+//
+// Note: This function should only be run under the runtime.LockOSThread() lock and on the same thread that called
+// setSELinuxExecContext().
+func resetSELinuxExecContext() error {
```
vinceaperri, line 49: `resetSELinuxExecContext` is now unused since the OS thread is always thrown away. Remove it?

toolkit/tools/internal/shell/shell.go
```diff
@@ -178,40 +178,51 @@ func trackAndStartProcess(cmd *exec.Cmd, capabilities []uintptr) (err error) {
...
 	close(result)
 }
 
-func setCapabilitiesAndStartCmdHelper(cmd *exec.Cmd, capabilities []uintptr) error {
-	// Lock the OS thread so that capabilities can be dropped without affecting other threads.
+func goStartCmdHelper(cmd *exec.Cmd, capabilities []uintptr, selinuxContext string) error {
+	// Lock the OS thread to avoid affecting other goroutines.
 	// Note: Dropping capabilities cannot be undone. Hence, the OS thread needs to be thrown away.
 	// Hence, UnlockOSThread is not called and a goroutine is used.
+	// Also, setting the SELinux context can be technically be undone. But its less code to just throw away the OS
```
vinceaperri: "can technically be undone"

"But its" -> "But it's"

vinceaperri (changes requested):

toolkit/tools/internal/shell/shell.go
cwize1: Fixed

toolkit/tools/internal/shell/execbuilder.go
cwize1: Fixed

toolkit/tools/internal/shell/shell.go
cwize1: Fixed

toolkit/tools/internal/shell/selinux.go
cwize1: I am keeping the function around in case anyone wants to optimize the code to not throw away the OS thread.

vinceaperri: Fair. A one-line comment on the function noting it is intentionally retained for a future non-throwaway optimization would stop the next reader from deleting it as dead code.

cwize1: Fixed

vinceaperri (approved):

PR #780, opened 2026-06-17
Auto-provision tools chroot for ACL package ops

liulanze:
Adds automatic provisioning of a tools chroot for distros without an in-image package manager, today: ACL. 

IC resolves a distro to OCI reference (e.g. mcr.microsoft.com/azurelinux/base/core:3.0), pulls and extracts it into a digest-keyed cache (--image-cache-dir → $XDG_CACHE_HOME → $HOME/.cache), and stages a working copy under <buildDir>/toolsroot/ for safechroot to drive tdnf. 

No new CLI flags. All wiring is gated on DistroHandler.NeedsToolsChroot(), which returns false for AZL3/AZL4/Ubuntu/Fedora, so existing distros are unchanged. 

New code lives in pkg/imagecustomizerlib/toolschroot/: resolver.go (distro to ref map), pull.go (oras-go fetch + multi-arch resolution + atomic cache promotion), extract.go (streaming OCI tar apply with whiteout handling).


toolkit/tools/pkg/imagecustomizerlib/toolschroot_setup.go
```diff
@@ -0,0 +1,86 @@
+// Copyright (c) Microsoft Corporation.
```
cwize1, line 1: Filenames should not have underscores in them, since underscore has a special meaning in Go. For example, `_test.go`.

The distrohandler files are an exception and should probably be renamed.

toolkit/tools/pkg/imagecustomizerlib/toolschroot/resolver.go
```diff
@@ -0,0 +1,46 @@
...
+
+var ErrUnsupportedDistro = errors.New("no built-in tools chroot container is registered for this distro/version")
+
+type mapKey struct {
+	distro    targetos.Distro
+	versionId string
+}
+
+// ACL has no package manager in its image; AZL3 base provides the tdnf userspace.
+var builtInContainerRefs = map[mapKey]string{
```
cwize1, line 23: Unfortunately, we need to verify the signature of the container.

For the Azure Linux containers, these should be signed with notary. Notary is also used to sign the VHDX OCIs. So, you should be able to reuse some of the existing IC code.

For Fedora, these are signed with sigstore cosign signatures. So, we'll need new logic for these. (So, it might be worth handling Fedora in a separate change.)

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -673,8 +674,38 @@ func customizeImageHelper(ctx context.Context, rc *ResolvedConfig, partitionsCus
...
+	// so that package operations can run as `tdnf --installroot=/_imageroot`.
+	// For these distros, mount the image inside the tools chroot at
+	// /<toolsRootImageDir>/; for everything else, keep the existing behavior of
+	// mounting the image directly under the build dir.
+	var (
+		toolsChroot      *safechroot.Chroot
+		connectBuildDir  = rc.BuildDirAbs
+		connectChrootDir = "imageroot"
+	)
+	if distroHandler.NeedsToolsChroot() {
```
cwize1, line 687: Customizing an ACL image doesn't need the tools chroot unless the user is trying to install/update packages. So, the tools chroot isn't always needed.

Also, I would like to avoid having such a big difference in behavior between distros. This type of inconsistency can be confusing for users.

Also, we should have an API that allows users to opt-in for the `customize` command.

So, I think we should add a new CLI arg to `create` and `customize` called something like `--tools-dir-auto` that invokes this behavior.

toolkit/tools/pkg/imagecustomizerlib/toolschroot/resolver.go
```diff
@@ -0,0 +1,46 @@
...
+	{targetos.AzureLinux, "2.0"}: "mcr.microsoft.com/cbl-mariner/base/core:2.0",
+	{targetos.AzureLinux, "3.0"}: "mcr.microsoft.com/azurelinux/base/core:3.0",
+	// AZL4 is only published under `-beta` until GA.
+	{targetos.AzureLinux, "4.0"}:          "mcr.microsoft.com/azurelinux-beta/base/core:4.0",
+	{targetos.AzureContainerLinux, "3.0"}: "mcr.microsoft.com/azurelinux/base/core:3.0",
+	{targetos.Fedora, "42"}:               "quay.io/fedora/fedora:42",
+}
+
+func Resolve(target targetos.TargetOs) (string, error) {
```
cwize1, line 32: Any reason `Resolve` is public? (Capital first letter.)

toolkit/tools/pkg/imagecustomizerlib/toolschroot_setup.go
```diff
@@ -0,0 +1,86 @@
...
+	"ToolsChroot:CacheDirUnresolved",
+	"could not determine a cache directory for the auto-provisioned tools chroot")
+
+const (
+	toolsChrootCacheSubdir           = "toolschroot"
+	defaultToolsChrootCacheNamespace = "imagecustomizer"
+)
+
+// resolveToolsChrootCacheDir picks the cache root for auto-provisioned tools chroots, preferring
+// --image-cache-dir, then $XDG_CACHE_HOME, then $HOME/.cache.
```
cwize1, line 29: Drop support for `XDG_CACHE_HOME` and `.cache`. 

These won't really work correctly in the container.

Also, I am not really a fan of tools that automatically place files in hidden places.

toolkit/tools/pkg/imagecustomizerlib/toolschroot/resolver_test.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestResolveSupportedDistros(t *testing.T) {
+	cases := []struct {
+		name        string
+		target      targetos.TargetOs
+		expectedRef string
+	}{
+		{"AzureLinux3", targetos.TargetOsAzureLinux3, "mcr.microsoft.com/azurelinux/base/core:3.0"},
```
cwize1, line 20: In general, you shouldn't duplicate product code in test code. And in particular, it isn't worth duplicating information stored in a declarative structure. So, I would just remove the `TestResolveSupportedDistros` test.

toolkit/tools/pkg/imagecustomizerlib/toolschroot/pull.go
```diff
@@ -0,0 +1,263 @@
...
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/logger"
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/targetos"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"oras.land/oras-go/v2"
+	"oras.land/oras-go/v2/content"
+	"oras.land/oras-go/v2/registry/remote"
+)
+
+var (
+	ErrCacheDirRequired = errors.New("tools chroot cache directory is required")
```
cwize1, line 27: We should probably move `ImageCustomizerError` into its own package so that other packages can reference it. Though, this will be a large change and so should be done in its own PR.

toolkit/tools/pkg/imagecustomizerlib/toolschroot/pull.go
```diff
@@ -0,0 +1,263 @@
...
+	return destDir, nil
+}
+
+func cachePathForDigest(cacheDir string, desc ocispec.Descriptor) string {
+	return filepath.Join(cacheDir, string(desc.Digest.Algorithm()), desc.Digest.Encoded())
+}
+
+// resolveImageManifest descends into a multi-arch index if needed, returning a
+// single-platform image manifest descriptor.
+func resolveImageManifest(ctx context.Context, source oras.ReadOnlyTarget, ref string) (ocispec.Descriptor, error) {
```
cwize1, line 86: Is it possible to deduplicate some of this logic with what's in `oras.go`?

toolkit/tools/pkg/imagecustomizerlib/toolschroot/pull.go
```diff
@@ -0,0 +1,263 @@
...
+		return true
+	default:
+		return false
+	}
+}
+
+// pullAndExtractToCache extracts layers into a staging dir and atomically
+// renames it onto destDir. If another process wins the race, the staging dir
+// is discarded.
+func pullAndExtractToCache(ctx context.Context, source content.ReadOnlyStorage,
```
cwize1, line 123: nit:

```golang
func pullAndExtractToCache(ctx context.Context, source content.ReadOnlyStorage,
	manifestDesc ocispec.Descriptor, destDir string,
) error {
    err := pullAndExtractToCacheHelper(ctx, source, manifestDesc , destDir)
    if err != nil {
        return fmt.Errorf("%w:\n%w", ErrPullFailed, err)
    }

    return nil
}

func pullAndExtractToCacheHelper(ctx context.Context, source content.ReadOnlyStorage,
	manifestDesc ocispec.Descriptor, destDir string,
) error {
    ...
{
```

That way, you don't have to duplicate `ErrPullFailed` on each error case.

toolkit/tools/pkg/imagecustomizerlib/toolschroot/pull.go
```diff
@@ -0,0 +1,263 @@
...
+	manifestDesc ocispec.Descriptor, destDir string,
+) error {
+	parent := filepath.Dir(destDir)
+	if err := os.MkdirAll(parent, 0o755); err != nil {
+		return fmt.Errorf("%w: creating cache parent (%s):\n%w", ErrPullFailed, parent, err)
```
cwize1, line 128: nit: `failed to create cache parent (%s)`

The error message should describe what failed, not what was being done.

Same comment elsewhere.

toolkit/tools/pkg/imagecustomizerlib/toolschroot/pull.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	stagingDir, err := os.MkdirTemp(parent, filepath.Base(destDir)+".tmp-*")
+	if err != nil {
+		return fmt.Errorf("%w: creating staging dir:\n%w", ErrPullFailed, err)
+	}
+	keepStaging := false
```
cwize1, line 135: RE `keepStaging`: In other places, I name this variable `ok`.

toolkit/tools/pkg/imagecustomizerlib/toolschroot/extract.go
```diff
@@ -0,0 +1,292 @@
...
+		}
+
+		if err := applyTarEntry(tr, hdr, root); err != nil {
+			return err
+		}
+	}
+}
+
+func applyTarEntry(tr *tar.Reader, hdr *tar.Header, root *os.Root) error {
+	cleanRel, err := safeJoinRel(hdr.Name)
```
cwize1, line 52: If `os.Root` is being used, then is `safeJoinRel` needed? Why not just let `os.Root` report the error?

toolkit/tools/pkg/imagecustomizerlib/toolschroot/extract.go
```diff
@@ -0,0 +1,292 @@
...
+)
+
+var ErrExtractFailed = errors.New("failed to extract OCI image layer")
+
+const (
+	whiteoutPrefix     = ".wh."
+	opaqueWhiteoutName = ".wh..wh..opq"
+)
+
+// extractTarLayer applies a tar stream onto destDir with OCI/Docker whiteout
```
cwize1, line 24: I think we should use [go-extract](https://github.com/hashicorp/go-extract) here since it is designed to handle a lot of the security problems. Also, it will likely simplify the code a lot.

cwize1 (changes requested):

liulanze:
Closed, in the future this PR can be a refer to more generic new API like `--tool-dirs-auto`

PR #781, opened 2026-06-17
Make package-information retrieval distro-aware

vinceaperri:
Package version/release information was retrieved with a single hard-coded
`tdnf info` call, which only works on Azure Linux 2.0/3.0. This routes it through
the `DistroHandler` so each distro queries its native package database. It
unblocks package-version lookups (e.g. `dracut`, `selinux-policy`) needed by the
LiveOS/ISO path on Azure Linux 4.0 and Fedora.

### Changes

- **`DistroHandler` interface:** add
  `GetPackageInformation(imageChroot, packageName) (*PackageVersionInformation, error)`.
  - Azure Linux 2.0/3.0, Azure Linux 4.0, Fedora, and Azure Container Linux
    delegate to their package manager.
  - Ubuntu returns an "unsupported" error. The RPM-style
    `PackageVersionInformation` has no Debian equivalent.
- **`rpmPackageManagerHandler` interface:** add `getPackageInformation(...)`.
  - **tdnf** — `tdnf info <pkg> --repo @system` (the previous behavior).
  - **dnf** — `rpm -q --queryformat 'Name : … Version : … Release : …'`, which
    reads the local rpm DB directly (works on read-only chroots, no dnf logdir
    writes) and reuses the same parser.
- **`packageinformation.go`:** extract the pure `parsePackageInfoOutput` parser
  from the old shell-invoking `getPackageInformation` so both package managers
  share it. Also fixes error messages that formatted a `nil` error with `%w`.
- **Caller (`liveosisoartifactstore.go`):** use
  `distroHandler.GetPackageInformation(...)` instead of the package-global helper.
- **Azure Linux 4.0:** remove the `ValidateConfig` guard that hard-rejected ISO
  and PXE output formats, as part of enabling those formats for AZL 4.0.

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/distrohandler_ubuntu.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/distrohandler_ubuntu.go
@@ -100,7 +100,7 @@ func (d *ubuntuDistroHandler) IsPackageInstalled(imageChroot safechroot.ChrootIn
 
 func (d *ubuntuDistroHandler) GetPackageInformation(imageChroot *safechroot.Chroot, packageName string,
 ) (*PackageVersionInformation, error) {
-	return nil, fmt.Errorf("Getting package information is not supported for Ubuntu images:\n%w",
+	return nil, fmt.Errorf("Getting package information is not supported yet for Ubuntu images:\n%w",
 		ErrUnsupportedUbuntuFeature)
 }
 
--- a/toolkit/tools/pkg/imagecustomizerlib/distrohandler_ubuntu.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/distrohandler_ubuntu.go
@@ -98,6 +98,12 @@ func (d *ubuntuDistroHandler) IsPackageInstalled(imageChroot safechroot.ChrootIn
 	return isPackageInstalledDeb(imageChroot, packageName)
 }
 
+func (d *ubuntuDistroHandler) GetPackageInformation(imageChroot *safechroot.Chroot, packageName string,
+) (*PackageVersionInformation, error) {
+	return nil, fmt.Errorf("Getting package information is not supported for Ubuntu images:\n%w",
+		ErrUnsupportedUbuntuFeature)
+}
+
 func (d *ubuntuDistroHandler) GetAllPackagesFromChroot(imageChroot safechroot.ChrootInterface) ([]OsPackage, error) {
 	return getAllPackagesFromChrootDeb(imageChroot)
 }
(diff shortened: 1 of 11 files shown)


toolkit/tools/pkg/imagecustomizerlib/distrohandler_ubuntu.go
```diff
@@ -98,6 +98,12 @@ func (d *ubuntuDistroHandler) IsPackageInstalled(imageChroot safechroot.ChrootIn
 	return isPackageInstalledDeb(imageChroot, packageName)
 }
 
+func (d *ubuntuDistroHandler) GetPackageInformation(imageChroot *safechroot.Chroot, packageName string,
+) (*PackageVersionInformation, error) {
+	return nil, fmt.Errorf("Getting package information is not supported for Ubuntu images:\n%w",
```
cwize1: Say `is not supported yet` if this is just missing functionality in IC, as opposed to missing functionality in the target OS.

toolkit/tools/pkg/imagecustomizerlib/packagemanager_dnf.go
```diff
@@ -195,6 +196,23 @@ func (pm *dnfPackageManager) isPackageInstalled(imageChroot safechroot.ChrootInt
...
+	// already expects from `tdnf info` (Name/Version/Release labels), so we share the parser.
+	packageInfo, _, err := shell.NewExecBuilder("rpm", "-q", "--queryformat",
+		"Name : %{NAME}\nVersion : %{VERSION}\nRelease : %{RELEASE}\n", "--", packageName).
+		LogLevel(logrus.TraceLevel, logrus.DebugLevel).
+		Chroot(imageChroot.ChrootDir()).
+		ExecuteCaptureOutput()
+	if err != nil {
+		return nil, fmt.Errorf("failed to query (%s) package information via rpm:\n%w", packageName, err)
+	}
+	return parsePackageInfoOutput(packageName, packageInfo)
```
cwize1: Add `if err != nil {` check for `parsePackageInfoOutput`.

While this makes the code longer, it also makes it easier to add append to the function or move stuff around in the future.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/distrohandler_ubuntu.go
vinceaperri: Done.

toolkit/tools/pkg/imagecustomizerlib/packagemanager_dnf.go
vinceaperri: Done.

cwize1 (approved):

PR #786, opened 2026-06-18
Consolidate per-distro preview flags into a single preview-distro-version flag

vinceaperri:
Image Customizer gated each preview distribution behind its own preview feature, so users had to know a different flag for every distro. This change replaces those per-distro flags with a single `preview-distro-version` flag that covers every preview distro and preview distro version.

Azure Linux 4.0 is still in preview, so it now also requires `preview-distro-version` to customize or create. Previously AZL4 was ungated.

`preview-distro-version` is distinct from `unsupported-distro-version`. The former opts into a known-but-preview distro or version. The latter applies default settings to a distro version Image Customizer does not yet recognize.

**Flag changes**

| Before | After |
|--------|-------|
| `fedora` | `preview-distro-version` |
| `ubuntu` | `preview-distro-version` |
| `azure-container-linux` | `preview-distro-version` |
| AZL4 (ungated) | `preview-distro-version` required |

Also in this change:
- Collapse the three `Err*PreviewFeatureRequired` errors into one `ErrPreviewDistroVersionFeatureRequired`.
- Update Go unit tests, testdata configs, and the vmtests suite (`test_create.py`, `test_logging.py`).
- Update the configuration and distribution-support docs, including a note on the difference between `preview-distro-version` and `unsupported-distro-version`.

---

### **Checklist**
- [x] Tests added/updated


test/vmtests/vmtests/imagecustomizer/test_create.py
```diff
@@ -162,7 +163,7 @@ def run_create_image_test(
 
     final_config_path = config_path
     if distro.lower() == "fedora":
```
cwize1: Change `distro.lower() == "fedora"` to `is_preview_distro_version`?

cwize1 (changes requested):

test/vmtests/vmtests/imagecustomizer/test_create.py
vinceaperri: Done.

cwize1 (dismissed):

cwize1 (dismissed):

cwize1 (approved):

PR #788, opened 2026-06-18
Make boot-file architecture config distro-aware

vinceaperri:
Boot-file paths and ESP layout are now chosen per distro rather than from a single architecture-only global, and the ISO-to-ISO pipeline detects the distro directly from the initrd.

Previously `getBootArchConfig()` was a global that only varied by runtime architecture, so every distro received the same Azure-Linux-style ESP layout. This moves the lookup behind the `DistroHandler` interface so each distro picks its own boot-file layout, and adds a way to identify the distro from an initramfs when no rootfs is mounted.

**What changed**
- Added `GetBootArchConfig()` to the `DistroHandler` interface. Each distro returns its own per-architecture boot-file map.
- Azure Linux 4 and Fedora now use a Fedora-style ESP layout (`/EFI/BOOT` with `BOOTX64.EFI` / `BOOTAA64.EFI`). Azure Linux 2/3, ACL, and Ubuntu keep the existing Azure Linux layout.
- Added `targetos.GetInitrdTargetOs()` and `NewDistroHandlerFromInitrd()`. These read an `os-release` or dracut `initrd-release` file straight out of the initramfs cpio, auto-detecting gzip and zstd compression from the stream's magic bytes.
- The ISO-to-ISO path now resolves the distro from the ISO's initrd, choosing the bootstrap per-kernel initrd or the full-OS initrd based on the ISO's initramfs type.
- Replaced the global `getBootArchConfig()` with `bootArchConfigFromMap()` and routed every boot-file lookup (UKI creation, PXE artifacts, ISO artifact store, output artifacts) through the distro handler.

**Distro to ESP boot-file layout**

| Distro | Boot-file layout |
|---|---|
| Azure Linux 2 / 3 | Azure Linux |
| Azure Linux 4 | Fedora-style |
| Fedora | Fedora-style |
| ACL | Azure Linux |
| Ubuntu | Azure Linux |

### Checklist
- [ ] Tests added/updated

### DIFF
--- a/toolkit/tools/internal/initrdutils/initrdread.go
+++ b/toolkit/tools/internal/initrdutils/initrdread.go
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-package targetos
+package initrdutils
 
 import (
 	"bufio"
@@ -29,10 +29,10 @@ var (
 	longestMagicSize = len(magicZstd)
 )
 
-// readFirstFileFromInitrd scans an initramfs cpio archive once and returns the content of the first candidate path in
+// ReadFirstFileFromInitrd scans an initramfs cpio archive once and returns the content of the first candidate path in
 // the provided list that exists as a regular file. Returns an error wrapping fs.ErrNotExist if none of the candidates
 // is present.
-func readFirstFileFromInitrd(initrdPath string, candidates []string) (content []byte, foundPath string, err error) {
+func ReadFirstFileFromInitrd(initrdPath string, candidates []string) (content []byte, foundPath string, err error) {
 	f, err := os.Open(initrdPath)
 	if err != nil {
 		return nil, "", fmt.Errorf("failed to open initrd (%s):\n%w", initrdPath, err)
--- a/toolkit/tools/internal/targetos/targetos.go
+++ b/toolkit/tools/internal/targetos/targetos.go
@@ -74,9 +74,24 @@ var (
 		Version:   []int{24, 4},
 	}
 
+	// OsReleaseCandidates lists the on-rootfs paths to probe for the os-release(5) file, in preference order.
 	OsReleaseFileCandidates = []string{
 		"/etc/os-release",
 		"/usr/lib/os-release",
+	}
+
+	// initrdReleaseCandidates lists the in-initrd paths to probe for an os-release(5)-type file, in preference order.
+	//
+	// os-release is the canonical filename and is the only candidate present in full-OS initrds built by Image
+	// Customizer (which pack the source rootfs directly, in which initrd-release does not exist at all).
+	//
+	// initrd-release is the dracut-runtime variant emitted by dracut's 99base module. os-release symlinks to this file
+	// in such cases, so it is a necessary fallback.
+	initrdReleaseCandidates = []string{
+		"/etc/os-release",
+		"/usr/lib/os-release",
+		"/etc/initrd-release",
+		"/usr/lib/initrd-release",
 	}
 )
 
@@ -90,11 +105,13 @@ func New(distroId Distro, versionId string) TargetOs {
 	}
 }
 
+// GetInstalledTargetOs reads the os-release(5) file from the given rootfs and resolves it to a TargetOs.
+//
+// Returns an error wrapping fs.ErrNotExist when none of the candidates exist on the rootfs.
 func GetInstalledTargetOs(rootfs string) (TargetOs, error) {
 	var err error
 	var fields map[string]string
 
-	found := false
 	for _, candidate := range OsReleaseFileCandidates {
 		fields, err = envfile.ParseEnvFile(filepath.Join(rootfs, candidate))
 		if errors.Is(err, fs.ErrNotExist) {
@@ -104,23 +121,45 @@ func GetInstalledTargetOs(rootfs string) (TargetOs, error) {
 			return TargetOs{}, fmt.Errorf("failed to read os-release file (%s):\n%w", candidate, err)
 		}
 
-		found = true
 		break
 	}
 
-	if !found {
+	if fields == nil {
 		return TargetOs{}, fmt.Errorf("no os-release file found (candidates=%s):\n%w", OsReleaseFileCandidates, err)
 	}
 
-	targetOs, err := GetInstalledTargetOsFromEnvFields(fields)
+	targetOs, err := GetInstalledTargetOsFromEnvFields(fields, "os-release")
+	if err != nil {
+		return TargetOs{}, fmt.Errorf("failed to determine target OS from os-release file:\n%w", err)
+	}
+
+	return targetOs, nil
+}
+
+// GetInitrdTargetOs reads an os-release(5)-type file (os-release or dracut's initrd-release) from the given initrd and
+// resolves it to a TargetOs.
+//
+// Returns an error wrapping fs.ErrNotExist when none of the candidates exist in the initrd.
+func GetInitrdTargetOs(initrdPath string) (TargetOs, error) {
+	content, foundPath, err := readFirstFileFromInitrd(initrdPath, initrdReleaseCandidates)
 	if err != nil {
 		return TargetOs{}, err
 	}
 
-	return targetOs, err
+	fields, err := envfile.ParseEnv(string(content))
+	if err != nil {
+		return TargetOs{}, fmt.Errorf("failed to read (%s) file from initrd (%s):\n%w", foundPath, initrdPath, err)
+	}
+
+	targetOs, err := GetInstalledTargetOsFromEnvFields(fields, "initrd-release")
+	if err != nil {
+		return TargetOs{}, fmt.Errorf("failed to determine target OS from initrd-release file:\n%w", err)
+	}
+
+	return targetOs, nil
 }
 
-func GetInstalledTargetOsFromEnvFields(fields map[string]string) (TargetOs, error) {
+func GetInstalledTargetOsFromEnvFields(fields map[string]string, sourceLabel string) (TargetOs, error) {
 	distroId := fields["ID"]
 	versionId := fields["VERSION_ID"]
 
@@ -171,7 +210,7 @@ func GetInstalledTargetOsFromEnvFields(fields map[string]string) (TargetOs, erro
 		}, nil
 
 	default:
-		return TargetOs{}, fmt.Errorf("unknown ID (%s) in os-release", distroId)
+		return TargetOs{}, fmt.Errorf("unknown ID (%s) in %s", distroId, sourceLabel)
 	}
 }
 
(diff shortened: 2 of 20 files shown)


toolkit/tools/internal/targetos/targetos.go
```diff
@@ -171,7 +210,7 @@ func GetInstalledTargetOsFromEnvFields(fields map[string]string) (TargetOs, erro
 		}, nil
 
 	default:
-		return TargetOs{}, fmt.Errorf("unknown ID (%s) in os-release", distroId)
+		return TargetOs{}, fmt.Errorf("unknown ID (%s) in %s", distroId, sourceLabel)
```
cwize1: You could simplify this to just `"unknown ID (%s)"` and then let the caller add the context. That way, you can delete `sourceLabel`.

toolkit/tools/internal/initrdutils/initrdread.go
```diff
@@ -0,0 +1,120 @@
+// Copyright (c) Microsoft Corporation.
```
cwize1, line 1: Would it make sense to move this logic into the `initrdutils` package?

toolkit/tools/pkg/imagecustomizerlib/liveosisoutils.go
cwize1: Can we rename `bootloaderFilesConfig` so that it is more distro specific, to match `bootloaderFilesConfigFedora`?

toolkit/tools/pkg/imagecustomizerlib/distrohandler_fedora.go
```diff
@@ -29,14 +29,55 @@ type fedoraDistroHandler struct {
...
 	grubToolsPackageFedora     = "grub2-tools"
 	grubPcModulesPackageFedora = "grub2-pc-modules"
 )
 
+// bootloaderFilesConfigFedora is the boot-files map for Fedora-style ESPs (Azure Linux 4 and Fedora).
+var bootloaderFilesConfigFedora = map[string]BootFilesArchConfig{
+	"amd64": {
+		bootBinary:                  bootx64BinaryFedora,
+		grubBinary:                  grubx64Binary,
+		grubNoPrefixBinary:          grubx64NoPrefixBinary,
```
cwize1: I don't think Fedora has a `grubx64-noprefix.efi` file. It is kind of an AZL3 thing.

toolkit/tools/pkg/imagecustomizerlib/liveospxe.go
```diff
@@ -116,12 +116,13 @@ func createPXEArtifacts(buildDir string, outputFormat imagecustomizerapi.ImageFo
 	// created because some of these files are needed by the ISO so it is
 	// bootable.
 
-	// Move bootloader files from under '<pxe-folder>/efi/boot' to '<pxe-folder>/'
-	_, bootFilesConfig, err := getBootArchConfig()
+	// Move bootloader files from under '<outputPXEArtifactsDir>/<isoBootloaderDir>' to '<outputPXEArtifactsDir>/'.
+	bootFilesConfig, err := distroHandler.GetBootArchConfig()
```
cwize1, line 120: There is a build error for `distroHandler`

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -28,6 +28,14 @@ type LiveOSConfig struct {
 	bootstrapFileUrl  string
 }
 
+// initramfsTypeFromFilesStore detects the initramfs type based on the files in the provided file store.
+func initramfsTypeFromFilesStore(filesStore *IsoFilesStore) imagecustomizerapi.InitramfsImageType {
+	if filesStore == nil || filesStore.squashfsImagePath == "" {
```
cwize1: Why do you check `filesStore == nil` here? This feels wrong.

toolkit/tools/pkg/imagecustomizerlib/liveosisoartifactstore.go
```diff
@@ -448,7 +448,50 @@ func createIsoFilesStoreFromIsoImage(isoImageFile, storeDir string) (filesStore
...
 
-	_, bootFilesConfig, err := getBootArchConfig()
+	// Resolve the distro from the ISO's initrd. Which initrd to read depends on the ISO's initramfsType.
+	squashfsPath := filepath.Join(artifactsDir, liveOSImagePath)
+	squashfsExists, err := file.PathExists(squashfsPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to check for squashfs at (%s):\n%w", squashfsPath, err)
+	}
+	if squashfsExists {
+		filesStore.squashfsImagePath = squashfsPath
```
cwize1, line 474: It might be worth adding a comment saying that `initramfsTypeFromFilesStore` needs the `squashfsImagePath` value.

toolkit/tools/pkg/imagecustomizerlib/liveosisoartifactstore.go
```diff
@@ -448,7 +448,50 @@ func createIsoFilesStoreFromIsoImage(isoImageFile, storeDir string) (filesStore
...
+	}
+
+	var initrdPath string
+	switch initramfsTypeFromFilesStore(filesStore) {
```
cwize1: This `initramfsTypeFromFilesStore` call is just reading `squashfsExists` but in a really roundabout way.

toolkit/tools/internal/targetos/targetos.go
```diff
@@ -90,11 +105,13 @@ func New(distroId Distro, versionId string) TargetOs {
...
 }
 
+// GetInstalledTargetOs reads the os-release(5) file from the given rootfs and resolves it to a TargetOs.
+//
+// Returns an error wrapping fs.ErrNotExist when none of the candidates exist on the rootfs.
 func GetInstalledTargetOs(rootfs string) (TargetOs, error) {
 	var err error
 	var fields map[string]string
 
-	found := false
```
cwize1, line 97: IMHO removing `found` makes the code more confusing since it relies on `envfile.ParseEnvFile` returning `nil` on an error.

toolkit/tools/pkg/imagecustomizerlib/distrohandler.go
```diff
@@ -181,3 +185,14 @@ func handleUnsupportedDistroVersion(rc *ResolvedConfig, targetOs targetos.Target
...
+
+// NewDistroHandlerFromInitrd creates a distro handler by detecting the OS from the dracut-emitted initrd-release file
+// inside the initramfs at initrdPath. Used in ISO-to-ISO pipelines where no rootfs is mounted yet but a canonical
+// distro identity is needed to pick the correct boot-file layout.
+func NewDistroHandlerFromInitrd(initrdPath string) (DistroHandler, error) {
+	targetOs, err := targetos.GetInitrdTargetOs(initrdPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to determine the target OS from initrd (%s):\n%w", initrdPath, err)
+	}
+	return NewDistroHandler(targetOs)
```
cwize1: Add the `if err != nil {`, since it makes it easier to modify the function in the future.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/distrohandler_fedora.go
vinceaperri: What is the suggestion, to remove it from this struct?

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
vinceaperri: Why does it feel wrong? It's a pointer argument.

toolkit/tools/pkg/imagecustomizerlib/distrohandler.go
vinceaperri: Done.

toolkit/tools/internal/initrdutils/initrdread.go
vinceaperri: Done.

toolkit/tools/internal/targetos/targetos.go
vinceaperri: Done.

toolkit/tools/pkg/imagecustomizerlib/liveosisoutils.go
vinceaperri: Done.

toolkit/tools/pkg/imagecustomizerlib/liveospxe.go
vinceaperri: Done.

toolkit/tools/pkg/imagecustomizerlib/liveosisoartifactstore.go
vinceaperri: Done.

toolkit/tools/pkg/imagecustomizerlib/liveosisoartifactstore.go
vinceaperri: Done.

toolkit/tools/internal/targetos/targetos.go
vinceaperri: Done.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_fedora.go
cwize1: Possibly something like set the value to `""` and then handle `""` where `grubNoPrefixBinary` is used.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
cwize1: Under what circumstances would `filesStore` be `nil`? And why is `"full-os"` the correct default?

I don't think there is a good default for the initramfs type. Either it is a bootstrap initramfs or a full-os initramfs. If we can't figure it out, then we should probably fail.

vinceaperri: Good point, it can't actually be nil here. The only caller, `resolveInitramfsType`, already guards with `if inputArtifactsStore != nil`, and both constructors (`createIsoArtifactStoreFromMountedImage` and `createIsoArtifactStoreFromIsoImage`) return nil on error and set `.files` to a non-nil store before returning on success, so a returned `IsoArtifactsStore` always has a non-nil `files`. The pre-existing code even accessed `inputArtifactsStore.files.squashfsImagePath` directly without a nil check, so the `filesStore == nil` branch I added is unreachable. I'll drop the nil check.

cwize1 (approved):

PR #791, opened 2026-06-19
Remove the container's run.sh script.

cwize1:
The oras API has been included in Image Customizer since v1.1. This is the replacement for the `run.sh` script functionality. There have now been 4 releases with both APIs. So, it is finally time to remove the old one.

PR #784 forgot to also modify the `run.sh` script. But instead of fixing it, this change just removes it.

This change also removes `test-container.sh`, which tests `run.sh`.

---

### **Checklist**

- [x] Tests added/updated


toolkit/tools/imagecustomizer/container/build-container.sh
```diff
@@ -81,15 +81,13 @@ stagingLibDir="${containerStagingFolder}/usr/lib/imagecustomizer"
 stagingLicensesDir="${containerStagingFolder}/usr/share/licenses"
 
 dockerFile="$scriptDir/imagecustomizer.Dockerfile"
```
vinceaperri, line 83: `run.sh` was the only in-container user of the `oras` and `jq` CLIs (imagecustomizer uses the oras-go library), so should they be dropped from the Dockerfile's `tdnf install` too?

cwize1: Good catch! Fixed

vinceaperri (approved):

PR #798, opened 2026-06-23
Use toolsChroot through IsPackageInstalled to unblock ACL UKI validation

liulanze:
ACL images ship no in-image tdnf/rpm, so the default `IsPackageInstalled`
(which shells out inside the image chroot) always returned false. That
blocked `uki.mode: create` and verity validation, which both rely on
`IsPackageInstalled` to check that `systemd-boot` / `device-mapper` are
present.


toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
```diff
@@ -110,8 +114,38 @@ func (d *aclDistroHandler) ManagePackages(ctx context.Context, buildDir string,
 		snapshotTime, d.packageManager)
 }
 
+// IsPackageInstalled checks the image's rpm database directly. ACL images
+// have no in-image tdnf/rpm binary, so the default chroot-shell-out
+// implementation cannot answer this query.
 func (d *aclDistroHandler) IsPackageInstalled(imageChroot safechroot.ChrootInterface, packageName string) bool {
-	return d.packageManager.isPackageInstalled(imageChroot, packageName)
```
cwize1, line 114: Is it not possible to use `--tools-dir` for this?

You'll need to plumb the `toolsChroot` to `isPackageInstalled`. But that should be doable.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -606,11 +606,11 @@ func detectBootloaderType(distroHandler DistroHandler, imageChroot safechroot.Ch
 ) (BootloaderType, string, error) {
 	for _, pkg := range grubEfiPackages {
 		logger.Log.Debugf("Checking if package (%s) is installed", pkg)
-		if distroHandler.IsPackageInstalled(imageChroot, pkg) {
+		if distroHandler.IsPackageInstalled(imageChroot, nil, pkg) {
```
cwize1: You should probably plumb `toolsChroot` to here as well, since this will likely be needed at some point.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
```diff
@@ -87,17 +88,24 @@ func (d *aclDistroHandler) checkForUnsupportedApis(rc *ResolvedConfig) error {
...
 		}
 	}
 
+	// ACL has no in-image package manager: every operation that needs to query or modify the rpm database has to
+	// happen through tdnf inside --tools-dir. That includes package ops (install/remove/update), UKI create
+	// (validates systemd-boot is installed), and verity (validates device-mapper is installed).
+	ukiCreate := rc.Uki != nil && rc.Uki.Mode == imagecustomizerapi.UkiModeCreate
+	verityEnabled := len(rc.Storage.Verity) > 0
+	if (pkgOps || ukiCreate || verityEnabled) && rc.Options.ToolsDir == "" {
+		return fmt.Errorf("ACL requires --tools-dir for package, UKI 'create', or verity operations: " +
```
cwize1: `or` -> `and`

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
```diff
@@ -110,8 +118,17 @@ func (d *aclDistroHandler) ManagePackages(ctx context.Context, buildDir string,
...
 
-func (d *aclDistroHandler) IsPackageInstalled(imageChroot safechroot.ChrootInterface, packageName string) bool {
-	return d.packageManager.isPackageInstalled(imageChroot, packageName)
+// IsPackageInstalled queries the image's rpm database via tdnf running inside toolsChroot. ACL images ship no
+// in-image tdnf/rpm, so the check requires --tools-dir (enforced up-front in checkForUnsupportedApis).
+func (d *aclDistroHandler) IsPackageInstalled(imageChroot safechroot.ChrootInterface,
+	toolsChroot *safechroot.Chroot, packageName string,
+) bool {
+	if toolsChroot == nil {
+		logger.Log.Warnf("ACL IsPackageInstalled called without --tools-dir; cannot query rpmdb for package (%q)",
```
cwize1: Why post a warning instead of erroring out?

toolkit/tools/pkg/imagecustomizerlib/liveosisoartifactstore.go
```diff
@@ -427,7 +427,7 @@ func createIsoInfoStoreFromMountedImage(buildDir string, imageRootDir string, di
 	// Note the MIC allows the user to install other selinux policy packages.
 	// So, the absence of selinux-policy does not mean that there are no selinux
 	// policy packages.
-	if distroHandler.IsPackageInstalled(chroot, "selinux-policy") {
+	if distroHandler.IsPackageInstalled(chroot, nil, "selinux-policy") {
```
cwize1: It would be good to plumb `toolsChroot` to `createIsoInfoStoreFromMountedImage` and `createBootstrapInitrdImage` to avoid needing to fix this later.

Though, you may need to move where `toolsChroot` is initialized.

toolkit/tools/pkg/imagecustomizerlib/baseconfigs_test.go
```diff
@@ -179,14 +179,14 @@ func TestBaseConfigsFullRun(t *testing.T) {
...
-	sshdInstalled := distroHandler.IsPackageInstalled(imageConnection.Chroot(), "openssh-server")
+	sshdInstalled := distroHandler.IsPackageInstalled(imageConnection.Chroot(), nil, "openssh-server")
 	assert.True(t, sshdInstalled)
 
 	systemdBootPkgName := "systemd-boot"
 	if baseImageInfo.Version == baseImageVersionAzl4 {
 		systemdBootPkgName = "systemd-boot-unsigned"
 	}
-	systemdBootInstalled := distroHandler.IsPackageInstalled(imageConnection.Chroot(), systemdBootPkgName)
+	systemdBootInstalled := distroHandler.IsPackageInstalled(imageConnection.Chroot(), nil, systemdBootPkgName)
```
vinceaperri: The `Test` commit only updates call sites to pass `nil`. Can the new ACL path (toolsChroot != nil, uki create with --tools-dir) be covered too?

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
```diff
@@ -87,17 +88,24 @@ func (d *aclDistroHandler) checkForUnsupportedApis(rc *ResolvedConfig) error {
...
 
 		if os.Overlays != nil {
 			return fmt.Errorf("overlays are not yet supported for ACL")
 		}
 	}
 
+	// ACL has no in-image package manager: every operation that needs to query or modify the rpm database has to
+	// happen through tdnf inside --tools-dir. That includes package ops (install/remove/update), UKI create
+	// (validates systemd-boot is installed), and verity (validates device-mapper is installed).
+	ukiCreate := rc.Uki != nil && rc.Uki.Mode == imagecustomizerapi.UkiModeCreate
```
vinceaperri: Unspecified `uki.mode` (with `os.uki` set) also creates UKIs and reaches ValidateUkiDependencies, but this only matches `create`, so ACL without --tools-dir slips past this gate and fails later with the confusing "systemd-boot is not installed". Should this be `Mode != passthrough && Mode != modify` instead of `Mode == create`?

vinceaperri (changes requested):

toolkit/tools/pkg/imagecustomizerlib/baseconfigs_test.go
liulanze: The ACL path needs a base image fixture in CI, which isn't in place yet. I'll add proper coverage in a follow-up PR once the ACL CI fixture lands. For this PR, the wiring is now plumbed end-to-end so the new test will be a drop-in.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
liulanze: Fixed, thanks!

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
liulanze: Fixed, thanks.

toolkit/tools/pkg/imagecustomizerlib/artifactsinputoutput.go
```diff
@@ -478,7 +478,7 @@ func prepareImageConversionData(ctx context.Context, rawImageFile string, buildD
 		return nil, nil, "", nil, [randomization.UuidSize]byte{}, "", nil, nil, err
 	}
 
-	osPackages, cosiBootMetadata, err := collectOSInfoHelper(ctx, buildDir, imageConnection, distroHandler)
+	osPackages, cosiBootMetadata, err := collectOSInfoHelper(ctx, buildDir, imageConnection, distroHandler, nil)
```
cwize1, line 481: At some point, I guess we'll want to expose `--tools-dir` on `convert` as well. Though, you can punt this to a future change.

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -386,8 +386,8 @@ func getAllPackagesFromChroot(imageConnection *imageconnection.ImageConnection,
 	return distroHandler.GetAllPackagesFromChroot(imageConnection.Chroot())
 }
 
-func extractCosiBootMetadata(buildDirAbs string, imageConnection *imageconnection.ImageConnection, distroHandler DistroHandler) (*CosiBootloader, error) {
-	bootloaderType, err := distroHandler.DetectBootloaderType(imageConnection.Chroot())
+func extractCosiBootMetadata(buildDirAbs string, imageConnection *imageconnection.ImageConnection, distroHandler DistroHandler, toolsChroot *safechroot.Chroot) (*CosiBootloader, error) {
```
cwize1: Line is a little long

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -783,15 +786,15 @@ func collectOSInfo(ctx context.Context, buildDir string, rawImageFile string, pa
 	return osPackages, cosiBootMetadata, nil
 }
 
-func collectOSInfoHelper(ctx context.Context, buildDir string, imageConnection *imageconnection.ImageConnection, distroHandler DistroHandler) ([]OsPackage, *CosiBootloader, error) {
+func collectOSInfoHelper(ctx context.Context, buildDir string, imageConnection *imageconnection.ImageConnection, distroHandler DistroHandler, toolsChroot *safechroot.Chroot) ([]OsPackage, *CosiBootloader, error) {
```
cwize1: Line is a little long

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -138,7 +139,7 @@ func createLiveOSFromRaw(ctx context.Context, buildDir string, inputArtifactsSto
 		return fmt.Errorf("failed to build live OS configuration from input configuration:\n%w", err)
 	}
 
-	err = createLiveOSFromRawHelper(ctx, buildDir, inputArtifactsStore, requestedSelinuxMode, liveosConfig, rawImageFile, outputFormat, outputPath, distroHandler)
+	err = createLiveOSFromRawHelper(ctx, buildDir, inputArtifactsStore, requestedSelinuxMode, liveosConfig, rawImageFile, outputFormat, outputPath, distroHandler, toolsChroot)
```
cwize1: Line is a little long

toolkit/tools/pkg/imagecustomizerlib/liveosisoartifactstore.go
```diff
@@ -427,7 +428,11 @@ func createIsoInfoStoreFromMountedImage(buildDir string, imageRootDir string, di
 	// Note the MIC allows the user to install other selinux policy packages.
 	// So, the absence of selinux-policy does not mean that there are no selinux
 	// policy packages.
-	if distroHandler.IsPackageInstalled(chroot, "selinux-policy") {
+	selinuxPolicyInstalled, err := distroHandler.IsPackageInstalled(chroot, toolsChroot, "selinux-policy")
```
cwize1: You probably need to ensure the `chroot` is placed in the `toolsRootImageDir` directory under the `toolsChroot`. Otherwise, this code won't work.

It might make sense to leave the liveos stuff to a future change.

Similar comment for `createBootstrapInitrdImage`.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/artifactsinputoutput.go
liulanze: Sure, the timeline for this project is too tight, will not fix this in this PR then.

toolkit/tools/pkg/imagecustomizerlib/baseconfigs_test.go
cwize1: You should add some AZL3 tests that use `--tools-dir`. That way, at least some of the logic in this PR is being exercised in tests.

toolkit/tools/pkg/imagecustomizerlib/customizepackages_test.go
```diff
@@ -948,6 +948,56 @@ func testCustomizeImagePackagesInstallOnline(t *testing.T, baseImageInfo testBas
 	ensurePackageCacheCleanup(t, imageConnection, baseImageInfo)
 }
 
+func TestCustomizeImagePackagesAddWithToolsDirAzl3(t *testing.T) {
```
cwize1: This is a good test to have. But considering this change is targeting fixing UKI, you should add a UKI test as well.

Considering that we have a bunch of UKI tests already, you could modify one to use `--tools-dir`. Testing multiple things in a single test is goodness, since it reduces to the total number of tests.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/baseconfigs_test.go
vinceaperri: Could you verify it locally with ACL?

cwize1 (dismissed):

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -507,7 +532,7 @@ func customizeOSContents(ctx context.Context, rc *ResolvedConfig) (imageMetadata
 	var cosiBootMetadata *CosiBootloader
 	if rc.OutputImageFormat == imagecustomizerapi.ImageFormatTypeCosi || rc.OutputImageFormat == imagecustomizerapi.ImageFormatTypeBareMetalImage {
 		osPackages, cosiBootMetadata, err = collectOSInfo(ctx, rc.BuildDirAbs, rc.RawImageFile, partitionsLayout,
-			im.distroHandler)
+			im.distroHandler, toolsChroot)
```
vinceaperri: Should this be `nil`, like the LiveOS callers that also mount outside the tools chroot? Otherwise, bootloader detection will fail.

vinceaperri (dismissed):

toolkit/tools/pkg/imagecustomizerlib/baseconfigs_test.go
liulanze: yes, verified locally, the IsPackageInstalled toolsChroot path is exercised by package install on an ACL VHD.

cwize1 (approved):

PR #802, opened 2026-06-23
Fix `TestCopyAdditionalFiles` test.

cwize1:
The file permissions verification was accidentally not doing anything.

---

### **Checklist**

- [x] Tests added/updated


liulanze (dismissed):

toolkit/tools/pkg/imagecustomizerlib/customizefiles_test.go
vinceaperri: On mismatch this reports bare octal modes with no path context. Use `assert.Equalf` with both paths like `verifyFileContentsSame` so failures show which files differ. Now that this will be checked, it's relevant.

vinceaperri (changes requested):

toolkit/tools/pkg/imagecustomizerlib/customizefiles_test.go
cwize1: Fixed

vinceaperri (approved):
Nice catch!

PR #805, opened 2026-06-25
Fix build directory when using tools directory.

cwize1:
When using the tools directory, the image needs to be mounted in the tools directory so that the tools directory chroot has access to the image mounts. But the tools directory should not be used as the build directory. This change fixes this.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
@@ -197,10 +197,12 @@ func createLiveOSFromRawHelper(ctx context.Context, buildDir string, inputArtifa
 		}
 	}()
 
+	imageMountPoint := filepath.Join(buildDir, "readonly-rootfs-mount")
+
 	logger.Log.Debugf("Connecting to raw image (%s)", rawImageFile)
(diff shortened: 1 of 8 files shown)


liulanze (dismissed):

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
```diff
@@ -197,10 +197,12 @@ func createLiveOSFromRawHelper(ctx context.Context, buildDir string, inputArtifa
 		}
 	}()
 
+	imageMountPoint := filepath.Join(buildDir, "readonly-rootfs-mount")
```
vinceaperri: Previously this mounted at `isoBuildDir/readonly-rootfs-mount`. Is this switch intentional?

vinceaperri (dismissed):

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder.go
cwize1: Fixed

vinceaperri (approved):

PR #806, opened 2026-06-25
Add preview features flag for --tools-dir.

cwize1:
The tools directory needs to be used in other places, including COSI generation, LiveOS generation, user/group modification, and SELinux labelling. So, place the API behind a preview feature flag so we can make behavior changes in the future, without it being considered a breaking change.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/imagecustomizerapi/previewfeaturetype.go
+++ b/toolkit/tools/imagecustomizerapi/previewfeaturetype.go
@@ -47,6 +47,9 @@ const (
 
 	// PreviewFeatureUnsupportedDistroVersion allows distro versions that are not supported yet.
 	PreviewFeatureUnsupportedDistroVersion PreviewFeature = "unsupported-distro-version"
+
+	// PreviewFeatureToolsDir enables support for specifying a tools directory.
+	PreviewFeatureToolsDir PreviewFeature = "tools-dir"
 )
 
 func (pf PreviewFeature) IsValid() error {
@@ -54,7 +57,7 @@ func (pf PreviewFeature) IsValid() error {
 	case PreviewFeatureUki, PreviewFeatureOutputArtifacts, PreviewFeatureInjectFiles, PreviewFeatureReinitializeVerity,
 		PreviewFeaturePackageSnapshotTime, PreviewFeatureKdumpBootFiles, PreviewFeatureDistroVersion,
 		PreviewFeatureBaseConfigs, PreviewFeatureInputImageOci, PreviewFeatureOutputSelinuxPolicy, PreviewFeatureBtrfs,
-		PreviewFeatureCreate, PreviewFeatureUnsupportedDistroVersion:
+		PreviewFeatureCreate, PreviewFeatureUnsupportedDistroVersion, PreviewFeatureToolsDir:
 		return nil
 	default:
 		return fmt.Errorf("invalid preview feature: %s", pf)
--- a/toolkit/tools/pkg/imagecustomizerlib/createimagevalidation.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/createimagevalidation.go
@@ -97,6 +97,7 @@ func validateCreateImageConfig(ctx context.Context, baseConfigPath string, confi
 			OutputImageFormat:   options.OutputImageFormat,
 			PackageSnapshotTime: options.PackageSnapshotTime,
 			BuildDir:            options.BuildDir,
+			ToolsDir:            options.ToolsDir,
 			PreviewFeatures:     options.PreviewFeatures,
 		})
 	if err != nil {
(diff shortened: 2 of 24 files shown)


toolkit/tools/pkg/imagecustomizerlib/createimagevalidation_test.go
```diff
@@ -240,6 +240,10 @@ func TestValidateCreateImageConfig_InvalidFieldsVerityConfig(t *testing.T) {
 		RpmsSources:       []string{testDir}, // Use the test directory as a dummy RPM source
 		OutputImageFormat: "vhdx",
 		BuildDir:          "./",
+		PreviewFeatures: []imagecustomizerapi.PreviewFeature{
```
vinceaperri, line 243: `validateCreateImageConfig` checks `config.PreviewFeatures` for `create` before `options.PreviewFeatures` is merged in, so this test fails at the create gate.

toolkit/tools/pkg/imagecustomizerlib/createimagevalidation_test.go
```diff
@@ -257,6 +261,10 @@ func TestValidateCreateImageConfig_InvalidFieldsOverlaysConfig(t *testing.T) {
...
```
vinceaperri, line 264: Same issue as the verity test, the create gate reads `config.PreviewFeatures`.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizeroptions.go
```diff
@@ -80,6 +80,12 @@ func (o *ImageCustomizerOptions) verifyPreviewFeatures(previewFeatures []imagecu
 		}
 	}
 
+	if o.ToolsDir != "" {
```
vinceaperri, line 83: `verity-usr-uki-azl3.yaml` and `verity-usr-uki-*-azl4.yaml` aren't updated, so `TestCustomizeImageVerityUsrUki` (which sets `--tools-dir`) now fails this check.

toolkit/tools/imagecustomizerapi/previewfeaturetype.go
```diff
@@ -47,14 +47,17 @@ const (
 
 	// PreviewFeatureUnsupportedDistroVersion allows distro versions that are not supported yet.
 	PreviewFeatureUnsupportedDistroVersion PreviewFeature = "unsupported-distro-version"
+
+	// PreviewFeatureToolsDir enables support for specifying a tools directory.
+	PreviewFeatureToolsDir PreviewFeature = "tools-dir"
```
vinceaperri, line 52: The `previewFeatures` blocks in the `create-image.md` examples need this preview feature.

vinceaperri (changes requested):

toolkit/tools/pkg/imagecustomizerlib/createimagevalidation_test.go
cwize1: Fixed

toolkit/tools/pkg/imagecustomizerlib/createimagevalidation_test.go
cwize1: Fixed

toolkit/tools/imagecustomizerapi/previewfeaturetype.go
cwize1: Fixed

toolkit/tools/pkg/imagecustomizerlib/imagecustomizeroptions.go
cwize1: Good catch! Fixed

toolkit/tools/pkg/imagecustomizerlib/createimagevalidation.go
```diff
@@ -97,12 +92,18 @@ func validateCreateImageConfig(ctx context.Context, baseConfigPath string, confi
...
 			PackageSnapshotTime: options.PackageSnapshotTime,
 			BuildDir:            options.BuildDir,
+			ToolsDir:            options.ToolsDir,
 			PreviewFeatures:     options.PreviewFeatures,
 		})
 	if err != nil {
 		return nil, err
 	}
 
+	if !slices.Contains(rc.PreviewFeatures, imagecustomizerapi.PreviewFeatureCreate) {
```
vinceaperri, line 102: A create run with no preview features now reports the `tools-dir` error before the `create` error. Is that intentional?

cwize1: It's more a side-effect of the easiest way to write the code. I don't think it causes too much pain since both preview flags have to be added.

vinceaperri (approved):

PR #808, opened 2026-06-26
Ignore fewer errors in DistroHandler.IsPackageInstalled.

cwize1:
For the `DistroHandler.IsPackageInstalled` implementations, don't ignore the error where the package query command fails to start. For example, when tdnf/dnf/dpkg-query is not installed.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/packagemanager_dnf.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/packagemanager_dnf.go
@@ -4,7 +4,9 @@
 package imagecustomizerlib
 
 import (
+	"errors"
 	"fmt"
+	"os/exec"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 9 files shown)


toolkit/tools/pkg/imagecustomizerlib/packagemanager_dnf.go
```diff
@@ -202,9 +204,17 @@ func (pm *dnfPackageManager) isPackageInstalled(imageChroot safechroot.ChrootInt
 		Chroot(chroot.ChrootDir()).
 		Execute()
 	if err != nil {
-		return false
+		var exitErr *exec.ExitError
```
vinceaperri, line 207: Create a shared helper for this exit-error handling? Currently duplicated across tdnf, dnf, and deb.

vinceaperri (approved):

toolkit/tools/pkg/imagecustomizerlib/packagemanager_dnf.go
cwize1: In the future, we may try to distinguish errors further. For example, not ignoring errors from failing to read the package database. If and when we do that, these error checks will probably diverge significantly. So, it probably isn't worth sharing code.

PR #809, opened 2026-06-26
Simple storage merging for base configs API.

cwize1:
1. Allow storage to be specified one time in the config tree. This covers a lot of uses cases, without the need for more complex merging logic.

2. Merge preview feature flags from base configs.

2. For the test config files that have distro variations, move common settings into a base config. This helps significantly reduce config duplication.

3. For the image creation test config files, always use a 3 GiB disk size, to avoid unneccessary differences between distros.

4. In the `TestOutputAndInjectArtifacts` and `TestOutputAndInjectArtifactsCosi` test, don't create a temporary copy of the config file. This makes it easier to use the base configs API in the config files.

5. For the `create-minimal-os` config files, move the common list of packages into a package list file.

---

### **Checklist**

- [x] Tests added/updated


toolkit/tools/pkg/imagecustomizerlib/configvalidation.go
```diff
@@ -260,8 +263,11 @@ func ValidateConfig(ctx context.Context, baseConfigPath string, config *imagecus
 }
 
 func resolveStorage(rc *ResolvedConfig) (imagecustomizerapi.Storage, error) {
+	foundStorage := imagecustomizerapi.Storage{}
+	foundStorageCount := 0
+
 	// Disallow storage fields in base configs until merging logic is implemented.
```
vinceaperri: Comment is stale. Storage in a base config is now allowed, as long as only one config sets it.

toolkit/tools/pkg/imagecustomizerlib/testdata/create-minimal-os-btrfs.yaml
```diff
@@ -1,71 +1,15 @@
...
-    - systemd-resolved
-    - systemd-udev
-    - tdnf
-    - tdnf-plugin-repogpgcheck
-    - util-linux
-    - zlib
-    - kernel
+
+    installLists:
+    - lists/azl-minimal-os.yaml
```
vinceaperri, line 15: The shared list adds e2fsprogs, which the btrfs image didn't install before. Intended?

toolkit/tools/pkg/imagecustomizerlib/createimagevalidation.go
vinceaperri, line 1: Now that storage can live in a base config, `validateCreateImageSupportedFields` only inspects the top-level `config.Storage`. The `resetPartitionsUuidsType` / `storage.verity` checks run against the resolved `rc.Storage` so a base config can't bypass the checks.

vinceaperri (changes requested):

toolkit/tools/pkg/imagecustomizerlib/createimagevalidation.go
cwize1: Fixed

toolkit/tools/pkg/imagecustomizerlib/configvalidation.go
cwize1: Fixed

toolkit/tools/pkg/imagecustomizerlib/testdata/create-minimal-os-btrfs.yaml
cwize1: Fixed

vinceaperri (approved):

PR #816, opened 2026-06-30
Add Azure Linux 4.0 ISO LiveOS bootstrap support

vinceaperri:
This enables Azure Linux 4.0 ISO (LiveOS) bootstrap and full-OS output. It was previously gated off because Azure Linux 4.0 uses Fedora-style BLS bootloader packaging that the LiveOS ISO builder did not handle, and this change adds distro-aware LiveOS hooks on `DistroHandler` to support it.

**What changed**
- Distro-aware LiveOS hooks on `DistroHandler` (required packages, grub EFI prefix directory, initrd dracut modules) for acl, azurelinux, azurelinux4, fedora, and ubuntu, plus the BLS grub.cfg and PXE boot-config helpers.
- BLS-layout builder fixes: preserve each loader entry's `boot/loader/entries` subpath when staging, skip dangling module symlinks such as `symvers` when building the bootstrap initrd, write a grub prefix redirector so Azure Linux 4.0's `EFI/azurelinux`-prefixed grub finds its config, preserve each grub `search --set=<target>` when relabeling for the ISO volume id, and fall back to the input ISO's distro handler when repackaging an ISO without OS customization.
- Omit the `selinux` dracut module for Fedora and Azure Linux 4.0, whose policy loads after `switch_root`.
- Add the `dracut-live` package to the Azure Linux 4.0 bootstrap config and unskip the Azure Linux 4.0 ISO vmtests.

PXE output stays explicitly unsupported for Azure Linux 4.0. That follows in a separate change.

---

### **Checklist**
- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/blsutils.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/blsutils.go
@@ -263,3 +263,119 @@ func updateBLSEntryOptions(content string, argsToRemove []string, newArgs []stri
 
 	return result
 }
+
+// forEachBLSEntry rewrites every non-rescue BLS .conf file under {bootDir}/loader/entries by passing its content
+// through edit.
+func forEachBLSEntry(bootDir string, edit func(content string) (string, error)) error {
+	entriesDir := filepath.Join(bootDir, "loader", "entries")
+	entries, err := os.ReadDir(entriesDir)
+	if err != nil {
+		return fmt.Errorf("failed to read BLS entries directory (%s):\n%w", entriesDir, err)
+	}
+
+	for _, entry := range entries {
+		if entry.IsDir() || !strings.HasSuffix(entry.Name(), ".conf") {
+			continue
+		}
+
+		absPath := filepath.Join(entriesDir, entry.Name())
+		contentBytes, err := os.ReadFile(absPath)
+		if err != nil {
+			return fmt.Errorf("failed to read BLS entry file (%s):\n%w", absPath, err)
+		}
+		content := string(contentBytes)
+
+		title := ""
+		for _, field := range bls.ParseFields(content) {
+			if field.Key == "title" {
+				title = field.Value
+				break
+			}
+		}
+		if bls.IsRescueEntryTitle(title) {
+			continue
+		}
+
+		updatedContent, err := edit(content)
+		if err != nil {
+			return fmt.Errorf("failed to update BLS entry file (%s):\n%w", absPath, err)
+		}
+
+		if updatedContent == content {
+			continue
+		}
+
+		err = os.WriteFile(absPath, []byte(updatedContent), 0o644)
+		if err != nil {
+			return fmt.Errorf("failed to write BLS entry file (%s):\n%w", absPath, err)
+		}
+	}
+
+	return nil
+}
+
+// setBLSEntryField replaces the value of the first line with the given key (e.g. "linux" or "initrd") in a BLS entry,
+// preserving every other byte. It returns an error if the key is absent unlike updateBLSEntryOptions above, which adds
+// an options line if none exists.
+func setBLSEntryField(content string, key string, newValue string) (string, error) {
+	for _, line := range bls.ParseLines(content) {
+		if line.Key != key {
+			continue
+		}
+		return content[:line.ContentStart] + key + " " + newValue + content[line.ContentEnd:], nil
+	}
+	return "", fmt.Errorf("failed to find the '%s' key in BLS entry", key)
+}
+
+// updateLiveOSBLSEntries applies the LiveOS-compatibility kernel-entry edits to the BLS entries under bootDir. It is
+// the BLS counterpart of the per-entry portion of updateGrubCfgForLiveOS.
+func updateLiveOSBLSEntries(bootDir string, initramfsImageType imagecustomizerapi.InitramfsImageType,
+	disableSELinux bool, savedConfigs *SavedConfigs,
+) error {
+	var argsToRemove []string
+	var argsToAppend []string
+
+	switch initramfsImageType {
+	case imagecustomizerapi.InitramfsImageTypeFullOS:
+		// Remove 'root' so that no pivoting takes place.
+		argsToRemove = append(argsToRemove, "root")
+	case imagecustomizerapi.InitramfsImageTypeBootstrap:
+		liveosKernelArgs := fmt.Sprintf(kernelArgsLiveOSTemplate, liveOSDir, liveOSImage)
+		argsToAppend = append(argsToAppend, strings.Fields(liveosKernelArgs)...)
+	default:
+		return fmt.Errorf("unsupported initramfs image type (%s)", initramfsImageType)
+	}
+
+	if disableSELinux {
+		argsToRemove = append(argsToRemove, selinuxArgNames...)
+		selinuxArgs, err := selinuxModeToArgs(imagecustomizerapi.SELinuxModeDisabled)
+		if err != nil {
+			return err
+		}
+		argsToAppend = append(argsToAppend, selinuxArgs...)
+	}
+
+	argsToAppend = append(argsToAppend, savedConfigs.LiveOS.KernelCommandLine.ExtraCommandLine...)
+
+	return forEachBLSEntry(bootDir, func(content string) (string, error) {
+		if initramfsImageType == imagecustomizerapi.InitramfsImageTypeFullOS {
+			var err error
+			content, err = setBLSEntryField(content, "initrd", isoInitrdPath)
+			if err != nil {
+				return "", fmt.Errorf("failed to update initrd path in BLS entry:\n%w", err)
+			}
+		}
+		return updateBLSEntryOptions(content, argsToRemove, argsToAppend), nil
+	})
+}
+
+// setLiveOSBLSEntriesRoot replaces the root= kernel arg in every BLS entry under bootDir and optionally appends
+// additional args.
+func setLiveOSBLSEntriesRoot(bootDir string, rootValue string, extraArgs []string) error {
+	argsToRemove := []string{"root"}
+	newArgs := append([]string{"root=" + rootValue}, extraArgs...)
+
+	return forEachBLSEntry(bootDir, func(content string) (string, error) {
+		return updateBLSEntryOptions(content, argsToRemove, newArgs), nil
+	})
+}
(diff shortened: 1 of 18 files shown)


toolkit/tools/pkg/imagecustomizerlib/blsutils.go
```diff
@@ -263,3 +263,119 @@ func updateBLSEntryOptions(content string, argsToRemove []string, newArgs []stri
...
+
+	return forEachBLSEntry(bootDir, func(content string) (string, error) {
+		if initramfsImageType == imagecustomizerapi.InitramfsImageTypeFullOS {
+			var err error
+			content, err = setBLSEntryField(content, "initrd", isoInitrdPath)
+			if err != nil {
+				return "", fmt.Errorf("failed to update initrd path in BLS entry:\n%w", err)
+			}
+		}
+		return updateBLSEntryOptions(content, argsToRemove, argsToAppend), nil
```
cwize1: nit:

```golang
content = updateBLSEntryOptions(content, argsToRemove, argsToAppend)
return content, nil
```

This mirrors the pattern seen in `setBLSEntryField`. So, it makes the code a little easier to understand at a glance.

toolkit/tools/pkg/imagecustomizerlib/blsutils.go
```diff
@@ -263,3 +263,119 @@ func updateBLSEntryOptions(content string, argsToRemove []string, newArgs []stri
...
+
+// setBLSEntryField replaces the value of the first line with the given key (e.g. "linux" or "initrd") in a BLS entry,
+// preserving every other byte. It returns an error if the key is absent unlike updateBLSEntryOptions above, which adds
+// an options line if none exists.
+func setBLSEntryField(content string, key string, newValue string) (string, error) {
+	for _, line := range bls.ParseLines(content) {
+		if line.Key != key {
+			continue
+		}
+		return content[:line.ContentStart] + key + " " + newValue + content[line.ContentEnd:], nil
```
cwize1, line 325: BLS entries are ended by newline, correct? Is this something we need to worry about?

toolkit/tools/pkg/imagecustomizerlib/distrohandler_azurelinux4.go
```diff
@@ -73,15 +73,16 @@ func (d *azureLinux4DistroHandler) ValidateConfig(rc *ResolvedConfig) error {
 	return nil
 }
 
+const azl4PxeUnsupportedError = "PXE output format is not supported yet for Azure Linux 4.0 images"
```
cwize1: Use `NewImageCustomizerError` and then use `assert.ErrorIs` in the test.

toolkit/tools/pkg/imagecustomizerlib/blsutils.go
cwize1, line 325: You don't need to fix this for this change: But it would be nice if we could deal with BLS in a purely declarative manner. I really dislike the code that handles the grub.cfg files. But the grub format is far too complex to fully parse. Whereas I think this is a lot more doable for BLS. Though, this would also require parsing and storing the comments and whitespace that are in the BLS file, so that we can write them back out.

toolkit/tools/pkg/imagecustomizerlib/liveosisoartifactstore.go
```diff
@@ -441,7 +441,8 @@ func createIsoInfoStoreFromMountedImage(buildDir string, imageRootDir string, di
 	return infoStore, nil
 }
 
-func createIsoFilesStoreFromIsoImage(isoImageFile, storeDir string) (filesStore *IsoFilesStore, err error) {
+func createIsoFilesStoreAndDistroHandlerFromIsoImage(isoImageFile, storeDir string,
```
cwize1: Why did you add `AndDistroHandler` to the name of `createIsoArtifactStoreAndDistroHandlerFromIsoImage` and `createIsoFilesStoreAndDistroHandlerFromIsoImage`?

Function names don't need to be all encompassing.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder_test.go
```diff
@@ -421,12 +413,42 @@ func VerifyBootstrapPXEArtifacts(t *testing.T, packageInfo *PackageVersionInform
...
+	assert.Regexp(t, pxeKernelIpArg, pxeKernelEntryContent)
+	assert.Regexp(t, pxeKernelRootArg, pxeKernelEntryContent)
+}
+
+// readLiveOSKernelEntryContent returns the text that carries the LiveOS kernel command line under artifactsPath,
+// along with the grub keyword that introduces it. For BLS distros (Azure Linux 4.0, Fedora) this is the concatenation
+// of the /boot/loader/entries/*.conf files and the keyword is "options"; otherwise it is the grub.cfg content and the
+// keyword is "linux".
+func readLiveOSKernelEntryContent(t *testing.T, artifactsPath string) (string, string) {
+	blsEntriesDir := filepath.Join(artifactsPath, "boot/loader/entries")
```
cwize1: Don't add dynamic assertion logic to tests. It can accidentally hide bugs in the future. The test should know if BLS entries are expected in the image or not.

toolkit/tools/pkg/imagecustomizerlib/liveosisoimages.go
```diff
@@ -46,9 +47,7 @@ hostonly="no"
 	usrLibLocaleDir = "/usr/lib/locale"
 )
 
-var (
-	kdumpInitramfsRegEx = regexp.MustCompile(`/initramfs-(.*)kdump\.img$`)
-)
+var kdumpInitramfsRegEx = regexp.MustCompile(`/initramfs-(.*)kdump\.img$`)
```
cwize1: `kdumpInitramfsRegEx` doesn't appear to be used

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/grubcfgutils.go
```diff
@@ -26,6 +26,9 @@ var (
 
 	// Finds the SELinux mode line in the /etc/selinux/config file.
 	selinuxConfigModeRegex = regexp.MustCompile(`(?m)^SELINUX=(\w+)$`)
+
+	// Captures the variable a grub 'search' command assigns via '--set=<var>' (e.g. root or boot).
+	grubSearchSetTargetRegex = regexp.MustCompile(`\s+--set=(\w+)`)
```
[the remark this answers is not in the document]
vinceaperri: Done.

toolkit/tools/pkg/imagecustomizerlib/blsutils.go
vinceaperri: Done.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_azurelinux4.go
vinceaperri: Done.

toolkit/tools/pkg/imagecustomizerlib/liveosisoartifactstore.go
vinceaperri: Reverted.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder_test.go
vinceaperri: Done.

toolkit/tools/pkg/imagecustomizerlib/liveosisoimages.go
vinceaperri: Removed.

toolkit/tools/pkg/imagecustomizerlib/liveosisogrub.go
```diff
@@ -184,16 +217,21 @@ func updateGrubCfgForPxe(inputContentString string, initramfsImageType imagecust
...
 	if err != nil {
 		return err
 	}
 
+	bootDir := filepath.Dir(filepath.Dir(inputGrubCfgPath))
+
 	// Update grub.cfg content to be 'live-os compatible'.
-	liveosContentString, err := updateGrubCfgForLiveOS(inputContentString, initramfsImageType, disableSELinux, savedConfigs, kernelVersions)
+	liveosContentString, err := distroHandler.UpdateLiveOSGrubCfgForLiveOS(inputContentString, bootDir,
+		initramfsImageType, disableSELinux, savedConfigs, kernelVersions)
```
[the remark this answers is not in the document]
vinceaperri: This PR enables ISO output for Azure Linux 4.0 and explicitly rejects AZL4 PXE.

toolkit/tools/pkg/imagecustomizerlib/blsutils.go
vinceaperri: No, `bls.ParseLines` sets `ContentEnd` to one past the last non-whitespace byte of the line, so the trailing newline (and any whitespace) stays in the `content[line.ContentEnd:]` suffix. `setBLSEntryField` only replaces the bytes between `ContentStart` and `ContentEnd`, so the line terminator is preserved.

vinceaperri: Agreed.

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder_test.go
```diff
@@ -468,15 +493,11 @@ func testCustomizeImageLiveOSKeepKdumpFilesA(t *testing.T, baseImageInfo testBas
...
-	if baseImageInfo.Version == baseImageVersionAzl4 {
-		assert.ErrorContains(t, err, azl4IsoPxeUnsupportedError)
-		return
-	}
 	if !assert.NoError(t, err) {
 		return
 	}
 
-	ValidateIsoContent(t, configA0, testTempDir, imagecustomizerapi.InitramfsImageTypeFullOS, outImageFilePath)
+	ValidateIsoContent(t, configA0, testTempDir, imagecustomizerapi.InitramfsImageTypeFullOS, outImageFilePath, baseImageInfo.Version)
```
cwize1: Pass the full `baseImageInfo` instead of just the distro version. That way, the code can easily be updated to also use distro as well.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/liveosisobuilder_test.go
vinceaperri: Done.

cwize1 (dismissed):

cwize1 (approved):

PR #817, opened 2026-06-30
Allow creation of tools directory for all distros.

cwize1:
Refactor the `download-test-utils.sh` script to allow for creation of the tools-directory for all distros, not just distros that have create support.

Also, remove the `-s` parameter from `download-test-utils.sh` and instead just always download the full set of RPMs for both create and customzie testing. If a distro doesn't have package lists for create testing, then skip that part and only download RPMs for customize testing.

Also, automatically skip tests that request offline RPMs or the tools directory for Ubuntu, since it isn't supported yet.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/internal/testutils/testrpms/download-test-utils.sh
+++ b/toolkit/tools/internal/testutils/testrpms/download-test-utils.sh
@@ -14,12 +14,11 @@ DISTRO_VERSION="3.0"
 CREATE_IMAGE="false"
 CONTAINER_REGISTRY=""
 
-while getopts "d:t:s:r:" flag
+while getopts "d:t:r:" flag
 do
     case "${flag}" in
         d) DISTRO="$OPTARG";;
         t) DISTRO_VERSION="$OPTARG";;
-        s) CREATE_IMAGE="$OPTARG";;
         r) CONTAINER_REGISTRY="$OPTARG";;
         h) ;;&
         ?) echo "Usage: download-test-utils.sh [-d DISTRO] [-t DISTRO_VERSION] [-s CREATE_IMAGE] [-r CONTAINER_REGISTRY]"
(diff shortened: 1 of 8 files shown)


toolkit/tools/internal/testutils/testrpms/download-test-utils.sh
```diff
@@ -14,12 +14,11 @@ DISTRO_VERSION="3.0"
 CREATE_IMAGE="false"
```
vinceaperri: `CREATE_IMAGE` is no longer read anywhere, so this can be removed.

toolkit/tools/internal/testutils/testrpms/download-test-utils.sh
```diff
@@ -14,12 +14,11 @@ DISTRO_VERSION="3.0"
...
-while getopts "d:t:s:r:" flag
+while getopts "d:t:r:" flag
 do
     case "${flag}" in
         d) DISTRO="$OPTARG";;
         t) DISTRO_VERSION="$OPTARG";;
-        s) CREATE_IMAGE="$OPTARG";;
         r) CONTAINER_REGISTRY="$OPTARG";;
         h) ;;&
         ?) echo "Usage: download-test-utils.sh [-d DISTRO] [-t DISTRO_VERSION] [-s CREATE_IMAGE] [-r CONTAINER_REGISTRY]"
```
vinceaperri: `-s` was removed from getopts, so this usage line and the `-s CREATE_IMAGE` description below are stale.

toolkit/tools/internal/testutils/testrpms/download-test-utils.sh
```diff
@@ -85,35 +84,34 @@ else
...
   fi
 
+# Handle package extraction for create subcommand testing.
+# Get configuration files for the distro-version.
+CONFIG_FILES="${DISTRO_CONFIG_MAP[${DISTRO}-${DISTRO_VERSION}]:-}"
+
+# Check if we have configuration for this distro.
+if [[ -z "$CONFIG_FILES" ]]; then
+  echo "Skipping package extraction."
+  echo "Unsupported distro '$DISTRO'"
```
vinceaperri: This prints `Unsupported distro 'azurelinux'` on a successful AZL2 run, which is misleading since only the create-test config is missing.

toolkit/tools/internal/testutils/testrpms/download-test-utils.sh
```diff
@@ -85,35 +84,34 @@ else
...
+$SCRIPT_DIR/create-tools-file.sh "$CONTAINER_IMAGE" "$TOOLS_FILE"
+echo "Tools file created successfully."
+
+TOOLS_DIR="${TOOLS_FILE%.tar.gz}-dir"
+echo "Extracting tools dir: $TOOLS_DIR"
+mkdir -p "$TOOLS_DIR"
+tar -x -z -f "$TOOLS_FILE" -C "$TOOLS_DIR"
+echo "Tools dir extracted successfully."
+
+# Check for python3 availability
```
vinceaperri: python3 is only used by the extraction loop, so should this check live below in the else block so distros without config files (e.g. AZL2) don't require it?

vinceaperri (changes requested):

toolkit/tools/internal/testutils/testrpms/download-test-utils.sh
cwize1: Fixed

toolkit/tools/internal/testutils/testrpms/download-test-utils.sh
cwize1: Fixed

toolkit/tools/internal/testutils/testrpms/download-test-utils.sh
cwize1: Fixed

toolkit/tools/internal/testutils/testrpms/download-test-utils.sh
cwize1: I removed this check since the `python3` call will fail if `python3` is not available.

vinceaperri (approved):

PR #818, opened 2026-06-30
Fix ISO->ISO customization with initrd.img distros.

cwize1:
The logic for finding initramfs files for ISO distro detection when `initrd.img-*` names are used is wrong.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/liveosisoartifactstore.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/liveosisoartifactstore.go
@@ -480,8 +480,7 @@ func createIsoFilesStoreFromIsoImage(isoImageFile, storeDir string) (filesStore 
 		// Any of the per-kernel dracut initrds is fine for distro detection since they're all from the same OS.
 		for _, f := range isoFiles {
 			base := filepath.Base(f)
-			isInitrd := ((strings.HasPrefix(base, initramfsPrefix) && strings.HasSuffix(base, ".img")) ||
-				strings.HasPrefix(base, initrdPrefix))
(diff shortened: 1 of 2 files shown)


toolkit/tools/pkg/imagecustomizerlib/liveosisoartifactstore.go
```diff
@@ -480,8 +480,9 @@ func createIsoFilesStoreFromIsoImage(isoImageFile, storeDir string) (filesStore
 		// Any of the per-kernel dracut initrds is fine for distro detection since they're all from the same OS.
 		for _, f := range isoFiles {
 			base := filepath.Base(f)
-			isInitrd := strings.HasPrefix(base, initramfsPrefix) || strings.HasPrefix(base, initrdPrefix)
-			if isInitrd && strings.HasSuffix(base, ".img") {
+			isInitrd := ((strings.HasPrefix(base, initramfsPrefix) && strings.HasSuffix(base, ".img")) ||
```
vinceaperri: `storeIfKernelSpecificFile` detects initrds by prefix alone. Is the `.img` suffix on the `initramfs-` branch needed here, or can it be dropped to match?

vinceaperri (changes requested):

toolkit/tools/pkg/imagecustomizerlib/liveosisoartifactstore.go
cwize1: I noticed that. But I wasn't sure why the .img check was added. So, I left it in out of caution. But I've now removed it.

vinceaperri (approved):

PR #822, opened 2026-07-02
Allow `--help` and `--version` to run unprivileged in container.

cwize1:
Allow the `--help` and `--version` commands to run without needing to pass `--privileged` to Docker. This will help with the usability of the container.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/test/vmtests/vmtests/imagecustomizer/test_help.py
+++ b/test/vmtests/vmtests/imagecustomizer/test_help.py
@@ -1,0 +1,32 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+from docker import DockerClient
+
+from ..utils.docker_utils import container_run
+
+
+# Ensure the --help command can run without the container being privileged.
+def test_container_help(
+    docker_client: DockerClient,
+    image_customizer_container_url: str,
+) -> None:
+    _, _ = container_run(
+        docker_client,
+        image_customizer_container_url,
+        ["--help"],
+        detach=True,
+    )
+
+
+# Ensure the --version command can run without the container being privileged.
+def test_container_version(
+    docker_client: DockerClient,
+    image_customizer_container_url: str,
+) -> None:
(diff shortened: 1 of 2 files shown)


toolkit/tools/imagecustomizer/container/entrypoint.sh
```diff
@@ -6,13 +6,15 @@
...
-        break
-    fi
+    case "$arg" in
+        "--disable-telemetry") ENABLE_TELEMETRY=false;;
+        "--help") HELP=true;;
+        "--version") HELP=true;;
+    esac
 done
 
 # Start telemetry service if enabled and connection string is set
```
vinceaperri, line 20: When a connection string is set, `--help`/`--version` still start the telemetry hopper and sleep 1s. Should telemetry be skipped when `HELP` is true, like the mount below?

test/vmtests/vmtests/imagecustomizer/test_help.py
```diff
@@ -0,0 +1,32 @@
...
+
+from ..utils.docker_utils import container_run
+
+
+# Ensure the --help command can run without the container being privileged.
+def test_container_help(
+    docker_client: DockerClient,
+    image_customizer_container_url: str,
+) -> None:
+    _, _ = container_run(
```
vinceaperri, line 14: Maybe explicitly pass `privileged=False` here and below too so we're not relying on docker defaults.

toolkit/tools/imagecustomizer/container/entrypoint.sh
cwize1: Fixed

test/vmtests/vmtests/imagecustomizer/test_help.py
cwize1: Fixed

vinceaperri (approved):

PR #823, opened 2026-07-02
Move COSI types into separate package.

cwize1:
Move the COSI metadata types into their own package. This will make it clearer when the code is using types that are specific to COSI.

This is being done in preparation of a new feature to create a package manifest, for use when the package tools and database are removed from the image.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
@@ -14,6 +14,7 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/cosiapi"
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/imagecustomizerapi"
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/file"
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/imageconnection"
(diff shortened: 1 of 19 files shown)


toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
```diff
@@ -601,25 +602,25 @@ func padToMegabyte(imageFile string) error {
 	return nil
 }
 
-// detectBootloaderType reports which bootloader is installed in imageChroot by probing for the presence of any
+// detectcosiapi.BootloaderType reports which bootloader is installed in imageChroot by probing for the presence of any
```
vinceaperri: Should be `detectBootloaderType`. The type rename rewrote the function name inside this comment.

toolkit/tools/pkg/imagecustomizerlib/extractpartitions_test.go
```diff
@@ -228,7 +229,7 @@ func TestAddSkippableFrame(t *testing.T) {
 	partitionFilepath, err := addSkippableFrame(tempPartitionFilepath, skippableFrameMetadata, partitionFilename, testDir)
 	assert.NoError(t, err)
 
-	// Verify decompression with skippable frame
+	// Verify decosiapi.Compression with skippable frame
```
vinceaperri: Should be `decompression`.

vinceaperri (dismissed):

toolkit/tools/pkg/imagecustomizerlib/cosicommon.go
cwize1: Thanks! Fixed

toolkit/tools/pkg/imagecustomizerlib/extractpartitions_test.go
cwize1: Thanks! Fixed

vinceaperri (approved):

PR #824, opened 2026-07-02
Docs: Fix comment for RPM repo priority.

cwize1:
The current docs claims the repos are prioritized in the order in which they are specified. However, this comment is incorrect. It was based on the early days of Azure Linux 2.0, where the tdnf version didn't support the repo priortization API of dnf. This changes fixes the docs.

Also, mention the `createrepo_c` command in the docs, for when a user wants to use a local RPMs directory in a `*.repo` file.

Also, unify the `--rpm-source` docs for `customize` and `create`.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/docs/imagecustomizer/api/cli/create.md
+++ b/docs/imagecustomizer/api/cli/create.md
@@ -97,11 +97,11 @@ Can be one of:
   environment and the URL will be replaced with the chroot equivalent URL. A
   `file+rel://` URL is handled similarly but the host path is relative to the repo
   file's parent directory. A `file+chroot://` URL refers to a file within the chroot
-  itself. If the tools chroot is being used (e.g. when using the [create](create.md)
-  subcommand), then this path is within the tools chroot instead of the OS chroot.
+  itself. If the tools chroot is being used, then this path is within the tools chroot
+  instead of the OS chroot.
 
-  If you use a repo file pointing to local directory containing RPM files, then you must
-  call `createrepo_c` (or `createrepo`) on the directory before using it as a repo:
+  If you use a repo file pointing to a local directory containing RPM files, then you
+  must call `createrepo_c` (or `createrepo`) on the directory before using it as a repo:
 
   ```bash
   createrepo_c --compatibility --update <rpms-directory>
--- a/docs/imagecustomizer/api/cli/create.md
+++ b/docs/imagecustomizer/api/cli/create.md
@@ -94,13 +94,24 @@ Can be one of:
 
   If the repo file's `baseurl` or `gpgkey` fields contain a `file://` URL, then the
   host's directories pointed to by the URL will be bind mounted into the chroot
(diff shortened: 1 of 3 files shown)


docs/imagecustomizer/api/cli/create.md
```diff
@@ -94,21 +94,33 @@ Can be one of:
 
   If the repo file's `baseurl` or `gpgkey` fields contain a `file://` URL, then the
   host's directories pointed to by the URL will be bind mounted into the chroot
-  environment and the URL will be replaced with the chroot equivalent URL.
+  environment and the URL will be replaced with the chroot equivalent URL. A
+  `file+rel://` URL is handled similarly but the host path is relative to the repo
+  file's parent directory. A `file+chroot://` URL refers to a file within the chroot
+  itself. If the tools chroot is being used (e.g. when using the [create](create.md)
```
vinceaperri: This links the create doc to itself.

docs/imagecustomizer/api/cli/create.md
```diff
@@ -94,21 +94,33 @@ Can be one of:
...
+  subcommand), then this path is within the tools chroot instead of the OS chroot.
+
+  If you use a repo file pointing to local directory containing RPM files, then you must
```
vinceaperri: `pointing to local directory` should be `pointing to a local directory` (same on line 207 of customize.md).

docs/imagecustomizer/api/cli/create.md
```diff
@@ -94,21 +94,33 @@ Can be one of:
...
+  ```bash
+  createrepo_c --compatibility --update <rpms-directory>
+  ```
 
   GPG signature checking is enabled by default.
   If you wish to disable GPG checking, then set both `gpgcheck` and `repo_gpgcheck` to
   `0` in the repo file.
 
-  The repo file will only be used during image creation and will not be added to
+  The repo file will only be used during image customization and will not be added to
```
vinceaperri, line 114: This is the `create` subcommand doc. Is `customization` intended here, or should it stay `creation`?

vinceaperri (changes requested):

docs/imagecustomizer/api/cli/create.md
cwize1: Fixed

docs/imagecustomizer/api/cli/create.md
cwize1: Fixed

docs/imagecustomizer/api/cli/create.md
cwize1: It's convenient to keep the create and customize texts identical so that it is easier to sync changes between them. I think in the context of this sentence, the word 'customization' is probably fine even for the create subcommand.

vinceaperri (approved):

PR #826, opened 2026-07-06
Add Azure Linux 4.0 PXE boot support for LiveOS images

vinceaperri:
A PXE-booted Azure Linux 4.0 image stopped at an empty grub menu, because AZL4 uses BLS (Boot Loader Specification) boot entries that grub cannot enumerate over TFTP. This change expands those BLS Type `#1` entries into explicit grub `menuentry` blocks when the PXE artifacts are built, so netboot works without grub reading `boot/loader/entries` over TFTP. It is wired per distro through new `DistroHandler.UpdateLiveOSGrubCfgForPxe` and `FinalizeLiveOSPxeBootConfig` hooks plus a shared BLS renderer, so other BLS distros benefit too.

Building the `pxe-tar` output also required `tarutils` to preserve symlinks (e.g. arm64 `/boot/dtb`) instead of dereferencing them, rejecting absolute or parent-traversal link targets on extraction, and to stop an inverted deferred-`Close` check from masking the real archive error.

**Tests** (in support of the above) — `test_pxe.py` adds bootstrap and full-OS PXE boot tests for AZL3 and AZL4 over a transient libvirt NAT network (dnsmasq DHCP and TFTP) plus an HTTP server over the extracted artifacts (`pxe_server.py`). They are skipped on Azure Linux hosts, whose `edk2-ovmf` lacks the UEFI network stack (EDK II NetworkPkg) and cannot netboot a VM, and run on Ubuntu hosts.

### DIFF
--- a/test/vmtests/vmtests/utils/libvirt_utils.py
+++ b/test/vmtests/vmtests/utils/libvirt_utils.py
@@ -262,6 +262,15 @@ def create_libvirt_domain_xml(libvirt_conn: libvirt.virConnect, vm_spec: VmSpec)
     video_model = ET.SubElement(video, "model")
     video_model.attrib["type"] = "vga"
 
+    # Provide a virtio RNG device. Azure Linux's OVMF builds its NetworkPkg drivers with a DXE dependency on
+    # EFI_RNG_PROTOCOL, which VirtioRngDxe only produces when a virtio-rng device is present. Without it the UEFI
+    # network stack never dispatches, so PXE/network boot silently falls through to the EFI shell.
+    rng = ET.SubElement(devices, "rng")
+    rng.attrib["model"] = "virtio"
+    rng_backend = ET.SubElement(rng, "backend")
+    rng_backend.attrib["model"] = "random"
+    rng_backend.text = "/dev/urandom"
+
     network_interface = ET.SubElement(devices, "interface")
     network_interface.attrib["type"] = "network"
 
--- a/test/vmtests/vmtests/utils/libvirt_utils.py
+++ b/test/vmtests/vmtests/utils/libvirt_utils.py
@@ -7,21 +7,51 @@ import os
 import platform
 import xml.etree.ElementTree as ET  # noqa: N817
 from pathlib import Path
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional
 
 import libvirt  # type: ignore
 
 
 class VmSpec:
     def __init__(
-        self, name: str, memory_mib: int, core_count: int, os_disk_path: Path, boot_type: str, secure_boot: bool
+        self,
+        name: str,
+        memory_mib: int,
+        core_count: int,
+        os_disk_path: Optional[Path],
+        boot_type: str,
+        secure_boot: bool,
+        pxe_boot: bool = False,
+        network_name: str = "default",
     ):
+        """Describe a libvirt VM for create_libvirt_domain_xml to render into a domain definition.
+
+        Args:
+            name: libvirt domain name.
+            memory_mib: Guest RAM, in MiB.
+            core_count: Number of vCPUs.
+            os_disk_path: OS disk image to attach (a .iso boots as a CD-ROM, anything else as a virtio disk). Must be
+                None when pxe_boot is True, and must be set otherwise.
+            boot_type: "efi" for UEFI boot.
+            secure_boot: Whether to enable UEFI Secure Boot.
+            pxe_boot: When True, attach no disk and boot from the network instead: the firmware PXE-boots off the NIC
+                on network_name.
+            network_name: The libvirt network the VM's NIC attaches to. For PXE this is also the network that serves
+                the boot artifacts (DHCP/TFTP/HTTP). Defaults to libvirt's "default" network.
+        """
+        if pxe_boot and os_disk_path is not None:
+            raise ValueError("os_disk_path must be None when pxe_boot is True")
+        if not pxe_boot and os_disk_path is None:
+            raise ValueError("os_disk_path is required when pxe_boot is False")
+
         self.name: str = name
         self.memory_mib: int = memory_mib
         self.core_count: int = core_count
-        self.os_disk_path: Path = os_disk_path
+        self.os_disk_path: Optional[Path] = os_disk_path
         self.boot_type: str = boot_type
         self.secure_boot: bool = secure_boot
+        self.pxe_boot: bool = pxe_boot
+        self.network_name: str = network_name
 
 
 def _get_domain_caps(
@@ -163,8 +193,6 @@ def create_libvirt_domain_xml(libvirt_conn: libvirt.virConnect, vm_spec: VmSpec)
     if vm_spec.boot_type == "efi":
         ET.SubElement(os_tag, "nvram")
 
-    os_boot = ET.SubElement(os_tag, "boot")
-
     if vm_spec.boot_type == "efi":
         loader = ET.SubElement(os_tag, "loader")
         loader.attrib["readonly"] = "yes"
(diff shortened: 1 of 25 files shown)


test/vmtests/vmtests/utils/pxe_server.py
```diff
@@ -0,0 +1,161 @@
...
+
+        network_xml = _build_pxe_network_xml(network_name, bridge_name, str(artifacts_dir), boot_loader_file)
+        logging.debug(f"Creating PXE libvirt network:\n{network_xml}")
+
+        try:
+            self._network = libvirt_conn.networkCreateXML(network_xml)
+
+            logging.debug(f"Starting PXE HTTP server on port {PXE_HTTP_PORT} serving ({artifacts_dir})")
+            self._http_log = open(http_log_file_path, "w", encoding="utf-8")
+            self._http_process = subprocess.Popen(
```
cwize1: Python has a [multiprocessing.Process](https://docs.python.org/3/library/multiprocessing.html#the-process-class) class. This behaves similarly to threads API but spawns a new process instead of a new thread. This may or may not be a better way of handling the HTTP server.

test/vmtests/vmtests/utils/pxe_server.py
```diff
@@ -0,0 +1,161 @@
...
+def _make_world_readable(root: Path) -> None:
+    os.chmod(root, 0o755)
+    for dir_path, dir_names, file_names in os.walk(root):
+        for dir_name in dir_names:
+            os.chmod(os.path.join(dir_path, dir_name), 0o755)
+        for file_name in file_names:
+            os.chmod(os.path.join(dir_path, file_name), 0o644)
+
+
+def _make_ancestors_traversable(leaf: Path) -> None:
```
cwize1: I can't say I'm fond of this. It is one thing messing with permissions of files the test creates. But reaching outside of that is probably not great.

cwize1 (changes requested):

test/vmtests/vmtests/utils/pxe_server.py
vinceaperri: Done, switched to `multiprocessing.Process` running an in-process `http.server`.

test/vmtests/vmtests/utils/pxe_server.py
vinceaperri: Good point. Reworked so the artifacts extract into a `/var/tmp` directory the test creates and owns. That is world-traversable, so no ancestor's permissions are touched anymore, and it is removed on close.

cwize1 (dismissed):

test/vmtests/vmtests/utils/pxe_server.py
cwize1: Out of interest, why do permissions matter? Does the HTTP server use the file permissions to know which files it can serve?

vinceaperri: No, the HTTP server doesn't gate on permissions. The server runs as root. libvirt's embedded dnsmasq, which serves the boot files over TFTP, runs unprivileged, so needs access.

toolkit/tools/internal/tarutils/tarutils.go
```diff
@@ -171,6 +181,17 @@
...
 			}
+		case tar.TypeSymlink:
+			if err := os.MkdirAll(filepath.Dir(target), 0755); err != nil {
+				return fmt.Errorf("failed to create parent folder for (%s)\n%w", target, err)
+			}
+			// Remove any pre-existing entry so re-extraction is idempotent (os.Symlink fails if the target exists).
+			if err := os.Remove(target); err != nil && !os.IsNotExist(err) {
+				return fmt.Errorf("failed to clear existing path (%s):\n%w", target, err)
+			}
+			if err := os.Symlink(header.Linkname, target); err != nil {
```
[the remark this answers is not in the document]
vinceaperri: Hardened the symlink case to validate the link target the same way the entry name is checked: absolute targets and any that traverse up out of the expansion root are rejected, so a link pointing outside it can't be created.

test/vmtests/vmtests/utils/pxe_server.py
```diff
@@ -0,0 +1,160 @@
...
+        try:
+            # Extract the artifacts under /var/tmp. dnsmasq (TFTP) and qemu run as their own unprivileged users, so
+            # every directory from the filesystem root down to the artifacts must be traversable by others. /var/tmp is
+            # world-traversable (1777) and on disk, unlike the tmpfs-backed /tmp (the artifacts include the full-OS
+            # image, which is too large for RAM).
+            artifacts_dir = Path(tempfile.mkdtemp(prefix="pxe-artifacts-", dir="/var/tmp"))
+            self._artifacts_dir = artifacts_dir
+            with tarfile.open(pxe_tar_path, "r:gz") as tar:
+                tar.extractall(artifacts_dir)
+
```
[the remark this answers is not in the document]
vinceaperri: Fixed — extraction now uses tarfile `filter="data"`, which rejects members with absolute paths, `..` traversal, or symlink/hardlink targets that escape the destination.

test/vmtests/vmtests/utils/pxe_server.py
```diff
@@ -0,0 +1,160 @@
...
+
+def _make_world_readable(root: Path) -> None:
+    os.chmod(root, 0o755)
+    for dir_path, dir_names, file_names in os.walk(root):
+        for dir_name in dir_names:
+            os.chmod(os.path.join(dir_path, dir_name), 0o755)
+        for file_name in file_names:
+            os.chmod(os.path.join(dir_path, file_name), 0o644)
+
```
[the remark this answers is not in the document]
vinceaperri: Fixed — `_make_world_readable` now skips symlinks so `chmod` can't follow a link to a target outside the extraction directory.

toolkit/tools/internal/tarutils/tarutils.go
cwize1, line 110: There is a [go-extract](https://github.com/hashicorp/go-extract) library which could be helpful, since that library handles a lot of the security protections already.

toolkit/tools/pkg/imagecustomizerlib/liveosisoimages.go
```diff
@@ -242,10 +242,19 @@ func stageLiveOSFile(stageDirPath string, stageFile StageFile) error {
...
+	// Skip dangling symlinks. Some kernel boot files (e.g. Azure Linux 4.0's symvers-<ver>.xz, a symlink into
+	// /lib/modules/<ver>/) dangle in the flattened artifacts directory. They are kernel module-build artifacts that
+	// are not needed to boot, and a dangling symlink on the boot media serves no purpose while potentially breaking
+	// consumers that extract the artifacts onto a real filesystem.
+	if info, lerr := os.Lstat(stageFile.sourcePath); lerr == nil && info.Mode()&os.ModeSymlink != 0 {
+		if _, serr := os.Stat(stageFile.sourcePath); serr != nil {
+			logger.Log.Debugf("Skipping dangling symlink while staging Live OS file (%s)", stageFile.sourcePath)
+			return nil
+		}
+	}
```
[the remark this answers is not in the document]
vinceaperri: Done.

toolkit/tools/pkg/imagecustomizerlib/liveospxe.go
```diff
@@ -112,6 +114,15 @@ func createPXEArtifacts(buildDir string, outputFormat imagecustomizerapi.ImageFo
...
+	// Finalize the pxe boot configuration. It must run after the embedded ISO above is sealed, since that ISO is built
+	// from the iso-flavored (root=live:LABEL) entries. For inline-grub distros this is a no-op
+	// (the pxe grub.cfg already carries the arg).
+	err = distroHandler.FinalizeLiveOSPxeBootConfig(filepath.Join(outputPXEArtifactsDir, "boot"), initramfsType,
+		bootstrapBaseUrl, bootstrapFileUrl)
+	if err != nil {
+		return fmt.Errorf("failed to finalize the PXE boot configuration:\n%w", err)
+	}
+
 	// Note that the moves/removes must take place afer the bootstrapped ISO is
```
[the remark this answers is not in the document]
vinceaperri: Fixed.

toolkit/tools/pkg/imagecustomizerlib/liveospxe.go
```diff
@@ -127,7 +138,10 @@ func createPXEArtifacts(buildDir string, outputFormat imagecustomizerapi.ImageFo
...
 	for _, bootloaderFile := range bootloaderFiles {
 		sourcePath := filepath.Join(bootloaderSrcDir, bootloaderFile)
-		targetPath := filepath.Join(outputPXEArtifactsDir, bootloaderFile)
+		// The PXE TFTP root uses the conventional lowercase boot loader names (bootx64.efi, grubx64.efi) documented in
+		// the PXE layout. Some distros name the ESP shim in uppercase (Fedora and Azure Linux 4 ship BOOTX64.EFI), so
+		// normalize the name at the PXE root.
+		targetPath := filepath.Join(outputPXEArtifactsDir, strings.ToLower(bootloaderFile))
 		err = file.Move(sourcePath, targetPath)
 		if err != nil {
 			return fmt.Errorf("failed to move boot loader file from (%s) to (%s) while generated the PXE artifacts folder:\n%w", sourcePath, targetPath, err)
```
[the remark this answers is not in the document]
vinceaperri: Fixed.

test/vmtests/vmtests/utils/libvirt_utils.py
```diff
@@ -238,38 +266,56 @@ def create_libvirt_domain_xml(libvirt_conn: libvirt.virConnect, vm_spec: VmSpec)
...
-            file_path=str(vm_spec.os_disk_path),
-            device_type="disk",
-            image_type="qcow2",
-            bus_type="virtio",
-            device_prefix="vd",
-            read_only=False,
-            next_disk_indexes=next_disk_indexes,
-        )
+    if vm_spec.pxe_boot:
+        # Provide a virtio RNG device. Azure Linux's OVMF builds its NetworkPkg drivers with a DXE dependency on
```
cwize1: I think the standard libvirt VM templates include a rng device as standard.  So, you could include this for all VMs.

test/vmtests/vmtests/utils/pxe_server.py
```diff
@@ -0,0 +1,198 @@
...
+            network_xml = _build_pxe_network_xml(network_name, bridge_name, str(artifacts_dir), boot_loader_file)
+            logging.debug(f"Creating PXE libvirt network:\n{network_xml}")
+            self._network = libvirt_conn.networkCreateXML(network_xml)
+
+            # Azure Linux 3.0's INPUT policy defaults to DROP, and libvirt opens DHCP, DNS and TFTP to the
+            # host for the networks it manages, but not HTTP, which is needed for the bootstrap initramfs PXE tests.
+            # Open that port strictly on the transient PXE bridge.
+            logging.debug(f"Opening HTTP port {PXE_HTTP_PORT} on the PXE bridge {bridge_name}")
+            _open_http_firewall(bridge_name, PXE_HTTP_PORT)
```
cwize1: Can we key the firewall rules on host distro / version?

For example, Fedora doesn't use iptables.

cwize1 (changes requested):

test/vmtests/vmtests/utils/libvirt_utils.py
vinceaperri: Done.

test/vmtests/vmtests/utils/pxe_server.py
vinceaperri: Done.

cwize1 (approved):

PR #832, opened 2026-07-08
Make lsof call more robust.

cwize1:
Image Customizer uses lsof to query for all the running processes in the chroot. But it has to supplement the lsof info with the processes' chroot value. However, sometimes the read of the chroot value fails with file not found. This might be because the process closed in-between the lsof call and the read against `/proc`.

The stop-running-processes operation is best effort anyway. So, just ignore this error to allow any other running processes to be stopped.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/internal/processes/processes.go
+++ b/toolkit/tools/internal/processes/processes.go
@@ -86,7 +86,13 @@ func GetProcessesUsingPath(path string) ([]ProcessRecord, error) {
 			}
 
 			record.ProcessRoot, err = os.Readlink(fmt.Sprintf("/proc/%d/root", record.ProcessId))
-			if err != nil && !os.IsNotExist(err) {
+			if err != nil {
+				if os.IsNotExist(err) {
+					// Process no longer exists.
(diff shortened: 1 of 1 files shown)


liulanze (dismissed):
lgtm

toolkit/tools/internal/processes/processes.go
```diff
@@ -86,7 +86,7 @@ func GetProcessesUsingPath(path string) ([]ProcessRecord, error) {
 			}
 
 			record.ProcessRoot, err = os.Readlink(fmt.Sprintf("/proc/%d/root", record.ProcessId))
-			if err != nil {
+			if err != nil && !os.IsNotExist(err) {
```
vinceaperri: A gone process returns an empty `ProcessRoot` and is then logged as a host process by `stopRunningProcesses`. Skip it here instead?

cwize1: Fixed

vinceaperri (approved):

PR #839, opened 2026-07-13
Populate COSI osPackages for ACL images when rpm db is unavailable

bfjelds:
Trident requires an os-packages list (even if it is empty):

| `osPackages` | [OsPackage](#ospackage-object)[]       | 1.0      | Yes (since 1.1) | The list of packages installed in the OS.        |

ImageCustomizer is specifying nil for the list currently for ACL.  This PR creates an empty list of os packages if rpm/rpmdb cannot be found.

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/distrohandler_test.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/distrohandler_test.go
@@ -10,8 +10,26 @@ import (
 	"testing"
 
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/imagecustomizerapi"
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/safechroot"
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/targetos"
 	"github.com/stretchr/testify/assert"
 )
+
+// Ensure the ACL package list is populated as an empty (non-nil) list when the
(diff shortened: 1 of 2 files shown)


frhuelsz (approved):

cwize1 (approved):

toolkit/tools/pkg/imagecustomizerlib/distrohandler_test.go
```diff
@@ -10,9 +10,27 @@ import (
...
 	"github.com/stretchr/testify/assert"
 )
 
+// Ensure the ACL package list is populated as an empty (non-nil) list when the
+// rpm database can't be read, so the COSI metadata's osPackages field
+// serializes as [] rather than null.
+func TestAclGetAllPackagesFromChrootReturnsEmptyListWhenNoRpm(t *testing.T) {
+	handler := newAclDistroHandler(targetos.TargetOsAzureContainerLinux3)
+
+	// An empty chroot dir has no rpm command and no rpm database.
```
vinceaperri, line 24: This test returns at the rpm-command-missing check, so the rpm-db-missing branch isn't exercised. Add a case where `rpm` exists but `/var/lib/rpm` doesn't?

bfjelds: Follow up PR with requested test: https://github.com/microsoft/azure-linux-image-tools/pull/841

PR #840, opened 2026-07-13
Trim trailing NUL padding from UKI .cmdline PE sections

vinceaperri:
Trims trailing NUL padding from a UKI addon's `.cmdline` PE section when reading it back, so re-customizing a UKI usr-verity image no longer aborts during base-image validation.

### What was happening
`extractCmdlineFromSinglePE` returned the `.cmdline` section's bytes verbatim. A `.cmdline` section holds a NUL-terminated string, but tools that rewrite it with shorter content (for example `objcopy --update-section`) keep the section's original size and zero-fill the freed bytes. The dumped command line therefore ended in trailing NULs, which corrupted parsing of the last argument -- notably the numeric `hash-offset` in `systemd.verity_usr_options` -- and aborted the re-customization.

### Fix
Trim trailing NUL bytes from the section contents when reading it back.

### Test
Adds an end-to-end regression test (`TestCustomizeImageVerityUsrUkiRecustomizeNulPaddedCmdline`) that builds an inline-usr-verity UKI image (inline verity is what emits the numeric `hash-offset`), rewrites the addon `.cmdline` with shorter NUL-terminated content to recreate the padding, then re-customizes and asserts success. The test and its `verity-usr-inline-uki-*` config family are parameterized across azl3 and azl4.

### Checklist
- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/customizeuki_test.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/customizeuki_test.go
@@ -8,11 +8,13 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"testing"
 
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/imagecustomizerapi"
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/imagegen/diskutils"
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/file"
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/shell"
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/testutils"
 	"github.com/stretchr/testify/assert"
 	"golang.org/x/sys/unix"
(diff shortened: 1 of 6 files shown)


toolkit/tools/pkg/imagecustomizerlib/customizeuki_test.go
```diff
@@ -662,3 +678,139 @@ func TestGetKernelNameFromUki(t *testing.T) {
...
+
+	// Build an inline-usr-verity UKI image whose addon command line ends with
+	// `systemd.verity_usr_options=...,hash-offset=<N>`.
+	ukiImageFilePath := filepath.Join(testTempDir, "uki-verity.raw")
+	configFile := filepath.Join(testDir, verityUsrInlineUkiConfigFile(t, baseImageInfo))
+	err := basicCustomizeImageWithConfigFile(t.Context(), buildDir, configFile, baseImage, ukiImageFilePath, "raw",
+		baseImageInfo.PreviewFeatures)
+	if !assert.NoError(t, err) {
+		return
+	}
```
cwize1, line 730: It might be nice to add some image validation here. For example, `verifyVerityUki` and co. That way, this test can test multiple things at once.

cwize1 (dismissed):

toolkit/tools/pkg/imagecustomizerlib/customizeuki_test.go
vinceaperri: Added.

cwize1 (approved):

PR #841, opened 2026-07-13
Adding more tests for ACL handling of COSI OsPackages

bfjelds:
Add test validating:
* that missing rpmdb behaves as expected
* that COSI metadata serializes empty os package list rather than omitting

Follow up to : https://github.com/microsoft/azure-linux-image-tools/pull/839


bfjelds:
@vinceaperri, here is PR with test for missing rpmdb.

vinceaperri (approved):

PR #842, opened 2026-07-13
Fix dnf log handling.

cwize1:
The handling of the dnf logs is pretty broken. In particular, unlike tdnf which prints all its logs to stdout, dnf splits its logs between stdout and stderr. So, handling dnf logs needs to be done in a different way to tdnf. Hence, this change moves the `executeRpmPackageManagerCommand` function into the `rpmPackageManagerHandler` interface (as `executeCommand`).

That being said, dnf logs aren't all that spammy (compared to tdnf). So, there isn't a need to do extensive filtering of the dnf logs. Therefore, this change also drastically simplifies the dnf log filtering logic.

Also, this change fixes an issue where the dnf transaction error logs are not printed as warning logs. (These can help the user diagnose package installation problems).

This change also moves the handling of the snapshot time into `tdnfPackageManager`.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/customizepackages_rpm.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/customizepackages_rpm.go
@@ -4,9 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"os"
 	"os/exec"
-	"path/filepath"
 	"strings"
 
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/cosiapi"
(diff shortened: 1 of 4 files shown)


toolkit/tools/pkg/imagecustomizerlib/customizepackages_rpm.go
```diff
@@ -21,29 +19,19 @@ import (
...
-		return fmt.Errorf("%w: package manager %s does not support snapshot time",
-			ErrSnapshotTimeNotSupported, pmHandler.getPackageManagerBinary())
-	}
-
-	// Setup distribution-specific configuration if needed
-	if pmHandler.supportsSnapshotTime() && snapshotTime != "" {
-		// Setup Azure Linux specific TDNF configuration with snapshot
-		err = createTempTdnfConfigWithSnapshot(packageManagerChroot, snapshotTime)
+	if snapshotTime != "" {
+		cleanup, err := pmHandler.configureSnapshotTime(packageManagerChroot, snapshotTime)
```
vinceaperri: `cleanup, err :=` shadows the named return `err`, so the deferred assignment updates the inner `err` and the cleanup error is never returned.

vinceaperri (changes requested):

toolkit/tools/pkg/imagecustomizerlib/customizepackages_rpm.go
cwize1: Thanks! Fixed

toolkit/tools/pkg/imagecustomizerlib/packagemanager_tdnf.go
vinceaperri, line 1: The comments in `isPackageInstalled` and `getPackageInformation` say "Matches the executeRpmPackageManagerCommand pattern", but this PR removes that function.

vinceaperri (changes requested):

toolkit/tools/pkg/imagecustomizerlib/packagemanager_tdnf.go
cwize1: Fixed

toolkit/tools/pkg/imagecustomizerlib/packagemanager_dnf.go
```diff
@@ -25,162 +25,54 @@ func newDnfPackageManager(version string) *dnfPackageManager {
...
+	toolsChroot *safechroot.Chroot,
+) error {
+	pmChroot := imageChroot
+	if toolsChroot != nil {
+		pmChroot = toolsChroot
+	}
 
-func (pm *dnfPackageManager) createOutputCallback() func(string) {
-	dnfDownloadRegex := regexp.MustCompile(dnfDownloadPattern)
+	fullArgs := []string{}
```
liulanze: Was `args` modified? It seems pass `args` directly to `NewExecBuilder` and drop `fullArgs` still work?

cwize1: Old code. Fixed

vinceaperri (dismissed):

vinceaperri (dismissed):

liulanze (approved):

PR #850, opened 2026-07-17
Remove LiveOS dependency version checks for AZL.

cwize1:
When support for LiveISO was initially being added to Image Customizer, we experienced a bunch of bugs in the Azure Linux system packages, particularly around PXE and SELinux. Once the fixes went in, we added package version checks to IC to warn the user when they are using unsupported versions of these packages. Unfortunately, these checks aren't being done in a distro-independent way.

A particularly annoying aspect is that the code parses the RPM EVR ( epoch-version-release) field, including breaking apart the release field, and then stores the component pieces in the `saved-configs.yaml` file in the ISO. This is done so that ISO->ISO customization can also do the version checks. But it means that to rewrite this logic to be distro-independent, it would require changing the format of the `saved-configs.yaml` file to be distro-independent.

So, instead this change just removes all the package version checks.

All the relevant bug fixes are 1-2 years old. So, rewriting these checks probably isn't worth it at this point.

---

### **Checklist**

- [x] Tests added/updated


toolkit/tools/pkg/imagecustomizerlib/distrohandler.go
```diff
@@ -54,9 +54,6 @@ type DistroHandler interface {
 	// toolsChroot) must return an error rather than a misleading false.
 	IsPackageInstalled(imageChroot safechroot.ChrootInterface, toolsChroot *safechroot.Chroot, packageName string) (bool, error)
 
-	// GetPackageInformation queries the installed-package database for packageName and returns its parsed information.
-	GetPackageInformation(imageChroot *safechroot.Chroot, toolsChroot *safechroot.Chroot, packageName string) (*PackageVersionInformation, error)
-
 	// Get all installed packages from the chroot.
 	// toolsChroot has the same semantics as in IsPackageInstalled.
 	GetAllPackagesFromChroot(imageChroot safechroot.ChrootInterface, toolsChroot *safechroot.Chroot) ([]cosiapi.OsPackage, error)
```
vinceaperri, line 65: Per-distro implementations of `GetPackageInformation` and the underlying `PackageVersionInformation` machinery are now dead code and can be removed

vinceaperri (changes requested):

toolkit/tools/pkg/imagecustomizerlib/distrohandler.go
cwize1: Thanks! Fixed

toolkit/tools/pkg/imagecustomizerlib/savedconfigs.go
```diff
@@ -84,9 +84,7 @@ func (p *PxeSavedConfigs) IsValid() error {
 }
 
 type OSSavedConfigs struct {
-	DracutPackageInfo        *PackageVersionInformation     `yaml:"dracutPackage"`
-	RequestedSELinuxMode     imagecustomizerapi.SELinuxMode `yaml:"selinuxRequestedMode"`
-	SELinuxPolicyPackageInfo *PackageVersionInformation     `yaml:"selinuxPolicyPackage"`
+	RequestedSELinuxMode imagecustomizerapi.SELinuxMode `yaml:"selinuxRequestedMode"`
```
vinceaperri, line 87: We need to keep / deprecate these fields instead of remove them to maintain backwards-compatibility. Otherwise, ISO->ISO recustomization of any previously-built ISO will fail with "field not found". This is because `UnmarshalYaml` decodes with `KnownFields(true)`.

toolkit/tools/pkg/imagecustomizerlib/savedconfigs.go
```diff
@@ -210,18 +206,11 @@ func updateSavedConfigs(savedConfigsFilePath string,
 		// newOSDracutVersion can be nil if the input is an ISO and the
```
vinceaperri: Remove this comment about `newOSDracutVersion` -- no longer exists.

vinceaperri (changes requested):

toolkit/tools/pkg/imagecustomizerlib/savedconfigs.go
cwize1: Fixed

toolkit/tools/pkg/imagecustomizerlib/savedconfigs.go
cwize1: Thanks! Fixed

vinceaperri (dismissed):
Build break

cwize1:
> Build break

Fixed

toolkit/tools/pkg/imagecustomizerlib/savedconfigs.go
```diff
@@ -84,9 +84,11 @@ func (p *PxeSavedConfigs) IsValid() error {
 }
 
 type OSSavedConfigs struct {
-	DracutPackageInfo        *PackageVersionInformation     `yaml:"dracutPackage"`
-	RequestedSELinuxMode     imagecustomizerapi.SELinuxMode `yaml:"selinuxRequestedMode"`
-	SELinuxPolicyPackageInfo *PackageVersionInformation     `yaml:"selinuxPolicyPackage"`
+	RequestedSELinuxMode imagecustomizerapi.SELinuxMode `yaml:"selinuxRequestedMode"`
```
vinceaperri, line 87: `RequestedSELinuxMode` no longer has a reader. Could be deprecated too. So the whole `os` section could be deprecated. Also, can we add a test that loads a `saved-configs.yaml` still containing all these deprecated values? Would ensure no regressions here.

toolkit/tools/pkg/imagecustomizerlib/liveosisoartifactstore.go
```diff
@@ -409,23 +405,11 @@ func createIsoFilesStoreFromMountedImage(inputArtifactsStore *IsoArtifactsStore,
 }
 
 func createIsoInfoStoreFromMountedImage(buildDir string, imageRootDir string, distroHandler DistroHandler,
-	toolsChroot *safechroot.Chroot,
 ) (infoStore *IsoInfoStore, err error) {
 	infoStore = &IsoInfoStore{}
 
-	var bindMount *safemount.Mount
 	chrootDir := imageRootDir
```
vinceaperri, line 411: Remove `chrootDir` var and just use imageRootDir throughout?

toolkit/tools/pkg/imagecustomizerlib/liveosisoartifactstore.go
```diff
@@ -442,37 +426,11 @@ func createIsoInfoStoreFromMountedImage(buildDir string, imageRootDir string, di
 	}
 	infoStore.seLinuxMode = imageSELinuxMode
```
vinceaperri, line 427: `seLinuxMode` is now only read by `dumpInfoStore`. Worth keeping around?

toolkit/tools/pkg/imagecustomizerlib/liveosisoartifactstore.go
cwize1: Fixed

toolkit/tools/pkg/imagecustomizerlib/liveosisoartifactstore.go
cwize1: Fixed

toolkit/tools/pkg/imagecustomizerlib/savedconfigs.go
cwize1: Deprecated the `os` section.

cwize1: I don't know if it is worth it to add regression tests for the deprecated fields. I doubt anyone is doing ISO->ISO customization with old images they have cached.

toolkit/tools/pkg/imagecustomizerlib/savedconfigs.go
```diff
@@ -110,11 +102,6 @@ func (c *SavedConfigs) IsValid() (err error) {
 		return fmt.Errorf("%w:\n%w", ErrIsoConfigInvalidPxeField, err)
 	}
 
-	err = c.OS.IsValid()
-	if err != nil {
-		return fmt.Errorf("%w:\n%w", ErrIsoConfigInvalidOsField, err)
-	}
-
 	return nil
```
vinceaperri, line 105: `ErrIsoConfigInvalidOsField` is now unused and can go too.

toolkit/tools/pkg/imagecustomizerlib/liveosisogrub.go
```diff
@@ -217,14 +217,6 @@ func updateGrubCfgForLiveOS(inputContentString string, initramfsImageType imagec
...
 
-	if disableSELinux {
-		inputContentString, err = updateSELinuxCommandLineHelperAll(inputContentString,
-			imagecustomizerapi.SELinuxModeDisabled)
-		if err != nil {
-			return "", fmt.Errorf("failed to set SELinux mode:\n%w", err)
-		}
-	}
-
 	savedArgs := GrubArgsToString(savedConfigs.LiveOS.KernelCommandLine.ExtraCommandLine)
```
vinceaperri, line 220: `updateSELinuxCommandLineHelperAll` loses its last caller here and can be removed too.

vinceaperri (changes requested):

toolkit/tools/pkg/imagecustomizerlib/savedconfigs.go
cwize1: Fixed

toolkit/tools/pkg/imagecustomizerlib/liveosisogrub.go
cwize1: Fixed

PR #851, opened 2026-07-18
Enable cross-arch ACL COSI convert

bfjelds:
When converting an image using a cross-arch container, `btrfs subvolume list` fails because it uses ioctls that are not supported. 

per copilot:
```
Most btrfs ioctls are translated — but the one  btrfs subvolume list  relies on is not.

Source:  linux-user/ioctls.h  (verified on  master ):
• Present:  BTRFS_IOC_SNAP_CREATE ,  SUBVOL_CREATE ,  INO_LOOKUP ,  GET_SUBVOL_INFO ,  GET_SUBVOL_ROOTREF , etc. (91 BTRFS references)
• Absent:  BTRFS_IOC_TREE_SEARCH  and  BTRFS_IOC_TREE_SEARCH_V2  — grep returns zero matches.

 btrfs subvolume list  issues  BTRFS_IOC_TREE_SEARCH_V2  (falling back to  TREE_SEARCH ). Neither is in the table → both hit the  target_cmd == 0  sentinel →  -TARGET_ENOTTY . That is precisely your  ERROR: can't perform the search: Inappropriate ioctl for device.
```

It would be easier for AgentBaker's conversion if cross-arch conversion worked.  And it seems that for ACL, the btrfs call is not really needed because the fstab is on a known path that doesn't require subvolumes (for example, the Trident mkcosi command did not use the btrfs call).

This change detects ACL (when fstab path is share/ic/etc/fstab) and avoids btrfs subvolume list call.  Non-ACL images are unaffected by change.

Validation:
* AzlImageTools-OneBranch-BuildImageCustomizerDev: https://dev.azure.com/mariner-org/polar/_build/results?buildId=1162525&view=results
* Tested ACL manually by downloading img from acldevel and IC container from above pipeline and running convert on amd64 ubuntu machine.
* Tested for regressions with Trident IC pipeline: https://dev.azure.com/mariner-org/ECF/_build/results?buildId=1162526&view=results

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
@@ -171,12 +171,12 @@ func findRootfsPartition(diskPartitions []diskutils.PartitionInfo, buildDir stri
 	// ioctl, which qemu-user does not translate, so `btrfs subvolume list`
 	// fails with ENOTTY when converting an ARM64 ACL image on an AMD64 host.
 	// The marker fully identifies the rootfs, so enumeration is unnecessary.
-	icPartition, err := findIcFstabPartition(diskPartitions, tmpDir)
+	aclPartition, err := findAclFstabPartition(diskPartitions, tmpDir)
 	if err != nil {
 		return nil, "", err
 	}
-	if icPartition != nil {
-		return icPartition, "share/ic", nil
+	if aclPartition != nil {
+		return aclPartition, "share/ic", nil
 	}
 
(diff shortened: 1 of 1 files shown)


cwize1:
The correct solution is to make IC cross-arch aware. That way, it can use the correct arch version of the `btrfs` command, which will avoid this problem.

FYI: This problem is caused by the fact that QEMU has to translate all syscalls, in case there are any differences between CPU archs (e.g. type size, type alignment, etc.) So, only a limited number of syscalls are supported.

PR #853, opened 2026-07-21
Fix kernel command-line options being carried over through hard-reset

vinceaperri:
When re-customizing a UKI image with `os.bootloader.resetType: hard-reset` and `os.uki.mode: create`, the kernel command line was not reset to the stock image. Arguments added by a previous customization (such as `rd.info`) leaked into the regenerated UKIs, and newly requested `os.kernelCommandLine.extraCommandLine` arguments (such as `rd.debug`) were dropped. This is a regression from #844.

## Fix

- **customizeos.go** — on a hard-reset, skip pre-seeding `uki-kernel-info.json` with the base image's current UKI command line. Pre-seeding carried the old command line through the reset.
- **customizeuki.go** — `extractKernelToArgs` now reads the kernel command line only from the regenerated grub boot config. It no longer falls back to the stale command line embedded in the existing UKIs.
- **grubcfgutils.go / distrohandler_azurelinux4.go / distrohandler_fedora.go** — for BLS-based distros (Azure Linux 4, Fedora), read the command line from `loader/entries/*.conf` and fall back to the `set kernelopts="..."` default that `grub2-mkconfig` writes into `grub.cfg`. That is where a hard-reset lands the regenerated command line on those distros.

## Behavior

Re-customizing a UKI image that previously added `rd.info`, now with hard-reset plus `extraCommandLine: [rd.debug]`:

| Kernel command-line arg | Before | After |
| --- | --- | --- |
| `rd.info` (prior customization) | kept (leaked) | dropped (reset to stock) |
| `rd.debug` (new extraCommandLine) | dropped | added |

## Tests

Added `TestCustomizeImageUkiCreateNoBootloaderReset` and reworked the verity UKI and grub tests to assert the reset command line — each expected arg present in every kernel, each dropped arg absent — across Azure Linux 3, Azure Linux 4, and the non-hard-reset create path.

---

### **Checklist**
- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/grubcfgutils.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/grubcfgutils.go
@@ -5,7 +5,10 @@ package imagecustomizerlib
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"io/fs"
+	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
@@ -29,6 +32,9 @@ var (
 
 	// Captures the variable a grub 'search' command assigns via '--set=<var>' or '--set <var>' (e.g. root or boot).
 	grubSearchSetTargetRegex = regexp.MustCompile(`\s+--set(?:=|\s+)(\w+)`)
+
+	// Captures the value of the `set kernelopts="..."` default that grub2-mkconfig writes into grub.cfg.
+	grubKerneloptsRegex = regexp.MustCompile(`(?m)^[ \t]*set kernelopts="([^"]*)"`)
 )
 
 const (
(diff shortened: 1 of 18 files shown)


toolkit/tools/pkg/imagecustomizerlib/grubcfgutils.go
```diff
@@ -29,6 +32,9 @@ var (
 
 	// Captures the variable a grub 'search' command assigns via '--set=<var>' or '--set <var>' (e.g. root or boot).
 	grubSearchSetTargetRegex = regexp.MustCompile(`\s+--set(?:=|\s+)(\w+)`)
+
+	// Captures the value of the `set kernelopts="..."` default that grub2-mkconfig writes into grub.cfg.
+	grubKerneloptsRegex = regexp.MustCompile(`(?m)^[ \t]*set kernelopts="([^"]*)"`)
```
cwize1: Parse the `grub.cfg` file using `grub.TokenizeConfig`, to properly handle string escapes.

cwize1 (changes requested):

cwize1:
Thank you for being so thorough on the test cases!

toolkit/tools/pkg/imagecustomizerlib/grubcfgutils.go
vinceaperri: Done.

toolkit/tools/pkg/imagecustomizerlib/grubcfgutils.go
```diff
@@ -627,6 +630,89 @@ func GrubArgsToString(args []string) string {
...
+			return "", false, err
+		}
+
+		for _, arg := range args {
+			if arg.Name == grubKernelOpts {
+				if arg.ValueHasVarExpansion {
+					// ParseCommandLineArgs clears values it cannot resolve statically, so a kernelopts default
+					// that references a variable comes back empty. Warn loudly rather than silently returning an
+					// empty kernel command line.
+					logger.Log.Warnf(
```
cwize1: I reserve WARN logs for stuff that the user has to go fix. For example, IC failed to cleanup mounts, etc. If something might be a problem but might not, then use the INFO level.

I've seen projects where lots of expected WARN logs created red herrings while debugging problems, which can be pretty frustrating for a user.

cwize1: Why not just error out in this case? Do we know of any scenario where we expect this to occur?

IC is a build tool, not an online service. So, I generally prefer to error out in unexpected situations instead of trying to be robust.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/grubcfgutils.go
vinceaperri: Happy to change this to an error. I would prefer to error here, but the existing code (`ParseCommandLineArgs`) does not treat it as an error, just returns an empty string, which could lead to kernel opts being dropped without the user even knowing it. This led me to compromise with adding a warning, so that a new error condition isn't created, but the user is still aware that something wrong happened.

To confirm, your suggestion is to treat this as an error instead of a warning?

vinceaperri: Removed.

vinceaperri: Switched to an error.

cwize1: `ParseCommandLineArgs` was originally written so that it could be used in situations where the args that were being searched for typically did not contain variable expansion, even if some of the args did contain variable expansion. Hence, `ParseCommandLineArgs` leaves it up the caller if `ValueHasVarExpansion=true` is problematic or not.

cwize1 (approved):

PR #858, opened 2026-07-24
Fix UKI recustomize kernel swap flow and add coverage

vinceaperri:
Re-customizing a UKI image with `os.uki.mode: create` broke whenever the run installed a new kernel. On Azure Linux 3 the build failed outright with `no command line arguments found for kernel (...)`. On Azure Linux 4 it silently gave the new kernel a stale command line, so that kernel booted without the options every other kernel in the image received.

## Problem

`prepareUkiHelper` has to resolve a kernel command line for every kernel it finds under `/boot`. It looked in two places, and a kernel installed during the current run is in neither.

| Source | Why the new kernel is missing from it |
| --- | --- |
| `uki-kernel-info.json` | Only covers the kernels the base image's existing UKIs were built for. |
| Grub boot config | Converting an image to UKI clears `/boot`, which deletes `grub.cfg` and the BLS entries. |

The two distros then diverge, because of what regenerates that boot config afterwards.

- **Azure Linux 3** removes grub2, so there is no `grub2-mkconfig` to rewrite `grub.cfg`. The lookup came up empty and the build aborted.
- **Azure Linux 4** keeps `kernel-install`, which ships with systemd. Installing a kernel package runs it and writes fresh `/boot/loader/entries/*.conf` files. That happens before `os.kernelCommandLine` changes are applied, so those entries hold a pre-customization command line and the new kernel silently inherited it.

## Fix

1. **Read `uki-kernel-info.json` first, and only consult the grub boot config when it doesn't exist or is empty.** Previously the grub boot config was read unconditionally and the saved UKI command lines merely won per kernel, so a kernel absent from the saved info picked up the stale regenerated entry instead. The two cases that genuinely need the boot config keep working, namely (1) a grub image being converted to UKI for the first time and (2) a hard bootloader reset, since both of those skip the `uki-kernel-info.json` save.
2. **Let a kernel with no command line of its own inherit the one the image's other kernels share.** The new `getFallbackKernelArgs` collects the distinct non-empty command lines and requires exactly one to remain. It returns an error rather than guessing when there is nothing to inherit, or when the kernels disagree. Standard use of Image Customizer should never encounter these errors because kernel command line changes are always applied to every kernel in the image, so they normally agree.

Debug logging now reports which source each kernel's command line came from.

## Tests

`TestGetFallbackKernelArgs` covers the fallback selection directly.

`TestCustomizeImageVerityUsrUkiRecustomizeMultiKernelThenKernelSwap` runs three passes.

| Pass | What it does |
| --- | --- |
| 0 | Converts the grub base image to a single-kernel usr-verity UKI image. Adds `rd.info`. |
| 1 | Re-customizes without a bootloader reset and installs extra kernels. Adds `systemd.log_level=debug`. |
| 2 | Re-customizes again and swaps `kernel` for `kernel-npi`. Adds `console=ttyAMA0,115200n8`. |

Azure Linux 4 stops after pass 1, because it currently has no alternative kernel package to swap to, only other versions of `kernel`. Before this fix, pass 1 failed on both distros. Azure Linux 3 aborted the build, and on Azure Linux 4 the build succeeded but the new kernels came out with a stale command line. Both now pass.

## Notes

- The multi-kernel package lists moved into `testdata/lists/` so both multi-kernel tests share them
- Azure Linux 4's list gained a second kernel, matching Azure Linux 2 and 3.
- `verity-reinit-usr-uki-cmdline.yaml` was renamed to `verity-reinit-usr-uki-noreset.yaml` to match what it actually does.
- The `usr` partition in `verity-usr-uki-base.yaml` grew from 1G to 2G to fit the extra kernels for Azure Linux 4.

### Checklist
- [x] Tests added/updated


toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
```diff
@@ -397,6 +418,42 @@ func prepareUkiHelper(ctx context.Context, buildDir string, uki *imagecustomizer
 	return nil
 }
 
+// getFallbackKernelArgs returns the non-empty kernel command line arguments that can serve as a fallback for a kernel
+// that has no command line of its own.
+func getFallbackKernelArgs(existingUkiCmdlines map[string]string, grubKernelToArgs map[string]string) (string, error) {
+	seen := make(map[string]struct{})
```
cwize1: You should add comments describing what this logic is, since it isn't obvious at first glance.

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
```diff
@@ -397,6 +418,42 @@ func prepareUkiHelper(ctx context.Context, buildDir string, uki *imagecustomizer
...
+		for _, cmdline := range grubKernelToArgs {
+			if cmdline != "" {
+				seen[cmdline] = struct{}{}
+			}
+		}
+	}
+
+	args := make([]string, 0, len(seen))
+	for arg := range seen {
+		args = append(args, arg)
```
cwize1: This seems weird. Why are we appending the args from all the different kernel versions?

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
vinceaperri: Done.

toolkit/tools/pkg/imagecustomizerlib/customizeuki.go
vinceaperri: Updated.

toolkit/tools/pkg/imagecustomizerlib/customizeuki_test.go
```diff
@@ -254,6 +254,105 @@ func testCustomizeImageVerityUsrUkiRecustomizeHelper(t *testing.T, baseImageInfo
 		addonFilesChecksums, newAddonFilesChecksums)
 }
 
+func TestCustomizeImageVerityUsrUkiRecustomizeMultiKernelThenKernelSwap(t *testing.T) {
+	for _, baseImageInfo := range baseImageAzureLinux3Plus {
+		if baseImageInfo.Version == baseImageVersionAzl4 {
+			t.Skip("Azure Linux 4.0 doesn't have an additional kernel to test with")
```
cwize1: `Skip()` raises a panic and therefore will break the for-loop. The `Skip()` call should in in the `t.Run` functor.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/customizeuki_test.go
vinceaperri: Fixed.

cwize1 (dismissed):

toolkit/tools/pkg/imagecustomizerlib/customizeos.go
```diff
@@ -220,7 +221,7 @@ func doOsCustomizations(ctx context.Context, rc *ResolvedConfig, imageConnection
 		}
 	}
 
-	err = prepareUki(ctx, rc.BuildDirAbs, rc.Uki, imageChroot, toolsChroot, distroHandler)
+	err = prepareUki(ctx, rc.BuildDirAbs, rc.Uki, hasUkis, imageChroot, toolsChroot, distroHandler)
```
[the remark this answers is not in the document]
vinceaperri: Fixed.

vinceaperri:
> ## Pull request overview
> Copilot reviewed 14 out of 15 changed files in this pull request and generated no new comments.
> 
> Comments suppressed due to low confidence (1)
> **toolkit/tools/pkg/imagecustomizerlib/customizeuki_test.go:313**
> 
> * `verifyUsrVerityUkiRecustomized` should validate the _re-customized output image_ (pass 1), but this call passes `outImageFilePath0` (the pass-0 image). That can mask regressions in the re-customization path because `/boot` layout checks are run against the wrong image.
> 
> ```
> 	verifyUsrVerityUkiRecustomized(t, buildDir, outImageFilePath0, ukiFilesChecksums0, ukiFilesChecksums1,
> 		addonFilesChecksums0, addonFilesChecksums1)
> ```

Fixed.

cwize1 (approved):

PR #859, opened 2026-07-28
test(cosi): validate emitted metadata against Trident COSI schema

frhuelsz:
## What

Adds a unit test that validates the COSI `metadata.json` Image Customizer emits against the JSON Schema authored in the Trident repo (the downstream consumer of COSI images).

Image Customizer emits COSI metadata that Trident consumes. Today, when IC changes the metadata shape, Trident is the first to notice, often only after picking up a new IC build in preview/release. This test moves that detection **upstream into IC's own CI**, so schema-breaking drift fails in the IC PR that introduces it.

## How

- New `toolkit/tools/cosiapi/cosimetadata_schema_test.go`:
  - Fetches the schema **live** from the Trident source of truth (`raw.githubusercontent.com/microsoft/trident/main/docs/Reference/Composable-OS-Image/cosi-metadata-v1.2.schema.json`), so validation always runs against the current contract rather than a snapshot that can go stale.
  - Builds representative COSI metadata from the `cosiapi` types (GRUB + dm-verity, and systemd-boot + UKI) and validates each against the schema.
  - Includes a version tripwire: if Trident bumps the COSI revision, the `Version` constant IC emits (`cosicommon.go`) must be updated to match.
- Adds `github.com/santhosh-tekuri/jsonschema/v6` (JSON Schema draft 2020-12 validator).
- Runs `./cosiapi` in the existing hermetic unit-test job (`binary-build.yml`); the package was not covered by CI before.

## Dependency / why draft

The test fetches the schema from Trident `main`, so it requires the schema to be present there. It is currently in **microsoft/trident#714** (`docs(cosi): add JSON Schema for COSI metadata v1.2`), not yet merged. **This PR stays draft until #714 lands on Trident main**; until then the `./cosiapi` job will fail with a 404.

Requires network access to `raw.githubusercontent.com` during `go test`.

### DIFF
--- a/toolkit/tools/cosiapi/cosimetadata_schema_test.go
+++ b/toolkit/tools/cosiapi/cosimetadata_schema_test.go
@@ -1,0 +1,256 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// This test guards the COSI metadata that Image Customizer emits against the
+// JSON Schema published by Trident (the downstream consumer). If a change to
+// the cosiapi types drifts the emitted metadata.json away from the contract
+// Trident expects, this test fails in IC's own CI, so the break is caught here
+// rather than downstream when Trident picks up a new IC build.
+//
+// The schema is fetched live from the Trident repository at test time so the
+// test always validates against the current contract (rather than a vendored
+// snapshot that can silently go stale). This means the test requires network
+// access to raw.githubusercontent.com.
+
+package cosiapi
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/santhosh-tekuri/jsonschema/v6"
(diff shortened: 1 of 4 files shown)


toolkit/tools/cosiapi/cosimetadata_schema_test.go
```diff
@@ -0,0 +1,256 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// This test guards the COSI metadata that Image Customizer emits against the
+// JSON Schema published by Trident (the downstream consumer). If a change to
+// the cosiapi types drifts the emitted metadata.json away from the contract
+// Trident expects, this test fails in IC's own CI, so the break is caught here
+// rather than downstream when Trident picks up a new IC build.
+//
+// The schema is fetched live from the Trident repository at test time so the
```
cwize1, line 10: This is in conflict with the ideal of reproducible builds.

PR #866, opened 2026-08-06
Fix ACL mount dirs dir overlay.

cwize1:
The ACL image doesn't create placeholder directories for special directories (e.g. `/dev`). In PR #771, a workaround was added when mounting the image as read-only that creates an overlay that holds placeholder directories for the special directories. However, this change accidentally messed up the overlay settings so that the host root directory was used instead of the customization OS's root directory.

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
@@ -679,7 +679,7 @@ func createMountsDirOverlay(buildDir string, includeDefaultMounts bool, layout [
 
 	// Overlay the ACL root on top of the special directories directory.
 	mountPoint := safechroot.NewPreDefaultsMountPoint("overlay", "/", "overlay", unix.MS_RDONLY,
-		fmt.Sprintf("lowerdir=%s:%s", mountDirsDir, rootMountPoint))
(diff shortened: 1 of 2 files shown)


liulanze (dismissed):

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
```diff
@@ -676,7 +679,7 @@ func createMountsDirOverlay(buildDir string, includeDefaultMounts bool, layout [
 
 	// Overlay the ACL root on top of the special directories directory.
 	mountPoint := safechroot.NewPreDefaultsMountPoint("overlay", "/", "overlay", unix.MS_RDONLY,
-		fmt.Sprintf("lowerdir=%s:/", mountDirsDir))
+		fmt.Sprintf("lowerdir=%s:%s", mountDirsDir, rootMountPoint))
```
vinceaperri: nit: `lowerdir` is ordered top to bottom, so the comment above has it backwards.

vinceaperri (dismissed):

toolkit/tools/pkg/imagecustomizerlib/partitionutils.go
cwize1: The comment represented the intended behavior. Fixed.

liulanze (approved):

PR #874, opened 2026-08-12
Handle ACL's /etc overlay during customization: assemble it for the chroot, fold it into the baseline

liulanze:
At runtime, ACL composes `/etc` from an overlay: the read-only baseline under
`/usr` (dm-verity protected) below the writable physical `/etc` on the root
partition. The bare rootfs that Image Customizer mounts holds only the physical
`/etc`, so in-chroot code sees an almost-empty `/etc`: package scriptlets
rebuild files like `/etc/passwd` from scratch (dropping users the baseline
defines — the root cause of the root-shell and rpc/core user resets we have
been compensating for in the ACL pipeline), and configuration staged in the
baseline is invisible to anything that reads `/etc`.

This change makes Image Customizer detect the overlay layout and handle it
automatically:

1. **Assemble** the same overlay the ACL initrd creates at runtime

2. **Finalize** it after customization:

### **Checklist**
- [x] Tests added/updated
- [x] Documentation updated (if needed) — not needed: no new configuration
  API or user-facing action; behavior is automatic for ACL images and this
  description documents it

### DIFF
--- a/toolkit/tools/pkg/imagecustomizerlib/acletcoverlay.go
+++ b/toolkit/tools/pkg/imagecustomizerlib/acletcoverlay.go
@@ -5,11 +5,11 @@ package imagecustomizerlib
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
 
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/imagecustomizerapi"
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/file"
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/logger"
 	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/safemount"
@@ -20,8 +20,8 @@ import (
 )
 
 var (
-	ErrEtcOverlayAssemble = NewImageCustomizerError("EtcOverlay:Assemble", "failed to assemble /etc overlay")
-	ErrEtcOverlayFinalize = NewImageCustomizerError("EtcOverlay:Finalize", "failed to finalize /etc overlay")
+	ErrAclEtcOverlayAssemble = NewImageCustomizerError("AclEtcOverlay:Assemble", "failed to assemble /etc overlay")
+	ErrAclEtcOverlayFinalize = NewImageCustomizerError("AclEtcOverlay:Finalize", "failed to finalize /etc overlay")
 )
 
 const (
@@ -31,22 +31,23 @@ const (
 	aclEtcBaselineDir = "usr/share/distro/etc"
 
 	// Scratch directory holding the overlay's work directory for the duration of customization.
-	etcOverlayWorkDirName = ".ic-etc-work"
+	// Matches the work directory the ACL initrd uses (and recreates) for the runtime mount.
+	aclEtcOverlayWorkDirName = ".etc-work"
 
 	// Temporary directory holding the merged /etc snapshot while folding.
-	etcOverlayFoldTmpName = ".ic-etc-fold"
+	aclEtcOverlayFoldTmpName = ".ic-etc-fold"
 )
 
-// etcOverlayFoldExclusions lists paths that are relative to /etc that must never be folded into the
+// aclEtcOverlayFoldExclusions lists paths that are relative to /etc that must never be folded into the
 // baseline. machine-id is per-machine identity: the ACL image build removes it before populating
 // the baseline so that systemd sees a missing file on first boot and runs its first-boot logic.
-var etcOverlayFoldExclusions = []string{
+var aclEtcOverlayFoldExclusions = []string{
 	"machine-id",
 }
 
-// EtcOverlay is a distro's /etc overlay, assembled over an image's /etc for the duration of OS
+// AclEtcOverlay is a distro's /etc overlay, assembled over an image's /etc for the duration of OS
 // customization. See DistroHandler.SetupEtcOverlay.
-type EtcOverlay struct {
+type AclEtcOverlay struct {
 	// Host path of the image's /etc: the overlay's mountpoint and its writable upper layer.
 	etcDir string
 	// Host path of the /etc baseline, the overlay's read-only lower layer.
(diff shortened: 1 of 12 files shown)


toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -727,6 +727,14 @@ func customizeImageHelper(ctx context.Context, rc *ResolvedConfig, partitionsCus
 		return nil, nil, nil, "", err
 	}
 
+	// Assemble the distro's /etc overlay if the distro has one, so that customization operates on
+	// the same merged /etc the booted OS sees.
+	etcOverlay, err := distroHandler.SetupEtcOverlay(ctx, imageConnection.Chroot())
```
cwize1: Any reason this is done here instead of in `doOsCustomizations`?

toolkit/tools/pkg/imagecustomizerlib/acletcoverlay.go
```diff
@@ -0,0 +1,287 @@
+// Copyright (c) Microsoft Corporation.
```
cwize1, line 1: Rename to `acletcoverlay.go`.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
```diff
@@ -757,6 +765,14 @@ func customizeImageHelper(ctx context.Context, rc *ResolvedConfig, partitionsCus
 		return nil, nil, nil, "", err
 	}
 
+	// Finalize the /etc overlay, this is a no-op for distros without one. When the distro's /etc
+	// baseline is writable, this folds the customized /etc into it. Must happen while the image is
+	// still mounted and before the filesystems are trimmed and verity is recalculated.
+	err = etcOverlay.Finalize(ctx)
```
cwize1: I think this should run before the `finalizeCustomization` scripts. This would allow users to mess with the contents of `/usr/share/distro/etc` if they need to.

toolkit/tools/pkg/imagecustomizerlib/etcoverlay.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+	if !baselineExists {
+		return nil, fmt.Errorf("image has no /etc baseline directory (%s)", aclEtcBaselineDir)
+	}
+
+	// The baseline lives on the /usr filesystem, which is only mounted read-write when the image is
+	// customized with reinitializeVerity: all. When it is writable, finalizing the overlay folds the
+	// merged /etc into the baseline, while when it is read-only, the /etc changes are left on the
+	// root partition, just like changes made on a booted system.
+	foldable := false
```
cwize1: Pass the `reinitializeVerity` setting into this function instead of querying the read-write status of the files.

toolkit/tools/pkg/imagecustomizerlib/etcoverlay.go
```diff
@@ -0,0 +1,287 @@
...
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/logger"
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/safemount"
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/shell"
+	"github.com/sirupsen/logrus"
+	"go.opentelemetry.io/otel"
+	"golang.org/x/sys/unix"
+)
+
+var (
+	ErrEtcOverlayAssemble = NewImageCustomizerError("EtcOverlay:Assemble", "failed to assemble /etc overlay")
```
cwize1: Please make sure all the names in this file have `Acl` in them, since this is specific to `Acl`. (Or you can use `Flatcar` if you wish.)

toolkit/tools/pkg/imagecustomizerlib/etcoverlay.go
```diff
@@ -0,0 +1,287 @@
...
+)
+
+const (
+	// The /etc baseline directory, relative to the image root. At runtime, ACL composes /etc as an
+	// overlay whose read-only lower layer is this directory and whose writable upper layer is the
+	// physical /etc on the root partition.
+	aclEtcBaselineDir = "usr/share/distro/etc"
+
+	// Scratch directory holding the overlay's work directory for the duration of customization.
+	etcOverlayWorkDirName = ".ic-etc-work"
```
cwize1: I would use `/.etc-work` to match what is used during runtime.

toolkit/tools/pkg/imagecustomizerlib/etcoverlay.go
```diff
@@ -0,0 +1,287 @@
...
+
+	// Swap the snapshot in as the new baseline. Both paths are on the same filesystem, so the
+	// window without a baseline is two renames wide.
+	oldBaselineDir := o.baselineDir + ".old"
+	err = os.RemoveAll(oldBaselineDir)
+	if err != nil {
+		return fmt.Errorf("failed to remove stale directory (%s):\n%w", oldBaselineDir, err)
+	}
+
+	err = os.Rename(o.baselineDir, oldBaselineDir)
```
cwize1: Why bother saving the old baseline directory, when you are going to delete it soon after?

toolkit/tools/pkg/imagecustomizerlib/etcoverlay.go
```diff
@@ -0,0 +1,287 @@
...
+
+func (o *EtcOverlay) foldTmpDir() string {
+	return filepath.Join(filepath.Dir(o.baselineDir), etcOverlayFoldTmpName)
+}
+
+// Finalize unmounts the overlay. When the baseline is writable, the merged /etc is first folded
+// into the baseline and the physical /etc is left empty; otherwise the accumulated /etc changes
+// remain on the root partition, matching what the same changes would produce on a booted system.
+func (o *EtcOverlay) Finalize(ctx context.Context) error {
+	if o == nil {
```
cwize1: `Finalize` runs in the good path. So, it should be error free. So, this nil check should be removed, since something is wrong if `o` is `nil` here.

toolkit/tools/pkg/imagecustomizerlib/etcoverlay.go
```diff
@@ -0,0 +1,287 @@
...
+		return fmt.Errorf("failed to remove overlay work directory (%s):\n%w", o.workDir, err)
+	}
+
+	return nil
+}
+
+// Close unmounts the overlay without folding. Intended as deferred error-path cleanup, the image
+// is expected to be discarded. No-op after a successful Finalize and on a nil receiver.
+func (o *EtcOverlay) Close() {
+	if o == nil || o.done {
```
cwize1: `o` will never be `nil` here.

cwize1: Remove `o.done`. Check the state of each sub-resource individually, if they don't do it themselves.

toolkit/tools/pkg/imagecustomizerlib/etcoverlay.go
```diff
@@ -0,0 +1,287 @@
...
+		return
+	}
+
+	o.mount.Close()
+
+	for _, dir := range []string{o.workDir, o.foldTmpDir()} {
```
cwize1: I don't think we need to remove these directories here. The customization process occurs on a copy of the base image. If the customization process fails, then the image is not output. So, it doesn't matter what the contents of the base image is.

toolkit/tools/pkg/imagecustomizerlib/etcoverlay.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+	if err != nil {
+		return fmt.Errorf("%w:\n%w", ErrEtcOverlayFinalize, err)
+	}
+
+	o.done = true
+	return nil
+}
+
+func (o *EtcOverlay) unmountHelper() error {
```
cwize1: Name the functions `unmount` and `foldAndUnmount`. In general, you don't need to add the `Helper` suffix to function names.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/etcoverlay.go
liulanze: Yeah, make sense, fixed.

toolkit/tools/pkg/imagecustomizerlib/imagecustomizer.go
liulanze: Hmm, yeah that is a better place, revised.

cwize1 (approved):

PR #883, opened 2026-08-28
Add symbolic link support to additionalDirs

suresh-thelkar:
## Update (2026-09-17): Reworked per Design Review — Please Re-review

Thanks @vinceaperri and @liulanze for the earlier feedback. Following additional design discussions with @chrisgun/@vinceaperri, this PR has been **reworked** since the initial review. The revised design addresses the feedback received and changes the behavior from *always preserving symlinks* to an explicit, opt-in model.

### Design Change: Opt-in `symlinkMode` (Previously: Always Preserve)

The original implementation preserved **all symlinks unconditionally**. Since `additionalDirs` is a stable API and existing users may depend on the current dereferencing behavior, symlink preservation is now **opt-in** through a new `symlinkMode` enum, gated behind a preview feature.

```yaml
previewFeatures:
  - preserve-symlinks

os:
  additionalDirs:
    - source: ./linkdir
      destination: /opt/linktest
      symlinkMode: preserve # dereference (default) | preserve
```

#### Behavior

- `dereference` (default, can be omitted)
  - Preserves the existing behavior.
  - Fully backward compatible.
  - Does not require any preview feature.

- `preserve`
  - Recreates symbolic links verbatim instead of dereferencing them.
  - Requires the `preserve-symlinks` preview feature to be enabled.

#### Why an Enum Instead of a Boolean?

- Provides room for future extensibility without introducing additional schema changes.
- Makes the intended behavior explicit and self-documenting.
- Naming aligns with familiar file-copy semantics (`cp --dereference`).

### Additional Notes

- Existing configurations continue to behave exactly as before.
- Users must explicitly opt in to symlink preservation.
- The preview feature gate allows us to gather feedback before making any future decisions regarding broader availability.
- The implementation, tests, and documentation have been updated to reflect the revised design.

Please take another look at the updated design and implementation when you get a chance. Thanks again for the feedback and review.

---
# Earlier Description
# Add symbolic link support to `additionalDirs` 

## Summary

`os.additionalDirs` copies a directory tree into the target image. The copy
(`file.CopyDir`) only handled **directories** and **regular files** — it had no
branch for **symbolic links**. As a result, symlinks in the source tree were
*dereferenced* (the target's contents were copied) or caused the whole build to
fail (dangling links, or links to special files like `/dev/zero`), instead of
being recreated as links in the image.

This change makes `additionalDirs` preserve symlinks **verbatim** (like
`cp --no-dereference`): each link is recreated in the image with the same target
string and is never followed at build time.

## Root cause

`toolkit/tools/internal/file/file.go` → `CopyDir` iterated the source entries with
only two branches:

```go
if entry.IsDir() {
    CopyDir(...)            // recurse
} else {
    NewFileCopyBuilder(...).SetFileMode(...).Run()   // regular file (dereferences symlinks)
}
```

A symlink reports `entry.IsDir() == false`, so it fell into the `else` branch and
`FileCopyBuilder.Run()` called `os.Stat` (which follows the link) → dereference or
error.

Call path: `customizeos.go doOsCustomizations` → `customizefiles.go
copyAdditionalDirs` → `safechroot.go AddDirs` → **`file.go CopyDir`** (root cause).

## The fix

- **`file.go` `CopyDir`** — added a third branch, checked **before** `IsDir()`, that
  detects `entry.Type()&os.ModeSymlink != 0` and recreates the link verbatim via the
  existing `FileCopyBuilder.SetNoDereference()` (which does `os.Lstat` → `os.Readlink`
  → `os.Symlink`). It intentionally does **not** apply `SetFileMode`, because
  permissions must not be changed on a symlink. Checking the symlink case first also
  means a symlink-to-directory is preserved as a link instead of being followed.

- **`imagehistory.go` `populateAdditionalDirs`** — the per-image history walks the
  source tree hashing each file with SHA256. `GenerateSHA256` opens (and therefore
  follows) the path, which fails on a dangling link and reads unbounded data from a
  special file. The walk callback now detects a symlink and hashes the **link target
  string** (`os.Readlink`) instead of the dereferenced contents.

### Why this is safe According to me (security note)

Recreating a link only reads the target *string* (`os.Readlink`) and writes the same
string (`os.Symlink`); it **never reads the target file on the build host**. The link
is resolved at image runtime against the *image's* own filesystem. So verbatim
preservation is both the requested behavior and strictly safer than the previous
dereferencing (which could copy host file contents into the image). No rejection of
absolute or tree-escaping links is required.

## What did NOT change

- `configvalidation.go` `validateAdditionalDirs` — only checks that each top-level
  `source` exists and is a directory; it never walks into the tree, and we reject no
  links.
- API schema — no config field was added, so `schema.json` is unchanged (no regen).

## How the bug was reproduced

Reproduced at the **unit tier** (fast, no root / no base image), because
`CopyDir` is the exact root cause:

```bash
cd toolkit/tools
go test ./internal/file/ -run TestCopyDirPreservesSymlinks -v
```

Before the fix this failed:

```
failed to copy file (.../src/dangling.txt) to (.../dst/dangling.txt):
  failed to read source file info:
  stat .../src/dangling.txt: no such file or directory
--- FAIL: TestCopyDirPreservesSymlinks
```

The failure shows the two symptoms:
1. A **dangling** symlink aborts the entire copy (`os.Stat` follows the link).
2. A **live** symlink is silently dereferenced into a regular file.

### Unit (no root, no base image)

- `internal/file/file_test.go` → `TestCopyDirPreservesSymlinks`
  Source dir with a **relative**, **absolute**, **dangling**, and
  **directory** symlink. Asserts each is recreated as a symlink with the
  original target string. Existing `TestCopyDir` still passes (no regression).
- `pkg/imagecustomizerlib/imagehistory_test.go` → `TestPopulateAdditionalDirsSymlink`
  Regular file + live link + dangling link. Asserts no error and that symlinks
  are hashed by target string (dangling link previously errored).

Both new unit tests run together in one command (fast, no root / no base image):

```bash
go test ./toolkit/tools/internal/file/ ./toolkit/tools/pkg/imagecustomizerlib/ \
  -run 'TestCopyDirPreservesSymlinks|TestPopulateAdditionalDirsSymlink' -v
```
**Unit Test Results**

```text
sthelkar@DESKTOP-0DD61PF:~/ImageCust/azure-linux-image-tools$ go test ./toolkit/tools/internal/file/ ./toolkit/tools/pkg/imagecustomizerlib/ \
  -run 'TestCopyDirPreservesSymlinks|TestPopulateAdditionalDirsSymlink' -v
=== RUN   TestCopyDirPreservesSymlinks
--- PASS: TestCopyDirPreservesSymlinks (0.00s)
PASS
ok      github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/file        (cached)
=== RUN   TestPopulateAdditionalDirsSymlink
--- PASS: TestPopulateAdditionalDirsSymlink (0.00s)
PASS
ok      github.com/microsoft/azure-linux-image-tools/toolkit/tools/pkg/imagecustomizerlib       (cached)

All pass.
```

### ADO pipeline validation

I also located and ran the
[`AzlImageTools-OneBranch-BuildImageCustomizerDev` pipeline](https://dev.azure.com/mariner-org/polar/_build/results?buildId=1192578)
against the feature branch. This run builds the Image Customizer binary and container and
executes the selected unit-test and VM-test matrix.

### Integration (requires root + base VHDX image)

- `pkg/imagecustomizerlib/customizefiles_test.go` →
  `TestCustomizeImageAdditionalDirsSymlinkPreserved`
  Builds an image with an `additionalDirs` source containing `zero -> /dev/zero`, then mounts the
  output and asserts `/a/zero` is a symlink pointing at `/dev/zero`.

  The old test expected the build to fail with "no space left on device" — that
  outcome was only a side effect of the dereferencing bug (it copied infinite bytes
  from `/dev/zero`). With symlinks preserved the build succeeds, so the test's
  expectation was flipped. `TestCustomizeImageAdditionalFilesInfiniteFile` uses
  `additionalFiles` with `source: /dev/zero` **directly** (not a symlink) and is
  unchanged — it still fails with out-of-space.

### Manual end-to-end validation

As an additional confidence check, I used the modified Image Customizer to produce a QEMU
image, booted the resulting image, and verified the symbolic links inside the running guest.
The links were preserved with their original target strings and behaved as expected.

## Docs

`docs/imagecustomizer/api/configuration/dirconfig.md` — added a **Symbolic links**
section describing verbatim preservation and that permission fields do not apply to
links (Added in v1.8).


toolkit/tools/internal/file/file.go
```diff
@@ -133,12 +133,21 @@ func CopyDir(src, dst string, newDirPermissions, childFilePermissions fs.FileMod
...
 		dstPath := filepath.Join(dst, entry.Name())
 
-		if entry.IsDir() {
+		switch {
+		case entry.Type()&os.ModeSymlink != 0:
+			// If it's a symlink, recreate it verbatim (no-dereference). We must not
+			// follow the link (which would copy the target's contents or fail on a
+			// dangling link) and must not chmod it. The link resolves inside the image
+			// at runtime, against the image's own filesystem.
+			if err := NewFileCopyBuilder(srcPath, dstPath).SetNoDereference().Run(); err != nil {
```
liulanze, line 142: `os.Symlink()` fails with `EEXIST` if the destination already exists. Files overwrite (O_TRUNC) and dirs merge, so symlink is now the only branch that hard fails.

vinceaperri: +1. Should the symlink branch remove dstPath first?

toolkit/tools/pkg/imagecustomizerlib/imagehistory.go
```diff
@@ -206,9 +208,23 @@ func populateAdditionalDirs(configAdditionalDirs imagecustomizerapi.DirConfigLis
...
+			if d.Type()&os.ModeSymlink != 0 {
+				// Symlinks are preserved verbatim in the image, so hash the link
+				// target string instead of following the link. Following it would
+				// fail on a dangling link and read an unbounded amount from a
+				// special file (e.g. a link to /dev/zero).
+				target, err := os.Readlink(path)
+				if err != nil {
+					return fmt.Errorf("error reading symlink %s:\n%w", path, err)
+				}
+				sum := sha256.Sum256([]byte(target))
```
vinceaperri, line 221: A regular file containing /bar and a symlink pointing at /bar hash identically here. The symlink input should be different (for example `"symlink:" + target`) so a change from a file to a link still shows up.

docs/imagecustomizer/api/configuration/dirconfig.md
```diff
@@ -69,3 +69,18 @@ os:
...
 
 Added in v0.3.
+
+## Symbolic links
+
+Symbolic links found inside the `source` directory are copied verbatim: each link is
+recreated in the target OS pointing at the same target path, and is **not** followed
+(dereferenced). The link is resolved at runtime against the target OS's own filesystem,
+so a link such as `foo -> /etc/hosts` points at the target OS's `/etc/hosts`, not the
+build host's. This also means dangling links and links to special files (for example
```
vinceaperri, line 79: additionalDirs now runs before copyAdditionalFiles and can place a link like `etc -> /` in the image, so a later additionalFiles destination resolves onto the build host. Is that in scope for this change?

docs/imagecustomizer/api/configuration/dirconfig.md
```diff
@@ -69,3 +69,18 @@ os:
...
+`/dev/zero`) are copied safely as links rather than by reading their targets.
+
+The permission fields (`newDirPermissions`, `mergedDirPermissions`,
+`childFilePermissions`) are not applied to symbolic links themselves.
+
+Added in v1.8.
```
vinceaperri, line 85: v1.7

vinceaperri (changes requested):

suresh-thelkar:
Making it draft as I am working as per the latest design doc and also considering the comments provided so far.

toolkit/tools/internal/file/file.go
suresh-thelkar: thanks, Good catch — fixed. The no-dereference branch now removes an existing destination before os.Symlink, matching the overwrite behavior of the file branch (dirs still merge). Added a regression test that runs the copy twice into the same destination.

toolkit/tools/pkg/imagecustomizerlib/imagehistory.go
suresh-thelkar: Fixed. A preserved symlink is now recorded as symlink:<sha256(target)> — the type tag lives in the stored value, not the hash input. Since a regular file's value is always bare hex, it can never collide with a link, even for a file whose contents are literally symlink:/bar.

docs/imagecustomizer/api/configuration/dirconfig.md
suresh-thelkar: @vinceaperri — that's a fair point.  Thank you for bringing this up. This PR does introduce a new way to create such links, since additionalDirs could not preserve symlinks previously.

That said, the underlying issue is broader than this feature. The redirect occurs because the copy layer resolves destinations on the host using filepath.Join and os.OpenFile, which will follow absolute or ../-escaping symlinks in the destination path, regardless of whether those links were introduced by this feature or already existed in the base image.

My view is that the fix belongs in the shared path-resolution layer, by using something like filepath-securejoin.SecureJoin(root, dest) in the chroot copy paths, rather than within the symlinkMode implementation itself. That approach would protect all additionalFiles and additionalDirs destinations, including symlinks originating from base images, and would allow us to add dedicated test coverage for absolute, relative, ..-escaping, and nested symlink scenarios.

Since this is effectively a security-hardening change with broader impact, I'd prefer to implement and review it separately rather than bundle it into the symlink preservation feature.

As an interim mitigation, symlinkMode: preserve remains opt-in and gated behind the preserve-symlinks preview feature. I'll open a follow-up issue and link it here and work accordingly on it . However, if you'd prefer that the SecureJoin hardening be included in this PR given the newly introduced vector, I'm happy to fold it in. Please share your thoughts.

suresh-thelkar:
@liulanze / @vinceaperri, Thank you very much for your feedback. Now that the design is approved, I started implementing it and addressed your comments. Please go through it and let me know if you have any further review comments.

docs/imagecustomizer/api/configuration/additionalfile.md
```diff
@@ -76,3 +76,23 @@ os:
 ```
 
 Added in v0.7.
+
+## symlinkMode [string]
```
cwize1, line 80: The preview flag should be attached to the entire field, not just the `preserve` value.

For example, we may want to change the name of the field in the future. Though, this isn't likely.

toolkit/tools/imagecustomizerapi/config.go
```diff
@@ -22,6 +22,19 @@ type Config struct {
 	BaseConfigs     []BaseConfig     `yaml:"baseConfigs" json:"baseConfigs,omitempty"`
 }
 
+// validateAdditionalFilesPreserveSymlinks requires the 'preserve-symlinks' preview feature
+// when any additionalFiles entry uses 'symlinkMode: preserve'. apiPath names the config
+// location (e.g. "os.additionalFiles") for the error message.
+func validateAdditionalFilesPreserveSymlinks(files AdditionalFileList, previewFeatures []PreviewFeature, apiPath string) error {
```
cwize1, line 28: Please put helper methods below public methods.

toolkit/tools/imagecustomizerapi/additionalfile.go
```diff
@@ -24,6 +24,12 @@ type AdditionalFile struct {
 	// The file permissions to set on the file.
 	Permissions *FilePermissions `yaml:"permissions" json:"permissions,omitempty"`
 
+	// Controls how a symbolic link 'source' is copied.
+	// 'dereference' (the default) follows the link and copies the target's contents.
```
cwize1, line 28: Attach the documentation for the individual enum values to the enum values themselves.

toolkit/tools/internal/file/file.go
```diff
@@ -90,9 +90,22 @@ func Copy(src, dst string) (err error) {
...
 // CopyDir copies src directory to dst, creating the dst directory if needed.
-// dst is assumed to be a directory and not a file.
+// Symbolic links are dereferenced to preserve the existing behavior.
 func CopyDir(src, dst string, newDirPermissions, childFilePermissions fs.FileMode, mergedDirPermissions *fs.FileMode) (err error) {
+	return CopyDirWithOptions(src, dst, newDirPermissions, childFilePermissions, mergedDirPermissions, CopyDirOptions{})
+}
+
+// CopyDirWithOptions copies src directory to dst using the requested symbolic link behavior.
+// dst is assumed to be a directory and not a file.
+func CopyDirWithOptions(src, dst string, newDirPermissions, childFilePermissions fs.FileMode,
```
cwize1, line 106: Add `newDirPermissions`, `childFilePermissions`, and `mergedDirPermissions` to `CopyDirOptions` as well?

cwize1, line 106: There aren't that many callers of `CopyDir` and `CopyDirWithOptions`. Why not just keep it as a single `CopyDir` function?

toolkit/tools/internal/file/file.go
```diff
@@ -133,12 +146,22 @@ func CopyDir(src, dst string, newDirPermissions, childFilePermissions fs.FileMod
...
 		dstPath := filepath.Join(dst, entry.Name())
 
-		if entry.IsDir() {
+		switch {
+		case options.NoDereference && entry.Type()&os.ModeSymlink != 0:
+			// If it's a symlink, recreate it verbatim (no-dereference). We must not
+			// follow the link (which would copy the target's contents or fail on a
+			// dangling link) and must not chmod it. The link resolves inside the image
+			// at runtime, against the image's own filesystem.
+			if err := NewFileCopyBuilder(srcPath, dstPath).SetNoDereference().Run(); err != nil {
```
cwize1, line 155: Considering that we know that the "file" is a symlink at this point, invoking the entire `FileCopyBuilder` seems a little much. I would move the copy-symlink logic into a helper function that both `FileCopyBuilder.Run` and this function can use.

toolkit/tools/pkg/imagecustomizerlib/customizefiles_test.go
```diff
@@ -272,38 +272,40 @@ func verifyAddDirs(t *testing.T, baseImageInfo testBaseImageInfo, buildDir strin
 	verifyFileContentsSame(t, animalsFileOrigPath, animalsFileNewPath)
 }
 
-func TestCustomizeImageAdditionalDirsInfiniteFile(t *testing.T) {
+func TestCustomizeImageAdditionalDirsSymlinkPreserved(t *testing.T) {
```
cwize1, line 275: Restore the `TestCustomizeImageAdditionalDirsInfiniteFile` test, while also keeping the `TestCustomizeImageAdditionalDirsSymlinkPreserved` test.

cwize1:
Please simplify the PR description. A single paragraph (or two) briefly describing the change is sufficient.

docs/imagecustomizer/api/configuration/additionalfile.md
suresh-thelkar: Done. Validation now requires the symlink-mode preview feature for any explicit symlinkMode value (including dereference), not just preserve — only omitting the field is free. Applied to both additionalDirs and additionalFiles in Config.IsValid, and the docs now say "setting symlinkMode (to any value) requires the preview feature." (Also renamed the feature preserve-symlinks → symlink-mode so, per your point about a future rename, the flag tracks the field name.)

toolkit/tools/imagecustomizerapi/config.go
suresh-thelkar: Done. validateAdditionalFilesSymlinkMode is now defined below Config.IsValid().

toolkit/tools/imagecustomizerapi/additionalfile.go
suresh-thelkar: Done. The per-value descriptions now live on the SymlinkMode constants in symlinkmode.go (SymlinkModeUnspecified/Dereference/Preserve); the struct field just references the type.

toolkit/tools/internal/file/file.go
suresh-thelkar: Done — All three permission values now live on CopyDirOptions alongside NoDereference.

suresh-thelkar: Done. Collapsed back to a single CopyDir(src, dst string, options CopyDirOptions); removed CopyDirWithOptions and updated the three in-module callers (safechroot.AddDirs, artifactsinputoutput.go, selinuxpolicyoutput.go).

toolkit/tools/internal/file/file.go
suresh-thelkar: Done. Extracted copySymlink(src, dst) (Lstat/Readlink/remove-existing/Symlink) and call it from both CopyDir's symlink branch and FileCopyBuilder.Run, so CopyDir no longer spins up a full FileCopyBuilder for a link.

toolkit/tools/pkg/imagecustomizerlib/customizefiles_test.go
suresh-thelkar: Done. TestCustomizeImageAdditionalDirsInfiniteFile is restored (default dereference still fails with out-of-space on the zero link) and TestCustomizeImageAdditionalDirsSymlinkPreserved is kept (opt-in preserve succeeds).

suresh-thelkar:
Thanks @cwize1 and @vinceaperri for the review. I have implemented your comments. Please look into the latest commit and let me know if there is any further feedback. Also rebased on latest main and re-verified (build/vet/schema/tests + a booted-VM check).

docs/imagecustomizer/api/configuration/dirconfig.md
```diff
@@ -32,6 +33,28 @@ The absolute path in the target OS that the source directory will be copied to.
 
 Added in v0.3.
 
+## symlinkMode [string]
+
+Controls how symbolic links inside the source directory are copied.
+
+Setting `symlinkMode` (to any value) requires the `symlink-mode`
```
cwize1: Please use the standard preview feature paragraph, to be consistent with the rest of the docs.

```Markdown
This is a preview feature.
Its API and behavior is subject to change.
You must enable this feature by specifying `symlink-mode` in the
[previewFeatures](./config.md#previewfeatures-string) API.
```

cwize1 (changes requested):

docs/imagecustomizer/api/configuration/dirconfig.md
suresh-thelkar: Done

suresh-thelkar:
Thanks @cwize1, @vinceaperri, and @liulanze for the reviews.

Latest changes:

Updated the symlinkMode docs to use the standard preview-feature paragraph (@cwize1).
Simplified the PR description.
Everything else from earlier rounds is already addressed. I've rebased on latest main and re-verified locally (build, vet, schema, unit tests).

I've also run the [GitHub Workflow ](https://github.com/microsoft/azure-linux-image-tools/actions/runs/35705728765)successfully after all the changes made so far. It ran successfully.

If there's no further feedback, I'd appreciate a final sign-off. Thanks!

cwize1 (approved):

suresh-thelkar:
Thank you, @cwize1 , for the thorough review and approval.

@vinceaperri / @liulanze , I am still waiting for your sign-off. If there are no additional comments or feedback, could you please approve the PR at your convenience? That will allow us to move forward with merging it.

cwize1 (approved):

PR #892, opened 2026-09-03
Fix roundDown function.

cwize1:
Fix the roundDown function.

Also, move the `RoundDown`, `RoundUp`, `DivRoundUp` functions into a shared `mathutils` package.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/internal/mathutils/mathutils.go
+++ b/toolkit/tools/internal/mathutils/mathutils.go
@@ -1,0 +1,33 @@
+package mathutils
+
+type Integer interface {
+	~int | ~int8 | ~int16 | ~int32 | ~int64 | ~uint | ~uint8 | ~uint16 | ~uint32 | ~uint64 | ~uintptr
+}
+
+func RoundUp[I Integer](size I, alignment I) I {
+	div := size / alignment
+	mod := size % alignment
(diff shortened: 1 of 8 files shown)


toolkit/tools/internal/mathutils/mathutils.go
```diff
@@ -0,0 +1,33 @@
+package mathutils
+
+type Integer interface {
```
liulanze: Integer includes signed types, but Go functions do not handle negative values correctly because Go integer division truncates toward zero rather than flooring. Should we restrict the constraint to unsigned integer types?

For example:

```
type UnsignedInteger interface {
    ~uint | ~uint8 | ~uint16 | ~uint32 | ~uint64 | ~uintptr
}
```

liulanze: But not a blocker, I think current production callers use non-negative sizes.

liulanze (dismissed):

toolkit/tools/internal/mathutils/mathutils.go
cwize1: Fair. Fixed

liulanze (approved):
LGTM

PR #894, opened 2026-09-04
Improve DiskSize.IsValid().

cwize1:
Move the DiskSize alignment checks from the unmarshal functions to the IsValid function. This will improve the error message, since the IsValid functions provide more context for the error. Also, it ensures that the check is performed even when the type is initialized programatically.

---

### **Checklist**

- [x] Tests added/updated

### DIFF
--- a/toolkit/tools/imagecustomizerapi/disksize.go
+++ b/toolkit/tools/imagecustomizerapi/disksize.go
@@ -21,6 +21,12 @@ var (
 type DiskSize uint64
 
 func (s *DiskSize) IsValid() error {
+	// The imager's diskutils works in MiB. So, restrict disk and partition sizes to multiples of 1 MiB.
+	if *s%DefaultPartitionAlignment != 0 {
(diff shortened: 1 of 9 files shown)


toolkit/tools/imagecustomizerapi/disksize.go
```diff
@@ -119,19 +125,15 @@ func parseDiskSize(diskSizeString string) (DiskSize, error) {
...
-	}
-
 	return DiskSize(num), nil
 }
 
 // String returns the string representation of DiskSize in the most appropriate unit
 // such that it matches the input format.
 func (s DiskSize) String() string {
 	switch {
+	case s == 0:
```
vinceaperri, line 135: 0 fails this type's JSONSchema pattern. Is this case necessary?

cwize1: Seeing `0T` is kind of weird. But reverted for now.

liulanze (approved):
lgtm

PR #900, opened 2026-09-16
Add package manifest creation and export

vinceaperri:
This adds package manifests to Image Customizer so we can remove the package manager without losing the inventory vulnerability scanners need.

The `package-manifest` preview feature adds `create`, `passthrough`, and `none` modes under `os.packages.manifest.mode`, plus export through `output.packageManifest.path` or `--output-package-manifest-file`. Creation reads the final RPM database after post-customization scripts and package-manager removal, but before database cleanup.

Manifests use SPDX 2.2, built with the SPDX and PURL libraries, and keep RPM epoch, architecture, and vendor metadata. Creation works on Azure Linux 3/4, Azure Container Linux, and Fedora. Ubuntu can preserve or remove manifests, but cannot yet create them.

I also consolidated the RPM helpers, shared read-only image mounting with SELinux policy export, and tightened Trivy report parsing. Validated, version-pinned license choices select Apache-2.0 for `github.com/spdx/tools-golang`.

This includes config validation, inheritance, and documentation updates, unit and golden-manifest tests, and Azure Linux 3/4 functional tests for creation, export, and package-manager removal.


toolkit/tools/internal/spdxmanifest/spdxmanifest.go
```diff
@@ -0,0 +1,171 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// Package spdxmanifest builds package manifests.
+package spdxmanifest
```
cwize1, line 5: Why aren't we using the [official SPDX go library](github.com/spdx/tools-golang)?

toolkit/tools/imagecustomizerapi/os.go
cwize1: It's weird that the `Packages` type doesn't have an `IsValid` function.

toolkit/tools/internal/purl/purl.go
```diff
@@ -0,0 +1,25 @@
...
+	return fmt.Sprintf("pkg:rpm/%s/%s@%s?%s",
+		purlEncode(namespace),
+		purlEncode(name),
+		purlEncode(version+"-"+release),
+		strings.Join(qualifiers, "&"))
+}
+
+func purlEncode(componentData string) string {
+	// QueryEscape encodes spaces as '+', but PURL components require '%20'.
+	return strings.ReplaceAll(url.QueryEscape(componentData), "+", "%20")
```
cwize1: `namespace` and `name` should use `url.PathEscape` not `url.QueryEscape`.

toolkit/tools/internal/spdxmanifest/spdxmanifest.go
vinceaperri: I hadn't considered it. This started as a port of the Python generator I wrote for ACL, which also doesn't use an SPDX library. I'll evaluate what we can reuse from tools-golang while preserving the output we need.

toolkit/tools/internal/targetos/targetos.go
```diff
@@ -168,6 +173,11 @@ func GetInstalledTargetOsFromEnvFields(fields map[string]string) (TargetOs, erro
 	distroId := fields["ID"]
 	versionId := fields["VERSION_ID"]
 
+	packageManifestVersionInfo := fields["VERSION"]
+	if packageManifestVersionInfo != "" && fields["BUILD_ID"] != "" {
+		packageManifestVersionInfo += "+" + fields["BUILD_ID"]
```
cwize1: This is placing package manifest logic in a place not related to package manifests.

At a minimum, you should store `VERSION` and `BUILD_ID` in `TargetOs`, then let the package manifest logic create the value it needs.

That being said, would it cause any harm for the package manifest logic to re-read the os-release file? Actually, it may be preferrable to do so, in case the os-release file is updated (e.g. in a `postCustomization` script).

toolkit/tools/pkg/imagecustomizerlib/configvalidation.go
```diff
@@ -1132,3 +1163,35 @@ func resolveRemovePackageManager(configChain []*ConfigWithBasePath) bool {
...
+			break
+		}
+	}
+
+	if removePackageManager && packageManifestMode == imagecustomizerapi.PackageManifestModeUnspecified {
+		return packageManifestMode, ErrPackageManifestRemovalModeRequired
+	}
+
+	if outputPackageManifestPath == "" {
+		return packageManifestMode, nil
```
cwize1: This early return is a tad confusing. It might be better to invert the if-statement and pull in the switch statement.

toolkit/tools/pkg/imagecustomizerlib/customizepackages_rpm.go
```diff
@@ -10,11 +10,140 @@ import (
...
+	ErrInvalidArch            = errors.New("invalid RPM package architecture")
+	ErrNevraArchSeparator     = errors.New("missing architecture separator in RPM NEVRA")
+	ErrNevraReleaseSeparator  = errors.New("missing release separator in RPM NEVRA")
+	ErrNevraVersionSeparator  = errors.New("missing version separator in RPM NEVRA")
+	ErrInvalidRpmQueryColumns = errors.New("expected '<nevra>\\t<vendor>'")
+)
+
+// RpmPackage is an installed RPM, identified by its name, epoch, version, release and architecture.
+// Vendor is not identifying information, but is included in the RPM query output.
+type RpmPackage struct {
```
cwize1: Give `RpmPackage` its own file.

toolkit/tools/pkg/imagecustomizerlib/customizepackages_rpm.go
```diff
@@ -10,11 +10,140 @@ import (
...
+	rpmQueryFormat = `%|ARCH?{%{NEVRA}\t%{VENDOR}\n}:{}|`
+
+	// rpmTagNone is what rpm's query format prints for a tag the package does not carry.
+	rpmTagNone = "(none)"
+
+	rpmQueryColumns = 2
+)
+
+var (
+	ErrInvalidName            = errors.New("invalid RPM package name")
```
cwize1: Use `NewImageCustomizerError` instead of `errors.New`?

cwize1: The name `ErrInvalidName` is a little too generic. Perhaps `ErrInvalidRpmName`?

toolkit/tools/pkg/imagecustomizerlib/customizepackages_rpm.go
```diff
@@ -10,11 +10,140 @@ import (
...
+		Epoch:   epoch,
+		Version: version,
+		Release: release,
+		Arch:    arch,
+		Vendor:  vendor,
+	}, nil
+}
+
+// newRpmPackageFromNevra builds an RpmPackage from a "name-[epoch:]version-release.arch" NEVRA string.
+func newRpmPackageFromNevra(nevra string, vendor string) (RpmPackage, error) {
```
cwize1: I believe you can parse NEVRA using a regex, since wildcards can be greedy (and are greedy by default).

For example: `^(\S+)-((\d+):)?(\S+)-(\S+)\.(\S+)$`

toolkit/tools/pkg/imagecustomizerlib/customizepackages_rpm.go
```diff
@@ -120,14 +249,9 @@ func installRpmPackages(ctx context.Context, allPackages []string,
 
 	args = append(args, allPackages...)
 
-	if toolsChroot != nil {
```
cwize1, line 123: It's generally nice to do this sort of cleanup change in a separate PR, to avoid adding extra complexity to this PR.

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
```diff
@@ -129,27 +130,55 @@ func needPackageCleanup(config *imagecustomizerapi.OS) bool {
...
+		installedPackages, err = preparePackageManifestCreate(ctx, distroHandler, imageChroot, toolsChroot,
+			packageManifestMode)
+		if err != nil {
+			return err
+		}
+	}
+
+	removedPackageIds := []string{}
+	if removePackageManager {
+		removedPackageIds, err = removeOsPackageManager(ctx, distroHandler, imageChroot, toolsChroot)
```
cwize1: One thing that's fun is that the apt logs don't provide the full package ID.

toolkit/tools/pkg/imagecustomizerlib/distrohandler.go
```diff
@@ -67,6 +71,10 @@ type DistroHandler interface {
 	// toolsChroot has the same semantics as in IsPackageInstalled.
 	GetAllPackagesFromChroot(imageChroot safechroot.ChrootInterface, toolsChroot *safechroot.Chroot) ([]cosiapi.OsPackage, error)
 
+	// ListInstalledPackages returns installed package IDs and metadata for manifest generation.
+	// toolsChroot has the same semantics as in IsPackageInstalled.
+	ListInstalledPackages(imageChroot safechroot.ChrootInterface, toolsChroot *safechroot.Chroot) ([]spdxmanifest.Package, error)
```
cwize1: Rename to something like `ListInstalledPackagesForSpdx`. Otherwise, it isn't necessarily clear what its difference is to `GetAllPackagesFromChroot`.

toolkit/tools/pkg/imagecustomizerlib/packagemanifest.go
```diff
@@ -0,0 +1,258 @@
...
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/safechroot"
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/spdxmanifest"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+)
+
+const packageManifestPath = "usr/share/os-manifests/package-manifest.spdx.json"
+
+var (
+	ErrPackageManifestRead                = NewImageCustomizerError("Packages:PackageManifestRead", "failed to read the package manifest")
```
cwize1: You could set the prefix to `PackageManifest:`.

toolkit/tools/pkg/imagecustomizerlib/packagemanifest.go
```diff
@@ -0,0 +1,258 @@
...
+func applyPackageManifestMode(ctx context.Context, distroHandler DistroHandler, imageChroot *safechroot.Chroot,
+	toolsChroot *safechroot.Chroot, packageManifestMode imagecustomizerapi.PackageManifestMode,
+	buildTime string, installedPackages []spdxmanifest.Package, removedPackages []string,
+) error {
+	if packageManifestMode == imagecustomizerapi.PackageManifestModeUnspecified {
+		return nil
+	}
+
+	manifestPath := filepath.Join(imageChroot.RootDir(), packageManifestPath)
+	exists, err := file.PathExists(manifestPath)
```
cwize1: Why is this `file.PathExists` before the telemetry `Start`?

toolkit/tools/pkg/imagecustomizerlib/packagemanifest.go
```diff
@@ -0,0 +1,258 @@
...
+		return fmt.Errorf("%w (path='%s'):\n%w", ErrPackageManifestRead, manifestPath, err)
+	}
+
+	ctx, span := otel.GetTracerProvider().Tracer(OtelTracerName).Start(ctx, "apply_package_manifest_mode")
+	defer span.End()
+	span.SetAttributes(attribute.String("mode", string(packageManifestMode)))
+
+	switch packageManifestMode {
+	case imagecustomizerapi.PackageManifestModePassthrough:
+		logger.Log.Infof("Skipping package manifest changes (mode='%s')", packageManifestMode)
```
cwize1: This `mode` value in the log seems redundant.

toolkit/tools/internal/spdxmanifest/spdxmanifest.go
```diff
@@ -0,0 +1,171 @@
...
+
+	// The namespace need not resolve. It only has to be a unique URI for the document.
+	documentNamespaceBase = "https://azurelinux.microsoft.com/spdxdocs"
+)
+
+// An SPDXID must be an "idstring"; everything else in a package name is replaced with a dash.
+var nonSpdxID = regexp.MustCompile(`[^A-Za-z0-9.\-]`)
+
+// BuildOptions carries the document-level values a refresh cannot derive from the package set.
+type BuildOptions struct {
```
cwize1: This seems more like "metadata" than "options"

toolkit/tools/pkg/imagecustomizerlib/packagemanifest.go
```diff
@@ -0,0 +1,258 @@
...
+	partitionsLayout []fstabEntryPartNum, distroHandler DistroHandler,
+) error {
+	logger.Log.Infof("Extracting package manifest from image")
+
+	_, span := otel.GetTracerProvider().Tracer(OtelTracerName).Start(ctx, "output_package_manifest")
+	defer span.End()
+
+	imageMountPoint := filepath.Join(buildDir, "package-manifest-extract")
+
+	imageConnection, _, err := reconnectToExistingImage(ctx, buildImage, buildDir, imageMountPoint,
```
cwize1: It might be nice to share the image connection between `outputSelinuxPolicy` and `outputPackageManifest`.

toolkit/tools/pkg/imagecustomizerlib/customizepackages_test.go
```diff
@@ -960,6 +961,68 @@ func testCustomizeImagePackagesInstallOnline(t *testing.T, baseImageInfo testBas
 	ensurePackageCacheCleanup(t, imageConnection, baseImageInfo)
 }
 
+func TestGetRpmRootArgs(t *testing.T) {
```
cwize1: Are the tests `TestGetRpmRootArgs` and `TestGetRpmRemoveArgs` needed? I think their functionality is exercised pretty thoroughly through existing tests.

cwize1 (changes requested):

toolkit/tools/internal/spdxmanifest/spdxmanifest.go
vinceaperri: We can use it to encode the JSON. Updated.

toolkit/tools/imagecustomizerapi/os.go
vinceaperri: Yes, I noticed that.

toolkit/tools/internal/spdxmanifest/spdxmanifest.go
vinceaperri: Updated.

go.mod
```diff
@@ -4,8 +4,10 @@ go 1.25.0
 
 require (
 	github.com/alecthomas/kong v1.16.1
+	github.com/anchore/go-rpmdb v0.2.0
 	github.com/asaskevich/govalidator v0.0.0-20230301143203-a9d515a09cc2
 	github.com/cavaliergopher/cpio v1.0.1
+	github.com/glebarez/go-sqlite v1.20.3
```
cwize1: I can't say I'm a fan of how old some of these dependencies are.

toolkit/tools/pkg/imagecustomizerlib/packagemanifest.go
vinceaperri: Fixed.

toolkit/tools/pkg/imagecustomizerlib/customizepackages.go
vinceaperri: In my tests, dnf also did not provide the full ID in its stdout. Switched to querying the RPM database directly.

toolkit/tools/pkg/imagecustomizerlib/customizepackages_rpm.go
```diff
@@ -343,6 +313,87 @@ func getAllPackagesFromChrootRpm(imageChroot safechroot.ChrootInterface, toolsCh
...
+	}
+	if err := database.Close(); err != nil {
+		return nil, fmt.Errorf("failed to close RPM database (path='%s'):\n%w", databasePath, err)
+	}
+
+	installedPackages := make([]spdxmanifest.Package, 0, len(packages))
+	installedIds := make(map[string]struct{}, len(packages))
+	for _, packageInfo := range packages {
+		if packageInfo.Arch == "" {
+			logger.Log.Debugf("Skipping RPM record without architecture (name='%s', version='%s', release='%s')",
```
cwize1, line 360: Did you observe this during testing? What packages are affected by this?

toolkit/tools/pkg/imagecustomizerlib/customizepackages_rpm.go
vinceaperri: This code was removed.

toolkit/tools/pkg/imagecustomizerlib/customizepackages_rpm.go
```diff
@@ -343,6 +313,87 @@ func getAllPackagesFromChrootRpm(imageChroot safechroot.ChrootInterface, toolsCh
...
+				packageInfo.Name, packageInfo.Version, packageInfo.Release)
+			continue
+		}
+		manifestPackage, err := newManifestPackageFromRpmPackage(*packageInfo, purlNamespace)
+		if err != nil {
+			return nil, err
+		}
+
+		if _, found := installedIds[manifestPackage.ID]; found {
+			logger.Log.Warnf("Duplicate installed package ID (id='%s')", manifestPackage.ID)
```
cwize1: Did you observe this during testing?

For IC, I reserve WARN for things that went wrong that the user must fix. For example, failing to unmount a mount. For things that might be a problem but can be safely ignored in some circumstances, then use INFO or below.

toolkit/tools/pkg/imagecustomizerlib/customizepackages_rpm.go
vinceaperri: Code was removed.

toolkit/tools/pkg/imagecustomizerlib/customizepackages_rpm.go
vinceaperri: Code was removed.

toolkit/tools/pkg/imagecustomizerlib/customizepackages_test.go
vinceaperri: Removed.

toolkit/tools/pkg/imagecustomizerlib/packagemanifest_test.go
```diff
@@ -0,0 +1,397 @@
...
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/safechroot"
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/shell"
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/spdxmanifest"
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/testutils"
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type packageManifestTestPackage struct {
```
cwize1: Why are we redeclaring the SPDX types in the test code? Just use the existing types?

toolkit/tools/pkg/imagecustomizerlib/packagemanifest_test.go
```diff
@@ -0,0 +1,397 @@
...
+		packageIdentity, err := url.PathUnescape(packageURL.Opaque)
+		assert.NoError(t, err, pkg.Name)
+
+		version := pkg.VersionInfo
+		epoch := ""
+		if parsedEpoch, parsedVersion, found := strings.Cut(version, ":"); found {
+			epoch, version = parsedEpoch, parsedVersion
+		}
+
+		assert.Equal(t, fmt.Sprintf("rpm/%s/%s@%s", namespace, pkg.Name, version), packageIdentity)
```
cwize1: Avoid duplicating the product logic in test code.

cwize1 (changes requested):

toolkit/tools/pkg/imagecustomizerlib/configvalidation.go
vinceaperri: Done.

toolkit/tools/internal/targetos/targetos.go
vinceaperri: Done.

toolkit/tools/internal/purl/purl.go
vinceaperri: Switched to `github.com/package-url/packageurl-go` for PURL construction and escaping.

toolkit/tools/pkg/imagecustomizerlib/packagemanifest.go
vinceaperri: Done.

toolkit/tools/pkg/imagecustomizerlib/packagemanifest.go
vinceaperri: Removed.

toolkit/tools/pkg/imagecustomizerlib/packagemanifest.go
vinceaperri: Updated.

toolkit/tools/pkg/imagecustomizerlib/distrohandler.go
vinceaperri: Done.

toolkit/tools/pkg/imagecustomizerlib/customizepackages_rpm.go
vinceaperri: The parser was removed when I switched to reading the RPM database directly.

toolkit/tools/pkg/imagecustomizerlib/packagemanifest_test.go
vinceaperri: Done.

toolkit/tools/pkg/imagecustomizerlib/packagemanifest_test.go
vinceaperri: Updated.

toolkit/tools/pkg/imagecustomizerlib/customizepackages_rpm.go
vinceaperri: No duplicate IDs were logged in the Azure Linux 3 and 4 manifest-creation runs. Changed this to INFO.

toolkit/tools/pkg/imagecustomizerlib/customizepackages_rpm.go
vinceaperri: Yes. `gpg-pubkey` records on both Azure Linux 3 and 4 had an empty architecture. Those records are skipped.

go.mod
vinceaperri: Updated the SQLite driver and its dependencies. Added `packageurl-go` v0.1.7 for PURL generation.

toolkit/tools/pkg/imagecustomizerlib/packagemanifest.go
```diff
@@ -0,0 +1,217 @@
...
+	span.SetAttributes(attribute.Int("package_count", len(packages)))
+
+	versionInfo, err := readPackageManifestVersion(imageChroot)
+	if err != nil {
+		return fmt.Errorf("%w:\n%w", ErrPackageManifestVersion, err)
+	}
+	metadata := spdxmanifest.BuildMetadata{
+		Name:        string(distroHandler.GetTargetOs().Distro),
+		VersionInfo: versionInfo,
+		Created:     buildTime,
```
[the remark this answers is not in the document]
vinceaperri: Fixed. Making buildTime UTC also touches /etc/image-customizer-release and the image history feature though -- @cwize1 FYI

toolkit/tools/internal/spdxmanifest/spdxmanifest.go
```diff
@@ -0,0 +1,174 @@
...
+			RefB:         spdxcommon.DocElementID{ElementRefID: pkgId},
+			Relationship: spdxcommon.TypeRelationshipContains,
+		})
+	}
+
+	spdxDocument := spdx.Document{
+		SPDXVersion:       spdx.Version,
+		SPDXIdentifier:    documentID,
+		DocumentName:      metadata.Name,
+		DocumentNamespace: documentNamespace(metadata.Name, metadata.VersionInfo, sortedPackages),
```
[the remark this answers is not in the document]
vinceaperri: Fixed. The entire document is the input now.

toolkit/tools/pkg/imagecustomizerlib/distrohandler_acl.go
```diff
@@ -131,6 +134,11 @@ func (d *aclDistroHandler) GetPackageInformation(imageChroot *safechroot.Chroot,
 	return d.packageManager.getPackageInformation(imageChroot, toolsChroot, packageName)
 }
 
+func (d *aclDistroHandler) ListInstalledPackagesForSpdx(imageChroot safechroot.ChrootInterface,
+) ([]spdxmanifest.Package, error) {
+	return listInstalledPackagesRpm(imageChroot, rpmDatabasePathAzl3, purlNamespaceAzureLinux)
+}
```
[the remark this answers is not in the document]
vinceaperri: Not a misidentification. The RPMs come from AZL3.

toolkit/tools/pkg/imagecustomizerlib/customizepackages_rpm.go
cwize1: Since IC is an offline tool, we don't necessarily need to be robust to weird situations. We can just error out and then fix it if and when a user hits the problem in a valid scenario.

toolkit/tools/pkg/imagecustomizerlib/createimagehelper.go
```diff
@@ -73,7 +73,7 @@ func doOsCustomizationsCreate(
 	distroHandler DistroHandler,
 ) error {
 	imageChroot := imageConnection.Chroot()
-	buildTime := time.Now().Format(buildTimeFormat)
+	buildTime := time.Now().UTC().Format(buildTimeFormat)
```
cwize1: Please split this off into a separate change.

toolkit/tools/imagecustomizerapi/os.go
```diff
@@ -127,9 +127,8 @@ func (s *OS) IsValid() error {
 		}
 	}
 
-	err = s.Packages.SnapshotTime.IsValid()
-	if err != nil {
-		return fmt.Errorf("invalid package snapshot time:\n%w", err)
+	if err := s.Packages.IsValid(); err != nil {
+		return err
```
cwize1: nit: `fmt.Errorf("invalid packages:\n%w", err)`

toolkit/tools/pkg/imagecustomizerlib/packagemanifest.go
```diff
@@ -0,0 +1,217 @@
...
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/file"
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/logger"
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/safechroot"
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/spdxmanifest"
+	"github.com/microsoft/azure-linux-image-tools/toolkit/tools/internal/targetos"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+)
+
+const packageManifestPath = "usr/share/os-manifests/package-manifest.spdx.json"
```
cwize1: nit: `"/usr/share/os-manifests/package-manifest.spdx.json"`

Go's `filepath.Join` method still joins the paths even if the addition is an absolute path.

toolkit/tools/pkg/imagecustomizerlib/customizepackages_rpm.go
vinceaperri: Done.

toolkit/tools/pkg/imagecustomizerlib/createimagehelper.go
vinceaperri: Removed.

toolkit/tools/imagecustomizerapi/os.go
vinceaperri: Updated.

toolkit/tools/pkg/imagecustomizerlib/packagemanifest.go
vinceaperri: Fixed.

cwize1 (approved):

